Issue #10, opened 2018-08-29
ResourceManager should be thread-safe


tamirkamara:
Implement in such a way that resources cached in the client will be handled by a single thread at a time.
This way, the refresh procedures will not happen more than one time in parallel.

PR #24, opened 2018-09-20
Resource manager thread safe

rabee333:
#10 ResourceManager must be thread safe.
Remove string formatting in log messages.


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -144,7 +144,7 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
             }
             return ingestionResult;
         } catch (Exception ex) {
-            log.error(String.format("ingestFromStream: Error while ingesting from stream. Error: %s", ex.getMessage()), ex);
+            log.error("ingestFromStream: Error while ingesting from stream. Error: {}", ex.getMessage(), ex);
```
toshetah: Do we need both getMessage and ex?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -59,7 +65,7 @@ public void run() {
                 try {
                     refreshIngestionResources();
                 } catch (Exception e) {
-                    log.error(String.format("Error in refreshIngestionResources: %s.", e.getMessage()), e);
+                    log.error("Error in refreshIngestionResources: {}.", e.getMessage(), e);
```
toshetah: Do we need both getMessage and ex?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -70,7 +76,7 @@ public void run() {
                 try {
                     refreshIngestionAuthToken();
                 } catch (Exception e) {
-                    log.error(String.format("Error in refreshIngestionAuthToken: %s.", e.getMessage()), e);
+                    log.error("Error in refreshIngestionAuthToken: {}.", e.getMessage(), e);
```
toshetah: same here

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -80,66 +86,97 @@ public void run() {
...
+            refreshIngestionResources();
+            try{
+                // If the write lock is locked, then the read will wait here.
+                // In other words if the refresh is running yet, then wait until it ends:
+                ingestionResourcesLock.readLock().lock();
+                if (!ingestionResources.containsKey(resourceType)) {
+                    throw new Exception("Unable to get ingestion resources for this type: " + resourceType.getName());
+                }
+            } finally {
+                ingestionResourcesLock.readLock().unlock();
```
toshetah: Can it be that the lock failed, and than we'll get here another error?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -80,66 +86,97 @@ public void run() {
...
             }
         }
-        return identityToken;
+        return ingestionResources.get(resourceType).nextValue();
```
toshetah: Shouldn't that be under the lock?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -80,66 +86,97 @@ public void run() {
...
-        return ingestionResources.get(resourceType).nextValue();
+        return identityToken;
     }
 
     int getSize(ResourceTypes resourceType){
         return ingestionResources.containsKey(resourceType) ? ingestionResources.get(resourceType).getSize() : 0;
     }
 
-    private void addValue(String key, String value) {
+    private void addIngestionResource(HashMap<ResourceTypes, IngestionResource> newIngestionResources, String key, String value) {
```
toshetah: minor: Rename newIngestionResources into ingestionResources

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -80,66 +86,97 @@ public void run() {
...
     }
 
-    public String getIngestionResource(ResourceTypes resourceType) throws Exception {
-        if (!ingestionResources.containsKey(resourceType)) {
-            refreshIngestionResources();
-            if (!ingestionResources.containsKey(resourceType)) {
-                throw new Exception("Unable to get ingestion resources for this type: " + resourceType.getName());
+    String getKustoIdentityToken() throws Exception {
+        if (identityToken == null) {
+            refreshIngestionAuthToken();
```
toshetah: In the current implementation, for every ask of the token, we'll try to acquire a write token. 
Isn't is better to first check if we need to refresh and only then try to acquire the lock?
There is the Double-checked locking (if lock if) pattern for that.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
tamirkamara: No, we talked about this - SLF4J prints out the message along with the stack

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
tamirkamara: same

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
rabee333: The implementation of the refreshIngestionResources() was changed so it finishes retrieving the resource values into a new resources map, and only then it replaces the old resources map with the new resources map in one atomic operation. Therefore, when calling the getIngestionResource() method, if the key is available, it will return it instantly. Only if the key is not available it runs refresh with write lock into it, and read lock in this method to ensure returning the value when the refresh finishes.
(If during periodic refresh, this method was called, then it will return the available value before the refresh ends, and this is OK.)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
rabee333: if the lock fails due to error, it will throw an exception to the caller of this method.
if the lock is already locked, it will wait here until it released.
(or I didn't understand the question :)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
rabee333: sure, it was a mistake from the last merge!

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
toshetah: Why would the token be null unless it's the first refresh?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -80,66 +86,97 @@ public void run() {
...
 
         } catch (Exception e) {
-            log.error(String.format("Error in initializing ResourceManager: %s.", e.getMessage()), e);
+            log.error("Error in initializing ResourceManager: {}.", e.getMessage(), e);
             throw e;
         }
     }
 
-    public void clean() {
+    void clean() {
```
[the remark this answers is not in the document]
rabee333: Removed

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -80,41 +86,49 @@ public void run() {
...
+                // In other words if the refresh is running yet, then wait until it ends:
+                ingestionResourcesLock.readLock().lock();
+                if (!ingestionResources.containsKey(resourceType)) {
+                    throw new Exception("Unable to get ingestion resources for this type: " + resourceType.getName());
+                }
+            } finally {
+                ingestionResourcesLock.readLock().unlock();
+            }
+        }
+        return ingestionResources.get(resourceType).nextValue();
```
toshetah: What will happen if ingestionResources is replaced at the middle of all of those operations?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -80,41 +87,49 @@ public void run() {
...
-
-        return ingestionResources.get(resourceType).nextValue();
-    }
-
     int getSize(ResourceTypes resourceType){
         return ingestionResources.containsKey(resourceType) ? ingestionResources.get(resourceType).getSize() : 0;
     }
 
-    private void addValue(String key, String value) {
+    private void addIngestionResource(HashMap<ResourceTypes, IngestionResource> ingestionResources, String key, String value) {
```
toshetah: Isn't the ingestionResources a class member? why do we need to pass it?

rabee333: Yes we have ingestionResource as a class member, but here we pass a new instance of the map to this method.
Only after we finish retrieving/parsing the values from the JSON we received from Kusto, we replace the values in the class member map.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -80,41 +87,49 @@ public void run() {
...
-            refreshIngestionResources();
-            if (!ingestionResources.containsKey(resourceType)) {
-                throw new Exception("Unable to get ingestion resources for this type: " + resourceType.getName());
-            }
-        }
-
-        return ingestionResources.get(resourceType).nextValue();
-    }
-
     int getSize(ResourceTypes resourceType){
```
tamirkamara: Looks like it's not in use. can we remove it?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -49,17 +53,20 @@ private ResourceTypes getResourceTypeByName(String name){
     private Timer timer = new Timer(true);
     private final Logger log = LoggerFactory.getLogger(ResourceManager.class);
 
+    private ReadWriteLock ingestionResourcesLock = new ReentrantReadWriteLock();
+    private ReadWriteLock authTokenLock = new ReentrantReadWriteLock();
+
     public ResourceManager(KustoClient kustoClient) throws Exception {
```
tamirkamara: Can this be lowered to package-private?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -49,17 +53,20 @@ private ResourceTypes getResourceTypeByName(String name){
     private Timer timer = new Timer(true);
```
tamirkamara: This can be moved to a local var in the method, right?

PR #28, opened 2018-10-04
Dev exception split

rabee333:
#8 Split exceptions to 2 types: client and service
Add validations to client methods


data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataClientException.java
```diff
@@ -0,0 +1,20 @@
+package com.microsoft.azure.kusto.data.exceptions;
+
+public class DataClientException extends Exception {
```
toshetah:
```suggestion
public class KustoClientException extends Exception {
```

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -30,12 +34,21 @@ public Results execute(String database, String command) throws Exception {
...
+        String jsonString;
+        try {
+            jsonString = new JSONObject()
                 .put("db", database)
                 .put("csl", command).toString();
+        } catch (JSONException e) {
+            throw new DataClientException(clusterEndpoint, String.format(clusterEndpoint, "Error in executing command: %s, in database: %s", command, database), e);
+        }
 
         return Utils.post(clusterEndpoint, aadAccessToken, jsonString);
```
toshetah: Is aadAccessToken the only header we send?
Can you please add an issue to send all relevant headers as they appear [here](https://github.com/Azure/azure-kusto-python/blob/4271142df391f526d241d96b691c3b5f3a0ed10d/azure-kusto-data/azure/kusto/data/request.py#L293)?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -36,46 +40,52 @@ static Results post(String url, String aadAccessToken, String payload) throws Ex
         httpPost.addHeader("Fed", "True");
         httpPost.addHeader("x-ms-client-version", "Kusto.Java.Client");
```
toshetah: Please also add an issue to send in that header the java client version

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -36,46 +40,52 @@ static Results post(String url, String aadAccessToken, String payload) throws Ex
...
                     }
-                    values.add(rowVector);
-                }
 
-                return new Results(columnNameToIndex, columnNameToType, values);
-            } else {
-                throw new WebException(responseContent, response);
+                    return new Results(columnNameToIndex, columnNameToType, values);
+                } else {
+                    throw new DataWebException(responseContent, response);
```
toshetah: What is DataWebException? anyway if it's a kusto error it must inherit either the client or the service exception.
I see now that you catch it and throw another exception. Can we ease that flow?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -36,46 +40,52 @@ static Results post(String url, String aadAccessToken, String payload) throws Ex
...
+                }
             }
+        } catch (DataWebException e) {
+            throw new DataServiceException(url, "Error in post request", e);
+        } catch (JSONException | IOException e) {
+            throw new DataClientException(url, "Error in post request", e);
         }
         return null;
```
toshetah: Do we ever want to return a null? why not throwing here? Do we swallow any exceptions here?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -144,27 +160,34 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
-        if (blobPath.contains(".zip") || blobPath.contains(".gz")) {
-            length = length * COMPRESSED_FILE_MULTIPLIER;
-        }
+            if (blobPath.contains(".zip") || blobPath.contains(".gz")) {
+                length = length * COMPRESSED_FILE_MULTIPLIER;
+            }
 
-        return length;
+            return length;
+        } catch (StorageException e) {
```
toshetah: IIUC, those excpetions are caught in ingestFromFile, do we have any value in catching them twice?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
rabee333: It's not the only header. Currently, we are sending those headers:

        httpPost.addHeader("Authorization", String.format("Bearer %s", aadAccessToken));
        httpPost.addHeader("Content-Type", "application/json");
        httpPost.addHeader("Accept-Encoding", "gzip,deflate");
        httpPost.addHeader("Fed", "True");
        httpPost.addHeader("x-ms-client-version", "Kusto.Java.Client");

rabee333: It's not the only header. Currently, we are sending those headers:

        httpPost.addHeader("Authorization", String.format("Bearer %s", aadAccessToken));
        httpPost.addHeader("Content-Type", "application/json");
        httpPost.addHeader("Accept-Encoding", "gzip,deflate");
        httpPost.addHeader("Fed", "True");
        httpPost.addHeader("x-ms-client-version", "Kusto.Java.Client");

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataClientException.java
rabee333: We agreed to remove "Kusto" from all class names. It is still in the package names.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
rabee333: DataWebException includes the response from the http request, and DataServiceException will wrap the DataWebException before throwing.
I moved the creation of the full exception to one place (instead of throwing and catching)

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
rabee333: in case of empty response, the entity is null, then it makes sense to return null.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
rabee333: We call this method (estimateBlobRawSize()) from ingestFromBlob().
You are right, we can just throw them and catch in ingestFromBlob().

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataClientException.java
rabee333: As above: We agreed to remove "Kusto" from all class names. It is still in the package names.

PR #37, opened 2018-10-15
fix device code auth

LizaShak:


data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -27,15 +29,21 @@
     private String userPassword;
     private String clusterUrl;
     private String aadAuthorityUri;
+    private AuthenticationType authenticationType;
 
-    AadAuthenticationHelper(ConnectionStringBuilder csb) {
-        clusterUrl = csb.getClusterUrl();
+    private enum AuthenticationType { USER, APP, DEVICE }
```
tamirkamara:
```suggestion
    private enum AuthenticationType { USER, APPLICATION, DEVICE }
```

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -73,12 +81,22 @@ AuthenticationResult acquireAccessTokenUsingDeviceCodeFlow() throws Exception {
...
+        try {
+            switch (authenticationType) {
+                case APP:
+                    return acquireAadApplicationAccessToken().getAccessToken();
+                case USER:
+                    return acquireAadUserAccessToken().getAccessToken();
+                case DEVICE:
+                    return acquireAccessTokenUsingDeviceCodeFlow().getAccessToken();
+                default:
+                    return acquireAccessTokenUsingDeviceCodeFlow().getAccessToken();
```
tamirkamara: Maybe we should throw an exception here?
or at least log.warn this since it doesn't made sense to reach the default...

data/src/main/java/com/microsoft/azure/kusto/data/ConnectionStringBuilder.java
```diff
@@ -63,4 +63,8 @@ public static ConnectionStringBuilder createWithAadApplicationCredentials(String
     {
         return createWithAadApplicationCredentials(resourceUri, applicationClientId, applicationKey, null);
     }
+
+    public static ConnectionStringBuilder createWithDeviceCredentials(String resourceUri){
```
tamirkamara: I think we should keep using the same term across. How does ADAL call this - DeviceCode? if yes, I think we should us this term everywhere...

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
LizaShak: it is not supposed to.. No way it can, I think we saw it together, that the compiler forces to use default since null can be assigned to enum..Anyway, I think I will throw exception here, since device is not really default...

data/src/main/java/com/microsoft/azure/kusto/data/ConnectionStringBuilder.java
LizaShak: Changed it

PR #38, opened 2018-10-22
Add async interface methods

rabee333:
#19 Implement async interface methods

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
@@ -27,6 +27,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 
 class IngestClientImpl implements IngestClient {
 
@@ -55,11 +56,14 @@ class IngestClientImpl implements IngestClient {
                 () -> {
                     try {
                         return ingestFromBlobImpl(blobSourceInfo, ingestionProperties);
-                    } catch (Exception e) {
+                    } catch (IngestionClientException | IngestionServiceException e) {
                         log.error("Error when ingestFromBlobAsync()", e);
-                        return null;
-                        /*todo: Might decide what to return in case of error.
-                            One suggestion is to return an object of ResultWrapper<IngestionResult, Exception>...
+                        // Here we throw a CompletionException which extends the RuntimeException.
+                        // the real exception itself would be in the <cause> of this CompletionException
+                        throw new CompletionException(e);
+                        /* We might decide what to return in case of error.
+                            One suggestion is to return an object of ResultWrapper<IngestionResult, Exception>
+                            instead of throwing an exception.
                          */
                     }
                 });
@@ -139,9 +143,9 @@ class IngestClientImpl implements IngestClient {
                 () -> {
                     try {
                         return ingestFromFileImpl(fileSourceInfo, ingestionProperties);
-                    } catch (Exception e) {
-                        log.error("Error when ingestFromFileAsync()", e);
-                        return null;
+                    } catch (IngestionClientException | IngestionServiceException e) {
+                        log.error("Error in ingestFromFileAsync()", e);
+                        throw new CompletionException(e);
                     }
                 });
     }
@@ -184,9 +188,9 @@ class IngestClientImpl implements IngestClient {
                 () -> {
                     try {
                         return ingestFromStreamImpl(streamSourceInfo, ingestionProperties);
-                    } catch (Exception e) {
+                    } catch (IngestionClientException | IngestionServiceException e) {
                         log.error("Error when ingestFromStreamAsync()", e);
-                        return null;
+                        throw new CompletionException(e);
                     }
                 });
     }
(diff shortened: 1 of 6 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -45,9 +47,33 @@
...
+
+    @Override
+    public CompletableFuture<IngestionResult> ingestFromBlobAsync(BlobSourceInfo blobSourceInfo, IngestionProperties ingestionProperties) {
+        return CompletableFuture.supplyAsync(
+                () -> {
+                    try {
+                        return ingestFromBlobImpl(blobSourceInfo, ingestionProperties);
+                    } catch (IngestionClientException | IngestionServiceException e) {
+                        log.error("Error when ingestFromBlobAsync()", e);
+                        // Here we throw a CompletionException which extends the RuntimeException.
```
limorl: please rephrase. logging the exception will log the stacktrace and it will include the method name.
Try ("Failed to ingest from blob: ", e)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -45,9 +47,33 @@
...
+                        // the real exception itself would be in the <cause> of this CompletionException
+                        throw new CompletionException(e);
+                        /* We might decide what to return in case of error.
```
limorl: This comment shouldn't be here. You should make a decision and follow it.
Assuming you want to deal with the ingestion exceptions in the caller function, the way to go is return CompletionException.
In the caller you can do:
catch(CompletionException ex) {
        try {
            throw ex.getCause();
        }
        catch(IngestionClientException | IngestionServiceException e) {
            // do something
        }
      ...
    }

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -105,7 +132,25 @@ public IngestionResult ingestFromBlob(BlobSourceInfo blobSourceInfo, IngestionPr
     }
 
     @Override
-    public IngestionResult ingestFromFile(FileSourceInfo fileSourceInfo, IngestionProperties ingestionProperties) throws IngestionClientException, IngestionServiceException {
+    public IngestionResult ingestFromFile(FileSourceInfo fileSourceInfo, IngestionProperties ingestionProperties)
+            throws IngestionClientException, IngestionServiceException {
+        return ingestFromFileImpl(fileSourceInfo, ingestionProperties);
```
limorl: Why did you use IngestFromFileImpl? You could still call IngestFromFile from the async method

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -105,7 +132,25 @@ public IngestionResult ingestFromBlob(BlobSourceInfo blobSourceInfo, IngestionPr
...
+    }
+
+    @Override
+    public CompletableFuture<IngestionResult> ingestFromFileAsync(FileSourceInfo fileSourceInfo, IngestionProperties ingestionProperties) {
+        return CompletableFuture.supplyAsync(
+                () -> {
+                    try {
+                        return ingestFromFileImpl(fileSourceInfo, ingestionProperties);
+                    } catch (IngestionClientException | IngestionServiceException e) {
+                        log.error("Error in ingestFromFileAsync()", e);
```
limorl: Same comment as above

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -134,6 +179,23 @@ public IngestionResult ingestFromFile(FileSourceInfo fileSourceInfo, IngestionPr
 
     @Override
     public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, IngestionProperties ingestionProperties) throws IngestionClientException, IngestionServiceException {
+        return ingestFromStreamImpl(streamSourceInfo, ingestionProperties);
```
limorl: same question regarding IngestFromStreamImpl

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
```diff
@@ -59,16 +58,10 @@ void tearDown() {
     @Test
     void ingestFromBlob() {
```
limorl: test name should indicate what are you testing. Each method have several tests, for happy flows, errors, etc.
You should follow java unit test naming convention.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
```diff
@@ -19,16 +19,13 @@
 class IngestClientImplTest {
```
limorl: I am missing tests here, including tests to validate all the thrown exceptions.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
rabee333: We have another branch to improve tests, including naming.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
rabee333: We have another branch for tests, and I have already added these kinds of tests.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
rabee333: Changes the exception messages in all interface methods

PR #39, opened 2018-10-22
Adding SourceInfo interface

rabee333:
This is to force implementing validate() in all sub-SourceInfo classes


tamirkamara:
Looks good

PR #41, opened 2018-10-22
Authenticate with certificate

LizaShak:


data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -148,4 +150,16 @@ AuthenticationResult acquireAccessTokenUsingDeviceCodeFlow() throws Exception {
         return result;
     }
 
+    public AuthenticationResult acquireWithClientCertificate(X509Certificate cert, PrivateKey privateKey)
+            throws IOException, InterruptedException, ExecutionException{
+
+        AuthenticationContext context = new AuthenticationContext(aadAuthorityUri, false, Executors.newFixedThreadPool(1));
```
rabee333: Instead of passing Executors.newFixedThreadPool(1) in second parameter, you might initiate in try{} block: 
service = Executors.newSingleThreadExecutor();
and then in finally shut it down: (as did in acquireAccessTokenUsingDeviceCodeFlow())
} finally {
            if (service != null) {
                service.shutdown();
            }
        }

data/src/test/java/com/microsoft/azure/kusto/data/KeyCert.java
```diff
@@ -0,0 +1,31 @@
+package com.microsoft.azure.kusto.data;
+import java.security.PrivateKey;
+import java.security.cert.X509Certificate;
+
+class KeyCert {
+
+    X509Certificate certificate;
+    PrivateKey key;
+
+    public KeyCert(X509Certificate certificate, PrivateKey key) {
```
rabee333: These publics might be package private?

Issue #18, opened 2018-09-13
App AuthN with certificate


vladikbr:
On parity with .Net, app AuthN with locally deployed cert by thumbprint and how this is possible with ADAL-java

Issue #14, opened 2018-09-05
Interactive Login option


tamirkamara:
This can be with a browser popup or other means

Issue #13, opened 2018-09-03
Re-design AzureStorageHelper


tamirkamara:
Things to think about:
1. Make methods non static
2. Keep container, queue references and not create new ones on every call (will require some kind of resource manager sync)

PR #42, opened 2018-10-24
OneWeek2018

rabee333:
Including the following items:
#18 App AuthN with certificate
#14 Interactive Login option
#13 Re-design AzureStorageHelper
#5 Support device login
And few more fixes.


data/src/test/java/com/microsoft/azure/kusto/data/AadAuthenticationHelperTest.java
```diff
@@ -0,0 +1,90 @@
...
+        X509Certificate x509Certificate = readPem(certFilePath, "basic").getCertificate();
+        PrivateKey privateKey = readPem(privateKeyPath, "basic").getKey();
+
+        ConnectionStringBuilder csb = ConnectionStringBuilder
+                .createWithAadApplicationCredentials("resource.uri", "client-id", "applicationKey", "authority-id");
+
+        AadAuthenticationHelper aadAuthenticationHelper = new AadAuthenticationHelper(csb);
+
+        Assertions.assertThrows(NullPointerException.class,
+                () -> aadAuthenticationHelper.acquireWithClientCertificate(x509Certificate, null));
```
LizaShak: add to connectionStringBuilder to verify null params

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/SourceInfo.java
```diff
@@ -1,17 +1,5 @@
...
-    public UUID getSourceId() {
-        return sourceId;
-    }
-
-    public void setSourceId(UUID sourceId) {
-        this.sourceId = sourceId;
-    }
-
+public interface SourceInfo {
+    void validate();
```
LizaShak: Guid Id

tamirkamara: @rabee333 I think the meaning here is that the interface should have the get/set for the sourceId and the abstract class will have the implementation (very basic) as well as the actual field

data/src/test/java/com/microsoft/azure/kusto/data/AadAuthenticationHelperTest.java
LizaShak: I added this validation for other functions as well.

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -22,36 +25,64 @@
...
     }
 
-    String acquireAccessToken() throws DataServiceException, DataClientException {
-        if (clientCredential != null) {
-            return acquireAadApplicationAccessToken().getAccessToken();
-        } else {
-            return acquireAadUserAccessToken().getAccessToken();
+    String acquireAccessToken() throws DataServiceException  {
+        try {
+            switch (authenticationType) {
```
toshetah: Aren't we missing here AAD_APPLICATION_CERTIFICATE?

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -95,7 +126,72 @@ private AuthenticationResult acquireAadApplicationAccessToken() throws DataServi
         return result;
     }
 
-    private Boolean isNullOrEmpty(String str) {
-        return (str == null || str.trim().isEmpty());
+    public AuthenticationResult acquireAccessTokenUsingDeviceCodeFlow() throws Exception {
```
toshetah: Can we make this method private?

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -95,7 +126,72 @@ private AuthenticationResult acquireAadApplicationAccessToken() throws DataServi
...
+        }
+        return result;
+    }
+
+    private AuthenticationResult waitAndAcquireTokenByDeviceCode(DeviceCode deviceCode, AuthenticationContext context)
+            throws  InterruptedException{
+        int timeout = 15 * 1000;
+        AuthenticationResult result = null;
+        while (timeout > 0){
+            try{
```
toshetah: IIUC, ExecutionException can be raised from the Future.get iff the context.acquireTokenByDeviceCode threw an exception. It that's the case, is there any value of retrying it?
Moreover, the get() should be blocking, so I am not sure there is any value of that busy wait.
Am I missing something?

LizaShak: This is the type of Exception thrown when it is still pending for user authentication. So the wait is not  for future result, but for user to finish the authentication.

PR #47, opened 2018-11-12
Draft: Add ByteArray for Spark support.

michazag:
Add ByteArray for Spark support.
This is a draft version for review


tamirkamara:
@michazag I think you should push this to the dev brunch not master as it is more advanced...

toshetah:
As talked F2F, there is no need in such a functionality.

PR #48, opened 2018-11-22
Update Utils.java

ohadbitt:
fix: Kusto now returns null cells which cannot be cast to string, thus failing the getString method.

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -69,7 +69,13 @@ public class Utils {
                     for (int j = 0; j < row.length(); j++) {
                         rowVector.add(row.getString(j));
                     }
-                    values.add(rowVector);
+                    Object obj = row.get(j);
+                    rowVector.add(row.getString(j));
+	                if(obj == JSONObject.NULL){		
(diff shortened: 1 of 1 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -69,7 +69,13 @@ public static KustoResults post(String url, String aadAccessToken, String payloa
                     for (int j = 0; j < row.length(); j++) {
                         rowVector.add(row.getString(j));
```
toshetah: I don't understand. Isn't this the problematic line?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -69,7 +69,13 @@ public static KustoResults post(String url, String aadAccessToken, String payloa
...
                     }
-                    values.add(rowVector);
+                    Object obj = row.get(j);
+                    rowVector.add(row.getString(j));
+	                if(obj == JSONObject.NULL){		
+	                    rowVector.add("");
```
toshetah: Why adding empty string when the value is null?

toshetah:
Can you please merge it into dev, and not into master?

PR #54, opened 2018-12-06
Samples update

tamirkamara:


toshetah:
I would suggest the readme.md to be aligned with our other clients readmes.
You can see examples at:
https://github.com/Azure/azure-kusto-node
https://github.com/Azure/azure-kusto-python

PR #55, opened 2018-12-10
Prepare version 1.0.0-BETA-01

tamirkamara:
A bunch of things here:
- Refactor data module
- Add options to authenticate interactively and with a certificate
- Dedicated exceptions
- Readme/samples updates
- Adjustments in POMs and project structure


data/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>kusto-client</artifactId>
         <groupId>com.microsoft.azure.kusto</groupId>
-        <version>0.2.0</version>
+        <version>1.0.0-BETA-01</version>
```
danield137: semver -> 1.0.0-beta.1 not, beta-01 https://semver.org/spec/v2.0.0.html#spec-item-11

also, maybe rc is more appropriate, it's not really a beat...

tamirkamara: [This ](https://github.com/Azure/adx-documentation-pr/blob/master/sdks/guidelines/Azure_Java_libraries_guidelines.md) only has the beta option.

01 instead of just 1 is to have the entire version string ordered the right way

Issue #40, opened 2018-10-22
Add/modify headers in post() method


rabee333:
Add/modify headers in post() method in Utils class, to include the headers as they appear [here](https://github.com/Azure/azure-kusto-python/blob/4271142df391f526d241d96b691c3b5f3a0ed10d/azure-kusto-data/azure/kusto/data/request.py#L293).

PR #57, opened 2018-12-11
Feature/client request properties

danield137:
Exposing ClientRequestProperties
* new class `ClientRequestProperties` that will contain consts / utility methods as time goes on, currently exposes timeout
* added timeout functionality on HttpRequest
* added tests + docs

### DIFF
--- a/data/pom.xml
+++ b/data/pom.xml
@@ -98,7 +98,12 @@
             <version>${mockito.version}</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.skyscreamer</groupId>
+            <artifactId>jsonassert</artifactId>
+            <version>1.5.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 
--- a/data/pom.xml
+++ b/data/pom.xml
@@ -15,6 +15,18 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                            <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
+                        </manifest>
+                    </archive>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${maven-compiler-plugin.version}</version>
                 <configuration>
--- a/data/pom.xml
+++ b/data/pom.xml
@@ -27,6 +27,7 @@
                 <configuration>
                     <archive>
                         <manifest>
+                            <mainClass>com.microsoft.azure.kusto.data.Client</mainClass>
                             <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
                             <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
                         </manifest>
--- a/data/pom.xml
+++ b/data/pom.xml
@@ -15,18 +15,6 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-jar-plugin</artifactId>
-                <configuration>
-                    <archive>
-                        <manifest>
-                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
-                            <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
-                        </manifest>
-                    </archive>
-                </configuration>
-            </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${maven-compiler-plugin.version}</version>
                 <configuration>
@@ -39,7 +27,8 @@
                 <configuration>
                     <archive>
                         <manifest>
-                            <mainClass>com.microsoft.azure.kusto.ingest.Client</mainClass>
+                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                            <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
                         </manifest>
                     </archive>
                     <descriptorRefs>
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -1,4 +1,5 @@
 package com.microsoft.azure.kusto.data;
+
 import com.microsoft.azure.kusto.data.exceptions.DataClientException;
 import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
 import org.apache.commons.lang3.StringUtils;
@@ -29,12 +30,12 @@ public class ClientImpl implements Client {
     }
 
     public Results execute(String database, String command) throws DataServiceException, DataClientException {
-        return execute(database,command, null);
+        return execute(database, command, null);
     }
 
     public Results execute(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException {
         // Argument validation:
-        if(StringUtils.isAnyEmpty(database, command)){
+        if (StringUtils.isAnyEmpty(database, command)) {
             throw new IllegalArgumentException("database or command are empty");
         }
 
@@ -61,8 +62,8 @@ public class ClientImpl implements Client {
         String jsonString;
         try {
             JSONObject json = new JSONObject()
-                .put("db", database)
-                .put("csl", command);
+                    .put("db", database)
+                    .put("csl", command);
 
             if (properties != null) {
                 json.put("properties", properties.toString());
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -1,5 +1,4 @@
 package com.microsoft.azure.kusto.data;
-
 import com.microsoft.azure.kusto.data.exceptions.DataClientException;
 import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
 import org.apache.commons.lang3.StringUtils;
@@ -7,6 +6,7 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.net.URISyntaxException;
+import java.util.concurrent.TimeUnit;
 
 public class ClientImpl implements Client {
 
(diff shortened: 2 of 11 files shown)


data/pom.xml
```diff
@@ -98,7 +98,12 @@
             <version>${mockito.version}</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
```
tamirkamara: Can't find where you're using this...

data/src/test/java/com/microsoft/azure/kusto/data/ClientRequestPropertiesTest.java
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.skyscreamer.jsonassert.JSONAssert;
+
+import java.util.concurrent.TimeUnit;
+
+public class ClientRequestPropertiesTest {
+    @Test
+    @DisplayName("test set/get timeout")
+    void testSetGetTimeout() {
```
tamirkamara: I'd avoid using the test prefix - everything in this class are tests, and also have the "@test" notation...
maybe name it timeout_SetGet

data/src/test/java/com/microsoft/azure/kusto/data/ClientRequestPropertiesTest.java
```diff
@@ -0,0 +1,34 @@
...
+        // before setting value should be null
+        Assertions.assertEquals(null, props.getTimeout());
+
+        props.setTimeout(expected);
+        Assertions.assertEquals(expected, props.getTimeout());
+    }
+
+    @Test
+    @DisplayName("test set/get timeout")
+    void testToString() throws JSONException {
```
tamirkamara: I'd avoid using the test prefix - everything in this class are tests, and also have the "@test" notation...
maybe name it timeout_ToString

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -27,31 +27,50 @@ public Results execute(String command) throws DataServiceException, DataClientEx
...
+
+        if (properties != null) {
+            timeoutMs = properties.getTimeout();
+        }
+
         String clusterEndpoint;
         if (command.startsWith(ADMIN_COMMANDS_PREFIX)) {
             clusterEndpoint = String.format("%s/%s/rest/mgmt", clusterUrl, API_VERSION);
+            if (timeoutMs == null) {
+                timeoutMs = TimeUnit.HOURS.toMillis(1) + TimeUnit.SECONDS.toMillis(30);
```
tamirkamara: Maybe use the time from a const?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -27,31 +27,50 @@ public Results execute(String command) throws DataServiceException, DataClientEx
...
-        }
-        return execute(database, command, clusterEndpoint);
-    }
-
-    private Results execute(String database, String command, String clusterEndpoint) throws DataServiceException, DataClientException {
-        // Argument validation:
-        if (StringUtils.isAnyEmpty(database, command, clusterEndpoint)) {
-            throw new IllegalArgumentException("database, command or clusterEndpoint are empty");
+            if (timeoutMs == null) {
+                timeoutMs = TimeUnit.MINUTES.toMillis(4) + TimeUnit.SECONDS.toMillis(30);
```
tamirkamara: Maybe use the time from a const?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -22,8 +24,16 @@
 
 class Utils {
 
-    static Results post(String url, String aadAccessToken, String payload) throws DataServiceException, DataClientException {
-        HttpClient httpClient = HttpClients.createSystem();
+    static Results post(String url, String aadAccessToken, String payload, Long timeoutMs) throws DataServiceException, DataClientException {
+
+        HttpClient httpClient;
+        if (timeoutMs != null) {
+            RequestConfig requestConfig = RequestConfig.custom().setSocketTimeout(timeoutMs.intValue()).build();
```
tamirkamara: I get that the socket timeout doesn't allow longs, so maybe we just use int from the start?

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public void removeOption(String name) {
+        ((HashMap<String, Object>)properties.get(OPTIONS_KEY)).remove(name);
+    }
+
+    public void clearOptions() {
+        ((HashMap<String, Object>)properties.get(OPTIONS_KEY)).clear();
+    }
+
+    public void setTimeout(Long timeoutInMs) {
```
tamirkamara: I think that we should be explicit here to help the user understand the units - name it setTimeoutInMilliSecs.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -0,0 +1,55 @@
...
+        ((HashMap<String, Object>)properties.get(OPTIONS_KEY)).put(OptionServerTimeout, timeoutInMs);
+    }
+
+    public Long getTimeout() {
```
tamirkamara: rename according to the setter

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -0,0 +1,55 @@
...
+ * Kusto supports attaching various properties to client requests (such as queries and control commands).
+ * Such properties may be used to provide additional information to Kusto (for example, for the purpose of correlating client/service interaction),
+ * may affect what limits and policies get applied to the request, and much more.
+ * For a complete list of available client request properties
+ * check out https://docs.microsoft.com/en-us/azure/kusto/api/netfx/request-properties#list-of-clientrequestproperties
+ */
+public class ClientRequestProperties {
+    private HashMap<String, Object> properties;
+    private static final String OPTIONS_KEY = "Options";
+    private static final String OptionServerTimeout  = "servertimeout";
```
tamirkamara: I think this var should also be all capitals, no?

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.HashMap;
+
+/*
+ * Kusto supports attaching various properties to client requests (such as queries and control commands).
+ * Such properties may be used to provide additional information to Kusto (for example, for the purpose of correlating client/service interaction),
+ * may affect what limits and policies get applied to the request, and much more.
+ * For a complete list of available client request properties
+ * check out https://docs.microsoft.com/en-us/azure/kusto/api/netfx/request-properties#list-of-clientrequestproperties
+ */
+public class ClientRequestProperties {
```
tamirkamara: This class generates many access warnings on the public methods, do we need all of them?
for example - the toJson is used internally, isn't it?

tamirkamara: There're unchecked conversions from Object to HashMap. I know we have that elsewhere, but this is a new class and probably there's a better way to deal with it...
One thing I don't understand about the usage of the HashMap - you have specific key strings defined as finals (inside the options key), meaning that you can do something similar with a well-defined class for the options. can't you?
Something like:
Class ClientRequestPropertiesOptions {
 private int serverTimeoutInMilliSecs;
 private int someOtherOptionKey;
 getters/setters
}

data/pom.xml
danield137: https://github.com/Azure/azure-kusto-java/pull/57/files#diff-dfdcac39245e8c607ad8d656e094db82R32

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
danield137: problem is TimeUnit is long, so I either do the conversion here, or on the caller. 
would you prefer it in the caller?

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
danield137: not sure I understand what you mean..
I only explicitly provided 1 valid option, but there are others many other options.
I want to allow a somewhat dynamic way to manipulate those options, without making code cumbersome.

If you feel there is a better way to do so, please provide a code sample as I couldn't quite get what you meant above 👼

danield137: public is not bad 😄  if you feel any of them should be private, I will change, but client might want to use any of those

danield137:
@tamirkamara fixed most of the comments.
please go over it again when you can 🙏

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
tamirkamara: Why would I (as a developer) would need it in a json format?

tamirkamara: I missed that there're methods that deal with an option with its name as an input.
I suggest to simplify the handling of the hashmaps with something like this:
`
public class ClientRequestProperties {
    private static final String OPTIONS_KEY = "Options";
    private static final String OPTION_SERVER_TIMEOUT = "servertimeout";

    private HashMap<String, Object> properties;
    private HashMap<String, Object> options;

    public ClientRequestProperties() {
        properties = new HashMap<>();
        options = new HashMap<>();
        properties.put(OPTIONS_KEY, options);
    }

    public void setOption(String name, Object value) {
        options.put(name, value);
    }

    public Object getOption(String name) {
        return options.get(name);
    }

    public void removeOption(String name) {
        options.remove(name);
    }

    public void clearOptions() {
        options.clear();
    }

        return (Long) getOption(OPTION_SERVER_TIMEOUT);
    }

    public void setTimeoutInMilliSec(Long timeoutInMs) {
        setOption(OPTION_SERVER_TIMEOUT, timeoutInMs);
    }

    public JSONObject toJson() {
        return new JSONObject(properties);
    }

    public String toString() {
        return toJson().toString();
    }
}
`

data/src/test/java/com/microsoft/azure/kusto/data/ClientRequestPropertiesTest.java
tamirkamara: The link is explaining display names, not overall naming convention. I remember us checking this in the past and reached the conclusion that it is redundant. In this case for JUnit it isn't required in the method names...
https://dzone.com/articles/7-popular-unit-test-naming
https://www.petrikainulainen.net/programming/testing/writing-clean-tests-naming-matters/

I don't really mind one way or the other, but we didn't do this until now.

tamirkamara:
also, this PR should addresses #40, right?
If yes, maybe include that comment in the PR description...

PR #58, opened 2018-12-12
Ingest from a ResultSet implementation

### LINKED ISSUE #2: Add support for ResultSet ingestion

tamirkamara:
This closes #2


tamirkamara:
Please don't merge this (I'll do it myself)

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
```diff
@@ -107,12 +116,127 @@ void ingestFromStream() {
...
+                () -> ingestClientSpy.ingestFromResultSet(resultSetSourceInfo, ingestionProperties));
+    }
+
+    @Test
+    void resultSetToCsv_Success() throws SQLException, IngestionClientException {
+        IngestClientImpl ingestClient = new IngestClientImpl(resourceManagerMock, azureStorageHelperMock);
+        ResultSet resultSet = getSampleResultSet();
+        StringWriter stringWriter = new StringWriter();
+
+        long numberOfCharsActual = ingestClient.resultSetToCsv(resultSet, stringWriter, false);
```
limorl: Since you are using a sample result set using getSampleResultSet, it's better to generate the expected string the same method:
ResultSet getSampleResultSet(StringBuilder sb) 
--> and then you write the expected string based on how you generated the result test using the StringBuilder

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
```diff
@@ -107,12 +116,127 @@ void ingestFromStream() {
...
+        final String expected = System.getProperty("line.separator").equals("\n") ?
+                "\"1\",\"leo\"\n\"2\",\"yui\"\n" :
+                "\"1\",\"leo\"\r\n\"2\",\"yui\"\r\n";
+
+        assertEquals(expected, stringWriter.toString()); // check the string values
+        assertEquals(expected.length(), numberOfCharsActual); // check the returned length
+    }
+
+    @Test
+    void resultSetToCsv_ResultSetClosed_Exception() throws IngestionClientException, IngestionServiceException, SQLException {
```
limorl: Is ResultSetClosed the only case when an exception is closed?

tamirkamara: Added another one for the writer

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -204,8 +222,124 @@ private String genBlobName(String fileName, String databaseName, String tableNam
...
+
+    private int writeResultSetRow(ResultSet resultSet, Writer writer, int numberOfColumns) throws IOException, SQLException {
+        int numberOfChars = 0;
+        String columnString;
+        String columnSeparator = "";
+
+        for (int i = 1; i <= numberOfColumns; i++) {
+            writer.write(columnSeparator);
+            writer.write('"');
+            columnString = resultSet.getObject(i).toString();
```
toshetah: Will that work with new lines?
Will that work with quotes in the columnString?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -204,8 +222,124 @@ private String genBlobName(String fileName, String databaseName, String tableNam
...
+            }
+        }
+    }
+
+    private int writeResultSetRow(ResultSet resultSet, Writer writer, int numberOfColumns) throws IOException, SQLException {
+        int numberOfChars = 0;
+        String columnString;
+        String columnSeparator = "";
+
+        for (int i = 1; i <= numberOfColumns; i++) {
```
toshetah: Why 1? what do we have in the first column?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -204,8 +222,124 @@ private String genBlobName(String fileName, String databaseName, String tableNam
...
+            writer.write(columnSeparator);
+            writer.write('"');
+            columnString = resultSet.getObject(i).toString();
+            writer.write(columnString);
+            writer.write('"');
+
+            columnSeparator = ",";
+            numberOfChars += columnString.length();
+        }
+
```
toshetah: Maybe we want to calculate and return the correct number? instead of calculating in the end.

data/src/test/resources/simplelogger.properties
```diff
@@ -0,0 +1,7 @@
+# SLF4J's SimpleLogger configuration file
```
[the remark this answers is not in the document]
tamirkamara: yes, it makes sure that when testing (travis) all logger will be in debug mode and we'll have an easier time checking failures

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -38,16 +44,28 @@
...
+        this.azureStorageHelper = new AzureStorageHelper();
+    }
+
+    IngestClientImpl(ResourceManager resourceManager) {
+        log.info("Creating a new IngestClient");
+        this.resourceManager = resourceManager;
         azureStorageHelper = new AzureStorageHelper();
     }
 
+    IngestClientImpl(ResourceManager resourceManager, AzureStorageHelper azureStorageHelper) {
```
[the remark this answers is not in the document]
tamirkamara: For testing purposes we need the ability to pass the client mocks (or spys) of those two classes.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
tamirkamara: JDBC ResultSet is not a 0 based, so column 0 is always empty

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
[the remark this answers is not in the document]
toshetah: Can you please open the following issue then:
Add PersistentStorageFactoryFactory that will know how to resolve requests according to the given uris.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
tamirkamara: I added a fix to qualify the text qualifier string - replace one " with two

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
tamirkamara: I believe the recent change makes it easier to understand

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -204,8 +222,124 @@ private String genBlobName(String fileName, String databaseName, String tableNam
...
+     */
     @Override
-    public IngestionResult ingestFromResultSet(ResultSetSourceInfo resultSetSourceInfo, IngestionProperties ingestionProperties) {
-        throw new UnsupportedOperationException();
+    public IngestionResult ingestFromResultSet(ResultSetSourceInfo resultSetSourceInfo, IngestionProperties ingestionProperties) throws IngestionClientException, IngestionServiceException {
+        try {
+            Objects.requireNonNull(resultSetSourceInfo, "resultSetSourceInfo cannot be null");
+            resultSetSourceInfo.validate();
+
+            File tempFile = File.createTempFile("kusto-resultset", ".csv.gz");
```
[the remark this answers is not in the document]
tamirkamara: Since we don't know the size of the resultset we can't know if we'll run out of space.
I've added an overload that enables the user to specify the temp path.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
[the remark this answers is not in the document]
tamirkamara: Issue is #59

PR #64, opened 2019-01-25
add sample code for blob ingestion

leoninez:
Add sample code to ingest blob json file into Kusto.


tamirkamara:
@leoninez can you please change the input parameters so that it would be possible to get from the command line?
you can see an example on the other samples..

PR #70, opened 2019-02-12
cache-lastAuthenticationResult

ohadbitt:
cache-lastAuthenticationResult


data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -29,6 +30,7 @@
     private X509Certificate x509Certificate;
     private PrivateKey privateKey;
     private AuthenticationType authenticationType;
+    private AuthenticationResult lastAuthenticationResult;
```
toshetah: Can you please hold an instance of AuthorizationContext as a class member, and not the result?
We do the same in .Net, python, and node.
See an example here:
https://github.com/Azure/azure-kusto-python/blob/master/azure-kusto-data/azure/kusto/data/security.py

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -62,23 +64,30 @@ public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URIS
     }
 
     String acquireAccessToken() throws DataServiceException  {
+        if(lastAuthenticationResult != null && lastAuthenticationResult.getExpiresOnDate().after(new Date())){
```
tamirkamara: What will happen if we have a really close expiration (but still valid) and by the time we issue the call to the service the token will expire?
Isn't it safer to have a buffer of a few seconds or a minute from the actual expiration to trigger a refresh of the cached one?

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -62,23 +64,30 @@ public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URIS
...
+            return lastAuthenticationResult.getAccessToken();
+        }
+
         try {
             switch (authenticationType) {
                 case AAD_APPLICATION_KEY:
-                    return acquireAadApplicationAccessToken().getAccessToken();
+                    lastAuthenticationResult = acquireAadApplicationAccessToken();
+                    return lastAuthenticationResult.getAccessToken();
```
tamirkamara: Consider combing all return statements in the switch block and have just one after it

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
ohadbitt: will do 1 minute

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -62,23 +65,29 @@ public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URIS
...
-                    return acquireAadApplicationAccessToken().getAccessToken();
-                case AAD_USERNAME_PASSWORD:
-                    return acquireAadUserAccessToken().getAccessToken();
-                case AAD_DEVICE_LOGIN:
-                    return acquireAccessTokenUsingDeviceCodeFlow().getAccessToken();
-                case AAD_APPLICATION_CERTIFICATE:
-                    return acquireWithClientCertificate().getAccessToken();
-                default:
-                    throw new DataServiceException("Authentication type: " + authenticationType.name() + " is invalid");
+        if(lastAuthenticationResult == null || lastAuthenticationResult.getExpiresOnDate().before(new Date(System.currentTimeMillis() + ONE_MINUTE_IN_MILLIS))) {
```
toshetah: What we actually need to do in case of existing and nearly expired tokens, is acquire a new token, using [acquireTokenByRefreshToken](https://github.com/Microsoft/azure-activedirectory-interactive-auth-library-for-java/blob/36aed916659ae176245129ebb10d4a87f6966d37/src/com/microsoftopentechnologies/auth/AuthenticationContext.java#L200)

tamirkamara:
I know this module is lite on test but I think that a test to validate the how the cached version works needs to be added

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -29,6 +31,9 @@
     private X509Certificate x509Certificate;
     private PrivateKey privateKey;
     private AuthenticationType authenticationType;
+    private AuthenticationResult lastAuthenticationResult;
+    ExecutorService service = Executors.newSingleThreadExecutor();
+    AuthenticationContext context;
```
toshetah: Did you test it in multiple threads environment?
I am not sure what will happen if multiple threads will share this service.

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -50,7 +55,7 @@ public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URIS
         } else if (csb.getX509Certificate() != null && csb.getPrivateKey() != null) {
             x509Certificate = csb.getX509Certificate();
             privateKey = csb.getPrivateKey();
-            clientCredential = new ClientCredential(csb.getApplicationClientId(), null);
+            clientCredential = new ClientCredential(csb.getApplicationClientId(), "null");
```
toshetah: Why "null"?

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -61,44 +66,51 @@ public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URIS
...
-                case AAD_DEVICE_LOGIN:
-                    return acquireAccessTokenUsingDeviceCodeFlow().getAccessToken();
-                case AAD_APPLICATION_CERTIFICATE:
-                    return acquireWithClientCertificate().getAccessToken();
-                default:
-                    throw new DataServiceException("Authentication type: " + authenticationType.name() + " is invalid");
+    String acquireAccessToken() throws DataServiceException {
+        boolean isAccessTokenExpired;
+        if (lastAuthenticationResult == null || ((isAccessTokenExpired = lastAuthenticationResult.getExpiresOnDate().before(dateInAMinute()))
+                                                                                                                    && lastAuthenticationResult.getRefreshToken() == null)) {
```
toshetah: Can you please fix the indentation here?

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -61,44 +66,51 @@ public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URIS
...
-                    return acquireAadUserAccessToken().getAccessToken();
-                case AAD_DEVICE_LOGIN:
-                    return acquireAccessTokenUsingDeviceCodeFlow().getAccessToken();
-                case AAD_APPLICATION_CERTIFICATE:
-                    return acquireWithClientCertificate().getAccessToken();
-                default:
-                    throw new DataServiceException("Authentication type: " + authenticationType.name() + " is invalid");
+    String acquireAccessToken() throws DataServiceException {
+        boolean isAccessTokenExpired;
+        if (lastAuthenticationResult == null || ((isAccessTokenExpired = lastAuthenticationResult.getExpiresOnDate().before(dateInAMinute()))
```
toshetah: The logic is unclear. Can you please separate the assignment and the checks?

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -107,21 +111,17 @@ private AuthenticationResult acquireAadUserAccessToken() throws DataServiceExcep
         return result;
     }
 
-    private AuthenticationResult acquireAadApplicationAccessToken() throws DataServiceException, DataClientException {
-        AuthenticationContext context;
+    private AuthenticationResult acquireAadApplicationAccessToken() throws DataServiceException, DataClientException, MalformedURLException {
+        if (context == null) {
```
toshetah: You can create the context in the constructor, instead of checking nulls.

toshetah: After changing this, please rerun the concurrent tests.

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -8,18 +8,21 @@
...
+import java.util.Date;
 import java.util.concurrent.*;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 public class AadAuthenticationHelper {
 
     private final static String DEFAULT_AAD_TENANT = "common";
     private final static String CLIENT_ID = "db662dc1-0cfe-4e1c-a843-19a68e65be58";
+    public final static long ONE_MINUTE_IN_MILLIS = 60000;
```
tamirkamara:
```suggestion
    final static long ONE_MINUTE_IN_MILLIS = 60000;
```

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -8,18 +8,21 @@
...
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.security.PrivateKey;
 import java.security.cert.X509Certificate;
+import java.util.Date;
 import java.util.concurrent.*;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 public class AadAuthenticationHelper {
```
tamirkamara:
```suggestion
class AadAuthenticationHelper {
```

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -38,6 +44,7 @@
     }
 
     public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URISyntaxException {
```
tamirkamara:
```suggestion
    AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URISyntaxException {
```

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -158,44 +165,107 @@ private AuthenticationResult acquireAccessTokenUsingDeviceCodeFlow() throws Exce
...
+            }
+        }
+        lastAuthenticationResultLock.unlock();
+    }
+
+    private boolean isTokenExpired() {
+        return lastAuthenticationResult.getExpiresOnDate().before(dateInAMinute());
+    }
+
+    AuthenticationResult refreshToken() throws DataServiceException {
```
tamirkamara: The name is confusing in my opinion - maybe **acquireAccessTokenByRefreshToken** ?

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -158,44 +165,107 @@ private AuthenticationResult acquireAccessTokenUsingDeviceCodeFlow() throws Exce
...
             }
         }
+
         if (result == null) {
             throw new ServiceUnavailableException("authentication result was null");
         }
         return result;
     }
 
+    private void acquireToken() throws DataServiceException {
```
tamirkamara: maybe rename to **acquireAccessToken** to better differentiate from the refresh token part?

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
tamirkamara: Well, thinking about this again, I think you should rename this according to the usage, something like:
MIN_ACCESS_TOKEN_VALIDITY_IN_MILLISECS

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
ohadbitt: there is already an acquireAccessToken

data/src/test/java/com/microsoft/azure/kusto/data/AadAuthenticationHelperTest.java
```diff
@@ -83,4 +96,37 @@ static KeyCert readPem(String path, String password)
...
+        AuthenticationResult authenticationResult = new AuthenticationResult("testType", "firstToken","refreshToken",0,"id", mock(UserInfo.class),false);
+        AuthenticationResult authenticationResultFromRefresh = new AuthenticationResult("testType", "fromRefresh",null,90,"id", mock(UserInfo.class),false);
+        AuthenticationResult authenticationResultNullRefreshTokenResult = new AuthenticationResult("testType", "nullRefreshResult",null,0,"id", mock(UserInfo.class),false);
+
+        doReturn(authenticationResultFromRefresh).when(aadAuthenticationHelperSpy).acquireAccessTokenByRefreshToken();
+        doReturn(authenticationResult).when(aadAuthenticationHelperSpy).acquireWithClientCertificate();
+
+        assertEquals("firstToken", aadAuthenticationHelperSpy.acquireAccessToken());
+        assertEquals("fromRefresh", aadAuthenticationHelperSpy.acquireAccessToken());
+        assertEquals("fromRefresh", aadAuthenticationHelperSpy.acquireAccessToken());
```
rabee333: Cannot this test be split to 2 tests?
if one of these assertEquals() fails, it won't continue to the last assertEquals().

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -67,24 +74,27 @@ public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URIS
...
+
+        if (lastAuthenticationResult == null) {
+            acquireToken();
+        } else {
+            if (isTokenExpired()) {
+                if (lastAuthenticationResult.getRefreshToken() == null) {
+                    acquireToken();
+                } else {
+                    lastAuthenticationResultLock.lock();
+                    if (isTokenExpired()) {
```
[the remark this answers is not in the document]
ohadbitt: its how we do in kusto locks in multithreaded environment - if one thread got to the lock already and acquired the token - there is no need for all the others who waited on the lock to acquire it too

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -67,24 +74,27 @@ public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URIS
...
+        }
+
+        if (lastAuthenticationResult == null) {
+            acquireToken();
+        } else {
+            if (isTokenExpired()) {
+                if (lastAuthenticationResult.getRefreshToken() == null) {
+                    acquireToken();
+                } else {
+                    lastAuthenticationResultLock.lock();
```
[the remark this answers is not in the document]
ohadbitt: i can't as i need acquireAccessTokenByRefreshToken to have a return value for the testing

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -204,4 +216,64 @@ AuthenticationResult acquireWithClientCertificate()
         return result;
     }
 
+    private void acquireToken() throws DataServiceException {
+        lastAuthenticationResultLock.lock();
```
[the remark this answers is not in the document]
ohadbitt: same as before

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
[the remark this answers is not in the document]
rabee333: you still can add the lock() to the beginning of the 'try' block, and the unlock() to the 'finally' block.

data/src/test/java/com/microsoft/azure/kusto/data/AadAuthenticationHelperTest.java
ohadbitt: i dont think i should. the test checks the flow - for the first time it should return a token from the regular acquireToken function. than it will refresh it as the last token is expired and will . Now it validates that calling the acquireAccessToken still gets the second token .
And the last assert checks that now (after time passed) - if token expires it will reacquire a token using the regular acquireToken function again.

but should i add comments ?

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -204,4 +216,64 @@ AuthenticationResult acquireWithClientCertificate()
...
+                case AAD_DEVICE_LOGIN:
+                    return context.acquireTokenByRefreshToken(lastAuthenticationResult.getRefreshToken(), CLIENT_ID, clusterUrl, null).get();
+                default:
+                    throw new DataServiceException("Authentication type: " + authenticationType.name() + " is invalid");
+            }
+        } catch (Exception e) {
+            throw new DataServiceException(e.getMessage());
+        } finally {
+            if (service != null) {
+                service.shutdown();
```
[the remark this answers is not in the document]
ohadbitt: we wait on the get method so the task is finished or canceled

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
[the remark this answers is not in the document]
rabee333: But if the lastAuthenticationResult is not null and not expired yet, you don't need the lock()

data/src/test/java/com/microsoft/azure/kusto/data/AadAuthenticationHelperTest.java
rabee333: OK I understand now.. you might add some comments.

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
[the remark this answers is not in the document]
rabee333: OK, Now I see since you need the lock() outside (from the above comment)

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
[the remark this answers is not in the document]
rabee333: Now I see in multi threaded environment, the lock() should be outside the if statement.

PR #71, opened 2019-02-12
authenticate-from-token

ohadbitt:
authenticate-from-token

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
@@ -36,7 +36,7 @@ public class AadAuthenticationHelper {
         AAD_APPLICATION_KEY,
         AAD_DEVICE_LOGIN,
         AAD_APPLICATION_CERTIFICATE,
-        AAD_USER_TOKEN
+        AAD_ACCESS_TOKEN
     }
 
     public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URISyntaxException {
@@ -55,7 +55,7 @@ public class AadAuthenticationHelper {
             clientCredential = new ClientCredential(csb.getApplicationClientId(), null);
             authenticationType = AuthenticationType.AAD_APPLICATION_CERTIFICATE;
         } else if (StringUtils.isNotBlank(csb.getAccessToken())) {
-            authenticationType = AuthenticationType.AAD_USER_TOKEN;
+            authenticationType = AuthenticationType.AAD_ACCESS_TOKEN;
             accessToken = csb.getAccessToken();
         } else {
             authenticationType = AuthenticationType.AAD_DEVICE_LOGIN;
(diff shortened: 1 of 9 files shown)


data/src/test/java/com/microsoft/azure/kusto/data/ConnectionStringBuilderTest.java
```diff
@@ -118,4 +118,22 @@ void createWithAadApplicationCertificate() throws CertificateException, Operator
...
+                () -> ConnectionStringBuilder
+                        .createWithAadUserTokenAuthentication(null, "token"));
+        Assertions.assertThrows(IllegalArgumentException.class,
+                () -> ConnectionStringBuilder
+                        .createWithAadUserTokenAuthentication("","token"));
+        Assertions.assertThrows(IllegalArgumentException.class,
+                () -> ConnectionStringBuilder
+                        .createWithAadUserTokenAuthentication("resource.uri", null));
+        Assertions.assertThrows(IllegalArgumentException.class,
+                () -> ConnectionStringBuilder
```
toshetah: Can you please add a positive test, where we initiate the CSB with valid token, and we get it when we try?

ohadbitt: we dont do it anywhere else but sure i can

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -52,6 +54,9 @@ public AadAuthenticationHelper(@NotNull ConnectionStringBuilder csb) throws URIS
             privateKey = csb.getPrivateKey();
             clientCredential = new ClientCredential(csb.getApplicationClientId(), null);
             authenticationType = AuthenticationType.AAD_APPLICATION_CERTIFICATE;
+        } else if (StringUtils.isNotEmpty(csb.getUserToken())){
```
tamirkamara: I think this should be blank (not empty)

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -29,12 +29,14 @@
     private X509Certificate x509Certificate;
     private PrivateKey privateKey;
     private AuthenticationType authenticationType;
+    private String userToken;
```
tamirkamara: I think this should be renamed across all files to **accessToken**

ohadbitt: i used the variable name from our .Net library

tamirkamara: Unless I'm missing something I think that name isn't right on the other implementation.
Look at AadAuthenticationHelper.acquireAccessToken - userToken is actually returned and used as access token.
You can confirm with @toshetah to make sure.

toshetah: The userToken is defined in the connection string builder.
In the AadHelpers it is accessToken.

data/src/test/java/com/microsoft/azure/kusto/data/ConnectionStringBuilderTest.java
toshetah: I meant more like Assertions.Equels(ConnectionStringBuilder           .createWithAadAccessTokenAuthentication("resource.uri","token").getAccessToken(), "token")

PR #82, opened 2019-04-18
ingestWithTagsPropeties

ohadbitt:
ingestWithTagsPropeties


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -1,5 +1,9 @@
 package com.microsoft.azure.kusto.ingest;
 
+import org.codehaus.jackson.map.ObjectMapper;
```
tamirkamara: Currently this is available through another dependency but should be included in the POM explicitly to avoid problems in the future.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -33,7 +42,54 @@ public IngestionReportMethod getReportMethod() {
...
+    public ArrayList<String> getAdditionalTags() {
+        return additionalTags;
+    }
+
+    public ArrayList<String> getIngestIfNotExists() {
+        return ingestIfNotExists;
+    }
+
+    public Map<String, String> getAdditionalProperties() throws IOException {
+        if(!dropByTags.isEmpty() || !ingestByTags.isEmpty() || !additionalTags.isEmpty()){
```
tamirkamara: The value is kept in additionalProperties even between get calls. What will happen if all those are empty - looks like the old "tags" will still be there, right?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -33,7 +42,54 @@ public IngestionReportMethod getReportMethod() {
...
+                    tags.add(String.format("%s%s", "ingest-by:", t));
+                }
+            }
+
+            ObjectMapper objectMapper = new ObjectMapper();
+            String tagsAsJson = objectMapper.writeValueAsString(tags);
+            additionalProperties.put("tags", tagsAsJson);
+        }
+
+        if (!ingestIfNotExists.isEmpty())
```
tamirkamara: Like above... Looks like there's no way to "clear" old keys from the map when the array list is empty.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -33,7 +42,54 @@ public IngestionReportMethod getReportMethod() {
...
+        {
+            ObjectMapper objectMapper = new ObjectMapper();
+            String ingestIfNotExistsJson = objectMapper.writeValueAsString(ingestIfNotExists);
+            additionalProperties.put("ingestIfNotExists", ingestIfNotExistsJson );
+        }
+
         return additionalProperties;
```
tamirkamara: I wonder - if the get method calculates the value every time from scratch and there's no direct set method, is there're a point in saving it?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -99,6 +155,19 @@ public void setAuthorizationContextToken(String token) {
         additionalProperties.put("authorizationContext", token);
     }
 
+    public void setDropByTags(ArrayList<String> dropByTags) {
```
tamirkamara: missing javadoc

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -99,6 +155,19 @@ public void setAuthorizationContextToken(String token) {
...
+        this.dropByTags = dropByTags;
+    }
+    public void setIngestByTags(ArrayList<String> ingestByTags) {
```
tamirkamara: missing javadoc

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -99,6 +155,19 @@ public void setAuthorizationContextToken(String token) {
...
+        this.ingestByTags = ingestByTags;
+    }
+    public void setAdditionalTags(ArrayList<String> additionalTags) {
```
tamirkamara: missing javadoc

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -99,6 +155,19 @@ public void setAuthorizationContextToken(String token) {
...
+        this.additionalTags = additionalTags;
+    }
+    public void setIngestIfNotExists(ArrayList<String> ingestIfNotExists) {
```
tamirkamara: missing javadoc

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -33,7 +42,54 @@ public IngestionReportMethod getReportMethod() {
...
+
+    public ArrayList<String> getAdditionalTags() {
+        return additionalTags;
+    }
+
+    public ArrayList<String> getIngestIfNotExists() {
+        return ingestIfNotExists;
+    }
+
+    public Map<String, String> getAdditionalProperties() throws IOException {
```
tamirkamara: Is it likely from a client application to ask for this?
If yes - it's missing javadoc
If no - access can be package-private

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
ohadbitt: will return a new object
Its actually very weird the way it works now - javadoc explains that all is added to additionalProperties which is private and shouldn't be known to user ...

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
ohadbitt: it is used in many other places in the library

tamirkamara: Right, but if it will be removed from there then you won't have it here and you will get the error only after "install" maven goal. It should be explicitly declared in each project it is being used at (probably we missed it before)

ohadbitt: its a

PR #83, opened 2019-05-01
reportStatusWithoutSecrets

ohadbitt:
Report table Status Without table sas


ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/BlobSourceInfo.java
```diff
@@ -29,20 +31,24 @@ public void setRawSizeInBytes(long rawSizeInBytes) {
 
     public BlobSourceInfo(String blobPath) {
         this.blobPath = blobPath;
+        this.blobPathWithoutSecrets = blobPath.split(";",2)[0].split("[?]",2)[0];
```
tamirkamara: Isn't it the same to split by the "[?]" without the first part?

```suggestion
        this.blobPathWithoutSecrets = blobPath.split("[?]",2)[0];
```

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/BlobSourceInfo.java
```diff
@@ -8,6 +8,8 @@
 public class BlobSourceInfo extends AbstractSourceInfo {
 
     private String blobPath;
+    private String blobPathWithoutSecrets;
```
tamirkamara: I suggest not "saving" this property in the object since it's just a simple calculation that doesn't cost much. So maybe just do the calculation in a method on demand...

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/BlobSourceInfo.java
ohadbitt: i guess u r right 
i took it from the .Net implementation - kusto replaces the '?' with ';' ... but its not relevant here

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
```diff
@@ -339,4 +340,16 @@ private String getSampleResultSetDump() {
                 "\"1\",\"leo\"\n\"2\",\"yui\"\n" :
                 "\"1\",\"leo\"\r\n\"2\",\"yui\"\r\n";
     }
+
+    @Test
+    void IngestFromBlob_IngestionReportMethodIsTable_RemovesSecrets() throws Exception {
```
tamirkamara: I would move this test to a new file dedicated to the new class you've created. it then can be scaled down to testing that everything is stripped like you do with the url comparison. No need to do that test here in the main class (IMO)

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
```diff
@@ -339,4 +340,16 @@ private String getSampleResultSetDump() {
...
+        BlobSourceInfo blobSourceInfo = new BlobSourceInfo("https://storage.table.core.windows.net/ingestionsstatus20190505?sv=2018-03-28&tn=ingestionsstatus20190505&sig=anAusomeSecret%2FK024xNydFzT%2B2cCE%2BA2S8Y6U%3D&st=2019-05-05T09%3A00%3A31Z&se=2019-05-09T10%3A00%3A31Z&sp=raud", 100);
```
toshetah: I hope it is not a real sas

ingest/src/main/java/com/microsoft/azure/kusto/ingest/SecretsHandler.java
```diff
@@ -0,0 +1,7 @@
+package com.microsoft.azure.kusto.ingest;
+
+class SecretsHandler {
```
toshetah: Maybe SecurityUtils?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
ohadbitt: of course not

ingest/src/main/java/com/microsoft/azure/kusto/ingest/SecurityUtils.java
```diff
@@ -0,0 +1,7 @@
+package com.microsoft.azure.kusto.ingest;
+
+class SecurityUtils {
+    static String removeSecretsFromUrl(String url) {
+        return url.split("[?]", 2)[0];
```
toshetah:
```suggestion
        return url.split("[?]", 2)[0].split("[;]", 2)[0];
```

PR #84, opened 2019-05-12
aadAuthencticatorLocksFix

ohadbitt:
aadAuthencticator Locks Fix


data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -85,10 +85,13 @@ String acquireAccessToken() throws DataServiceException {
                 acquireToken();
             } else {
                 lastAuthenticationResultLock.lock();
-                if (isTokenExpired()) {
-                    lastAuthenticationResult = acquireAccessTokenByRefreshToken();
+                try{
+                    if (isTokenExpired()) {
+                        lastAuthenticationResult = acquireAccessTokenByRefreshToken();
```
toshetah: Let's also fix the isTokenExpired method.

PR #89, opened 2019-05-22
AddExceptionsToResults

ohadbitt:
Temporary fix (until moved to v2 client) - to add exceptions got from server if status is OK, that is when both data and exception are returned

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Results.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Results.java
@@ -7,12 +7,14 @@ public class Results {
     private HashMap<String, Integer> columnNameToIndex;
     private HashMap<String, String> columnNameToType;
     private ArrayList<ArrayList<String>> values;
+    private String exceptionMessage;
 
     public Results(HashMap<String, Integer> columnNameToIndex, HashMap<String, String> columnNameToType,
-                   ArrayList<ArrayList<String>> values) {
+                   ArrayList<ArrayList<String>> values, String exceptionMessage) {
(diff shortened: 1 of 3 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/Results.java
```diff
@@ -34,4 +36,8 @@ public String getTypeByColumnName(String columnName) {
     public ArrayList<ArrayList<String>> getValues() {
         return values;
     }
+
+    public String getExceptions() {
```
danield137: why the plural? is it exceptions or exception?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/ResourceManagerTest.java
```diff
@@ -122,7 +122,7 @@ private static Results generateIngestionResourcesResult() {
         valuesList.add(new ArrayList<>((Arrays.asList("TempStorage", STORAGE_2))));
         valuesList.add(new ArrayList<>((Arrays.asList("IngestionsStatusTable", STATUS_TABLE))));
 
-        return new Results(colNameToIndexMap, colNameToTypeMap, valuesList);
+        return new Results(colNameToIndexMap, colNameToTypeMap, valuesList, "");
```
danield137: shouldn't this be an overload?

danield137: instead of passing an empty string , shouldn't we add another overload without the extra param?

PR #93, opened 2019-06-24
Streaming ingest client

ArielYehezkely:


ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -0,0 +1,181 @@
...
+        Ensure.argIsNotNull(ingestionProperties, "ingestionProperties");
+
+        fileSourceInfo.validate();
+        ingestionProperties.validate();
+
+        try{
+            String filePath = fileSourceInfo.getFilePath();
+            InputStream stream = new FileInputStream(filePath);
+            StreamSourceInfo streamSourceInfo = new StreamSourceInfo(stream,false, fileSourceInfo.getSourceId());
+            streamSourceInfo.setIsCompressed(this.azureStorageClient.isCompressed(filePath));
```
ohadbitt: is this how we check compression in .NET?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -0,0 +1,181 @@
...
+        this.azureStorageClient = new AzureStorageClient();
+    }
+
+    @Override
+    public IngestionResult ingestFromFile(FileSourceInfo fileSourceInfo, IngestionProperties ingestionProperties) throws IngestionClientException, IngestionServiceException {
+        Ensure.argIsNotNull(fileSourceInfo, "fileSourceInfo");
+        Ensure.argIsNotNull(ingestionProperties, "ingestionProperties");
+
+        fileSourceInfo.validate();
+        ingestionProperties.validate();
```
ohadbitt: it looks like validate here check if there is a report method -what is the report method for streaming ?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -0,0 +1,181 @@
...
+
+        try{
+            String filePath = fileSourceInfo.getFilePath();
+            InputStream stream = new FileInputStream(filePath);
+            StreamSourceInfo streamSourceInfo = new StreamSourceInfo(stream,false, fileSourceInfo.getSourceId());
+            streamSourceInfo.setIsCompressed(this.azureStorageClient.isCompressed(filePath));
+            return ingestFromStream(streamSourceInfo, ingestionProperties);
+        } catch (FileNotFoundException e) {
+            log.error("File not found error when ingesting a file.", e);
```
ohadbitt: drop the 'error' word from log

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    @Override
+    public IngestionResult ingestFromResultSet(ResultSetSourceInfo resultSetSourceInfo, IngestionProperties ingestionProperties) throws IngestionClientException, IngestionServiceException {
+        return ingestFromResultSet(resultSetSourceInfo, ingestionProperties, "");
+    }
+
+    @Override
+    public IngestionResult ingestFromResultSet(ResultSetSourceInfo resultSetSourceInfo, IngestionProperties ingestionProperties, String tempStoragePath) throws IngestionClientException, IngestionServiceException {
```
ohadbitt: tempStoragePath -> tempFilePath

ingest/src/main/java/com/microsoft/azure/kusto/ingest/Utils.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public static long estimateFileRawSize(AzureStorageClient azureStorageClient, String filePath) {
+        File file = new File(filePath);
+        long fileSize = file.length();
+
+        return azureStorageClient.isCompressed(filePath) ?
+                fileSize * COMPRESSED_FILE_MULTIPLIER : fileSize;
+    }
+
+    public static long resultSetToCsv(ResultSet resultSet, Writer writer, boolean includeHeaderAsFirstRow)
```
ohadbitt: name isn't informative - maybe - writeResultSetToWriterAsCsv

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -0,0 +1,181 @@
...
+
+        try {
+            File tempFile;
+            if (StringUtils.isBlank(tempStoragePath)) {
+                tempFile = File.createTempFile("kusto-resultset", ".csv.gz");
+            } else {
+                log.debug("Temp file will be created in a user specified folder: {}", tempStoragePath);
+                tempFile = File.createTempFile("kusto-resultset", ".csv.gz", new File(tempStoragePath));
+            }
+            FileOutputStream fos = new FileOutputStream(tempFile, false);
```
ohadbitt: why using a file and not streams ?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -0,0 +1,181 @@
...
+            GZIPOutputStream gzipos = new GZIPOutputStream(fos);
+            Writer writer = new OutputStreamWriter(new BufferedOutputStream(gzipos), StandardCharsets.UTF_8);
+            log.debug("Writing resultset to temp csv file: {}", tempFile.getAbsolutePath());
+            long numberOfChars = Utils.resultSetToCsv(resultSetSourceInfo.getResultSet(), writer, false);
+
+            // utf8 chars are 2 bytes each
+            FileSourceInfo fileSourceInfo = new FileSourceInfo(tempFile.getAbsolutePath(), numberOfChars * 2);
+            IngestionResult ingestionResult = ingestFromFile(fileSourceInfo, ingestionProperties);
+
+            //noinspection ResultOfMethodCallIgnored
```
ohadbitt: minor: space in comment

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -0,0 +1,181 @@
...
+        String format = ingestionProperties.getDataFormat();
+        if (format == null) {
+            return "csv";
+        }
+        return format;
+    }
+
+    private String getMappingReference(IngestionProperties ingestionProperties, String format) throws IngestionClientException {
+        String mappingReference = ingestionProperties.getIngestionMapping().IngestionMappingReference;
+        if (mappingRequiredFormats.contains(format) && StringUtils.isAnyEmpty(mappingReference)) {
```
ohadbitt: isAnyEmpty -> isBlank()

ingest/src/main/java/com/microsoft/azure/kusto/ingest/Utils.java
```diff
@@ -0,0 +1,110 @@
...
+    private final static Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+    public static long estimateBlobRawSize(AzureStorageClient azureStorageClient, String blobPath) throws StorageException, URISyntaxException {
+        long blobSize = azureStorageClient.getBlobSize(blobPath);
+
+        return azureStorageClient.isCompressed(blobPath) ?
+                blobSize * COMPRESSED_FILE_MULTIPLIER : blobSize;
+    }
+
+    public static long estimateFileRawSize(AzureStorageClient azureStorageClient, String filePath) {
```
ohadbitt: can the azure client check if local files are compressed?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/Utils.java
```diff
@@ -0,0 +1,110 @@
...
+            throws IngestionClientException {
+        final String LINE_SEPARATOR = System.getProperty("line.separator");
+
+        try {
+            String columnSeparator = "";
+
+            ResultSetMetaData metaData = resultSet.getMetaData();
+            int numberOfColumns = metaData.getColumnCount();
+
+            if (includeHeaderAsFirstRow) {
```
ohadbitt: code repetition from writeResultSetRow (not relevent if u listen to my above comment)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/Utils.java
ohadbitt: I think it would be better to change this to simply use Java CsvWriter as we do in the spark connector (com.univocity.parsers.csv.CsvWriter)

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -16,6 +20,7 @@
     private static final String DEFAULT_DATABASE_NAME = "NetDefaultDb";
     private static final Long COMMAND_TIMEOUT_IN_MILLISECS = TimeUnit.MINUTES.toMillis(10) + TimeUnit.SECONDS.toMillis(30);
     private static final Long QUERY_TIMEOUT_IN_MILLISECS = TimeUnit.MINUTES.toMillis(4) + TimeUnit.SECONDS.toMillis(30);
+    private static final Long STREAMING_INGEST_TIMEOUT_IN_MILLISECS = TimeUnit.MINUTES.toMillis(10);
```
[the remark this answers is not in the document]
ArielYehezkely: 10 minutes =)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/Utils.java
ArielYehezkely: Its just checks the path suffix:
return fileName.endsWith(".gz") || fileName.endsWith(".zip");

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
ArielYehezkely: .Net doesn't support ingestion of compressed data right now, there is no check and we always compress the data before ingestion

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -42,6 +41,22 @@ public void clearOptions() {
         options.clear();
     }
 
+    public void setParameter(String name, Object value) {
```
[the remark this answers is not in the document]
ArielYehezkely: This is to support query parameters as described in:
https://docs.microsoft.com/en-us/azure/kusto/query/queryparametersstatement#rest-api

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -6,16 +6,22 @@
...
+import java.io.*;
 import java.net.URISyntaxException;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.zip.GZIPOutputStream;
 
-public class ClientImpl implements Client {
+public class ClientImpl implements Client, StreamingIngestProvider {
```
danield137: not sure why we want to mix ingestion stuff into data package

ingest/src/main/java/com/microsoft/azure/kusto/ingest/Utils.java
ArielYehezkely: Will change in another PR

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
```diff
@@ -0,0 +1,12 @@
...
+
+public class IngestionMapping {
+    public INGESTION_MAPPING_KIND IngestionMappingKind;
+    public String IngestionMappingReference;
+
+    public IngestionMapping() {
+        IngestionMappingKind = INGESTION_MAPPING_KIND.unknown;
+    }
+
+    public enum INGESTION_MAPPING_KIND {unknown, csv, json, avro}
```
danield137: why is this caps? https://google.github.io/styleguide/javaguide.html#s4.8.1-enum-classes ?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -263,78 +261,7 @@ public IngestionResult ingestFromResultSet(
...
-
-            columnSeparator = ",";
-            numberOfChars += columnString.length();
-        }
-
-        return numberOfChars
-                + numberOfColumns * 2 * columnSeparator.length() // 2 " per column
-                + numberOfColumns - 1 // last column doesn't have a separator
-                ;
+    long writeResultSetToWriterAsCsv(ResultSet resultSet, Writer writer, boolean includeHeaderAsFirstRow) throws IngestionClientException {
```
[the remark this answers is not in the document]
ArielYehezkely: For unit testing purposes, can't mock static class Utils

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -169,6 +170,18 @@ public void setIngestIfNotExists(ArrayList<String> ingestIfNotExists) {
             fullAdditionalProperties.put("ingestIfNotExists", ingestIfNotExistsJson);
         }
         fullAdditionalProperties.putAll(additionalProperties);
+
+        switch (ingestionMapping.IngestionMappingKind) {
```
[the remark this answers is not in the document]
ArielYehezkely: Those key names are part of the ingestion properties object we pass the DM, I think it is belongs in here because its a part of the ingestion properties definition solely and not as logic in the ingestion mapping object

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    StreamingIngestClient(StreamingIngestProvider streamingIngestProvider) {
+        log.info("Creating a new StreamingIngestClient");
+        this.streamingIngestProvider = streamingIngestProvider;
+        this.azureStorageClient = new AzureStorageClient();
+    }
+
+    @Override
+    public IngestionResult ingestFromFile(FileSourceInfo fileSourceInfo, IngestionProperties ingestionProperties) throws IngestionClientException, IngestionServiceException {
```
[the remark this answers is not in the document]
ArielYehezkely: Ingest from file might be slow but it is a common scenario that we want to support. 
We agree that ingesting from blob is not recommended with streaming ingest but we want to give the same API of ingestion across clients. 
I'll comment it isn't recommended.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/StreamingIngestClientTest.java
```diff
@@ -0,0 +1,450 @@
...
+        gzipOutputStream.close();
+        inputStream = new ByteArrayInputStream(byteArrayOutputStream.toByteArray());
+        streamSourceInfo.setStream(inputStream);
+        streamSourceInfo.setIsCompressed(true);
+        status = streamingIngestClient.ingestFromStream(streamSourceInfo, ingestionProperties).getIngestionStatusCollection().get(0).status;
+        assert status == OperationStatus.Succeeded : "Ingestion failed with status: " + status.toString();
+    }
+
+    @Test
+    void IngestFromStream_StreamSourceInfo_IllegalArgumentException() {
```
tamirkamara: You should use names and/or messages that will help you understand what's wrong when the test fails

ingest/src/test/java/com/microsoft/azure/kusto/ingest/StreamingIngestClientTest.java
```diff
@@ -0,0 +1,450 @@
...
+        assertThrows(IllegalArgumentException.class,
+                () -> streamingIngestClient.ingestFromStream(null, ingestionProperties),
+                "Expected IllegalArgumentException to be thrown, but it didn't");
+    }
+
+    @Test
+    void IngestFromStream_IngestionProperties_IllegalArgumentException() {
```
tamirkamara: There're a bunch of tests here (and in methods below) with generic messages, how will you know what's wrong when something will fail?
One option would be to separate those into different tests

ingest/src/test/java/com/microsoft/azure/kusto/ingest/StreamingIngestClientTest.java
```diff
@@ -0,0 +1,450 @@
...
+        // Avro format with incompatible mapping reference
+        ingestionProperties.setIngestionMapping("CsvMapping", IngestionMapping.IngestionMappingKind.csv);
+        ingestionClientException = assertThrows(IngestionClientException.class,
+                () -> streamingIngestClient.ingestFromStream(streamSourceInfo, ingestionProperties),
+                "Expected IngestionClientException to be thrown, but it didn't");
+        assertTrue(ingestionClientException.getMessage().contains("Wrong ingestion mapping for format avro, found csv mapping kind."));
+    }
+
+    @Test
+    void IngestFromStream_EmptyStream() {
```
tamirkamara: Usually we use a convention of method-tested_what-is-tested_expected-result

ingest/src/test/java/com/microsoft/azure/kusto/ingest/StreamingIngestClientTest.java
```diff
@@ -0,0 +1,450 @@
...
+        InputStream inputStream = new ByteArrayInputStream(new byte[0]);
+        StreamSourceInfo streamSourceInfo = new StreamSourceInfo(inputStream);
+        IngestionClientException ingestionClientException = assertThrows(IngestionClientException.class,
+                () -> streamingIngestClient.ingestFromStream(streamSourceInfo, ingestionProperties),
+                "Expected IngestionClientException to be thrown, but it didn't");
+        assertTrue(ingestionClientException.getMessage().contains("Empty stream."));
+    }
+
+    @Test
+    void IngestFromFile() throws Exception {
```
tamirkamara: Usually we use a convention of method-tested_what-is-tested_expected-result

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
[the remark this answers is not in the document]
tamirkamara: Is this resolved?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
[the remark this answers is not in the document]
tamirkamara: When saying clients, are you talking about batch vs. streaming?
(a suggestions: log a warning for the non-recommended options so that the user will know)

samples/src/main/java/StreamingIngest.java
```diff
@@ -0,0 +1,153 @@
...
+    static private final String appKey = "";
+    static private final String tenant = "";
+    static private final String database = "";
+    static private final String table = "";
+    static private final String blobContainer = "https://<storage_account>.blob.core.windows.net/<container>/";
+
+    private static ConnectionStringBuilder csb;
+    private static StreamingIngestClient streamingIngestClient;
+
+    public static void main(String[] args) {
```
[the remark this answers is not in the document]
tamirkamara: @ArielYehezkely 
Is this resolved?

samples/src/main/java/StreamingIngest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    static void IngestFromStream() throws Exception {
+        IngestionProperties ingestionProperties = new IngestionProperties(database, table);
+
+        // Create Stream from string and Ingest
+        String data = "0,00000000-0000-0000-0001-020304050607,0,0,0,0,0,0,0,0,0,0,2014-01-01T01:01:01.0000000Z,Zero,\"Zero\",0,00:00:00,,null";
+        InputStream inputStream = new ByteArrayInputStream(Charset.forName("UTF-8").encode(data).array());
+        StreamSourceInfo streamSourceInfo = new StreamSourceInfo(inputStream);
+        OperationStatus status = streamingIngestClient.ingestFromStream(streamSourceInfo, ingestionProperties).getIngestionStatusCollection().get(0).status;
+        assert status == OperationStatus.Succeeded : "Ingestion failed with status: " + status.toString();
```
tamirkamara: remove the asserts, it's a sample... just print the status

samples/src/main/java/StreamingIngest.java
```diff
@@ -0,0 +1,129 @@
...
+
+            IngestFromStream();
+            IngestFromFile();
+
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
+
+    static void IngestFromStream() throws Exception {
```
tamirkamara: I think there're too much samples here which are very similar. I'd put just a couple of examples that are different with respect to Kusto, like with/without mapping, compress/uncompressed - those are the interesting things and not the csv vs. tsv.

Also, make it easy for users, and treat csv like it's not the default, meaning set a data format for it too. It makes it easier to understand...

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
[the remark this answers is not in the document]
tamirkamara: Not sure I'm following the logic here. Let's discuss...

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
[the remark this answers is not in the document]
ArielYehezkely: Yes. Maybe a bug with Github's changes review.
https://github.com/Azure/azure-kusto-java/blob/StreamingIngestClient/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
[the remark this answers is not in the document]
ArielYehezkely: We want to have both the same Interface for different client classes within each SDK (so yes in this case batched vs streaming) and the same functionality across all SDKs.

PR #94, opened 2019-07-28
Result set to csv

ArielYehezkely:
Use Univocity's CsvRoutines to convert ResultSet to CSV instead of using own implementation


ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
```diff
@@ -171,89 +181,29 @@ void IngestFromStream_NullStreamSourceInfo_IllegalArgumentException() {
...
-                IngestionClientException.class,
-                () -> ingestClientSpy.ingestFromResultSet(resultSetSourceInfo, ingestionProperties));
+                IllegalArgumentException.class,
+                () -> ingestClientImpl.ingestFromResultSet(null, ingestionProperties));
     }
 
     @Test
-    void IngestFromResultSet_FileIngest_IngestionClientException()
-            throws IngestionClientException, IngestionServiceException, SQLException {
+    void IngestFromResultSet_StreamIngest_IngestionClientException() throws Exception {
```
ohadbitt: why did you remove the old test?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -214,54 +210,26 @@ private String genBlobName(String fileName, String databaseName, String tableNam
...
-            File tempFile;
-
-            if (StringUtils.isBlank(tempStoragePath)) {
-                tempFile = File.createTempFile("kusto-resultset", ".csv.gz");
-            } else {
-                log.debug("Temp file will be created in a user specified folder: {}", tempStoragePath);
-                tempFile = File.createTempFile("kusto-resultset", ".csv.gz", new File(tempStoragePath));
-            }
+            ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+            new CsvRoutines().write(resultSetSourceInfo.getResultSet(), byteArrayOutputStream);
```
ohadbitt: nice!!

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -214,54 +210,26 @@ private String genBlobName(String fileName, String databaseName, String tableNam
         return String.format("%s__%s__%s__%s", databaseName, tableName, UUID.randomUUID().toString(), fileName);
     }
 
-    public IngestionResult ingestFromResultSet(
-            ResultSetSourceInfo resultSetSourceInfo, IngestionProperties ingestionProperties)
+    public IngestionResult ingestFromResultSet(ResultSetSourceInfo resultSetSourceInfo, IngestionProperties ingestionProperties)
```
danield137: I would cut off the line at throws, or reformat a bit
```suggestion
    public IngestionResult ingestFromResultSet(
               ResultSetSourceInfo resultSetSourceInfo, 
               IngestionProperties ingestionProperties
     ) throws ...
```

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
ArielYehezkely: It is the same test, the old implementation use temp storage to convert ResultSet to CSV and then calls IngestFromFile. In the new implementation the library reads the ResultSet in memory and writes it to stream so the method calls IngestFromStream instead.

PR #98, opened 2019-08-29
Parquet Ingestion and Mapping

ohadbitt:
Support parquet Ingestion and Mapping

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
@@ -172,9 +172,9 @@ public class IngestionProperties {
         }
         fullAdditionalProperties.putAll(additionalProperties);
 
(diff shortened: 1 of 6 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -171,17 +172,11 @@ public void setIngestIfNotExists(ArrayList<String> ingestIfNotExists) {
...
-            case csv:
-                fullAdditionalProperties.put("csvMappingReference", ingestionMapping.getIngestionMappingReference());
-                break;
-            case json:
-                fullAdditionalProperties.put("jsonMappingReference", ingestionMapping.getIngestionMappingReference());
-                break;
-            case avro:
-                fullAdditionalProperties.put("avroMappingReference", ingestionMapping.getIngestionMappingReference());
-                break;
+        String mapping = ingestionMapping.getIngestionMappingReference();
```
ArielYehezkely: String mappingReference

PR #100, opened 2019-09-19
changing Results impl using java interfaces instead than concrete imp…

jesinity:
I have slightly changed the signature of  `Results` class, exposing java interfaces (Map,List) rather than concrete impl (HashMap, ArrayList).
This breaks compatibility, so I increased version to 1.2.0-snapshot.

I also added the `headers()` method to return headers in order

what you think?


ohadbitt:
This is not relevant as we are going to change the results returned completely ..

jesinity:
better late than never!

PR #107, opened 2019-11-06
Bug fixes

ohadbitt:
Add gz to file name for ingest from stream
replaces url spaces with %20 for databases with spaces and streaming ingest

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -34,6 +34,8 @@ class Utils {
         } else {
             httpClient = HttpClients.createSystem();
         }
+
+        url = url.replace(" ", "%20");
         HttpPost httpPost = new HttpPost(url);
 
         // Request parameters and other properties.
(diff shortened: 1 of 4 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -34,6 +34,8 @@ static Results post(String url, String payload, InputStream stream, Integer time
         } else {
             httpClient = HttpClients.createSystem();
         }
+
+        url = url.replace(" ", "%20");
```
danield137: Please don't do this manually, check this out :
https://stackoverflow.com/a/10786112/929164

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -172,7 +172,9 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
                 throw new IngestionClientException("Stream");
             }
             String blobName = genBlobName(
-                    "StreamUpload", ingestionProperties.getDatabaseName(), ingestionProperties.getTableName());
+                    "StreamUpload",
+                    ingestionProperties.getDatabaseName(),
+                    ingestionProperties.getTableName()) + ".gz";
```
danield137: now that you have compression types in other PR, just use that instead of this string

PR #119, opened 2020-01-23
Better dealing with compression

ohadbitt:


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -191,7 +191,7 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
             );
```
ArielYehezkely: If compressionType != null we shouldn't compress the stream, also what about streams of parquet and orc?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientImpl.java
```diff
@@ -181,17 +181,22 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
                     ingestionProperties.getDatabaseName(),
                     ingestionProperties.getTableName(),
                     ingestionProperties.getDataFormat(),
-                    streamSourceInfo.getCompressionType());
+                    streamSourceInfo.getCompressionType() != null ? streamSourceInfo.getCompressionType() : CompressionType.gz);
```
ArielYehezkely: This condition should count shouldCompress as well (otherwise we will add redundant .gz extension when its parquet or orc streams).

ingest/src/main/java/com/microsoft/azure/kusto/ingest/AzureStorageClient.java
```diff
@@ -183,9 +183,10 @@ CompressionType getCompression(String fileName) {
     }
 
     // We don't support compression of Parquet and Orc files
-    boolean shouldCompress(CompressionType sourceCompressionType, String data_format){
+    static boolean shouldCompress(CompressionType sourceCompressionType, String data_format){
         return sourceCompressionType == null
-            && !data_format.equals(IngestionProperties.DATA_FORMAT.parquet.name())
-            && !data_format.equals(IngestionProperties.DATA_FORMAT.orc.name());
+            && (data_format == null ||
```
ArielYehezkely: return !(sourceCompressionType != null ||
data_format.equals(IngestionProperties.DATA_FORMAT.parquet.name() || 
data_format.equals(IngestionProperties.DATA_FORMAT.orc.name());

PR #122, opened 2020-03-08
Inline ingestion mapping

ohadbitt:
Support inline ingestion mapping

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ColumnMapping.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ColumnMapping.java
@@ -1,0 +1,20 @@
+package com.microsoft.azure.kusto.ingest;
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Abstract class to column mapping.
+ */
+public abstract class ColumnMapping implements Serializable {
+    private String columnName;
+    private String cslDataType;
+    private Map Properties;
+
+    public ColumnMapping(String columnName, String cslDataType, Integer ordinal) {
+        this.columnName = columnName;
+        this.cslDataType = cslDataType;
+        this.Properties = new HashMap<String, String>();
+    }
+}
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ColumnMapping.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ColumnMapping.java
@@ -1,6 +1,10 @@
 package com.microsoft.azure.kusto.ingest;
 
+import com.microsoft.azure.kusto.ingest.source.TransformationMethod;
+import org.apache.commons.lang3.StringUtils;
+
 import java.io.Serializable;
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -11,20 +15,92 @@ public class ColumnMapping implements Serializable {
     public String columnType;
     public Map<String, String> properties;
 
-    public ColumnMapping(String columnName, String cslDataType, Map properties) {
+    public ColumnMapping(String columnName, String cslDataType) {
+        this(columnName, cslDataType, new HashMap<>());
+    }
+
+    public ColumnMapping(String columnName, String cslDataType, Map<String, String> properties) {
         this.columnName = columnName;
         this.columnType = cslDataType;
         this.properties = properties;
     }
 
-    public void addProperty(MappingConst propertyName, String value){
-        if (propertyName != null){
-            properties.put(propertyName.getName(), value);
-        }
+    public void setJsonPath(String path) {
+        properties.put(MappingConst.PATH.getName(), path);
     }
 
-    public String getProperty(MappingConst propertyName){
-        String propertyValue = properties.get(propertyName.getName());
-        return propertyValue == null ? "" : propertyValue;
+    public String getJsonPath() {
+        return properties.get(MappingConst.PATH.getName());
+    }
+
+    public void setJsonTransform(TransformationMethod transform) {
+        properties.put(MappingConst.TRANSFORMATION_METHOD.getName(), transform.name());
+    }
+
+    public TransformationMethod getJsonTransform() {
+        String transform = properties.get(MappingConst.TRANSFORMATION_METHOD.getName());
+        return StringUtils.isEmpty(transform) ? null : TransformationMethod.valueOf(transform);
+    }
+
+    public void setCsvOrdinal(Integer ordinal) {
+        properties.put(MappingConst.ORDINAL.getName(), String.valueOf(ordinal));
+    }
+
+    Integer getCsvOrdinal() {
+        String ordinal = properties.get(MappingConst.ORDINAL.getName());
+        return StringUtils.isEmpty(ordinal) ? null : Integer.valueOf(ordinal);
+    }
+
+    public void setCsvConstantValue(String constValue) {
+        properties.put(MappingConst.CONST_VALUE.getName(), constValue);
+    }
+
+    public String getCsvConstantValue() {
+        return properties.get(MappingConst.CONST_VALUE.getName());
+    }
+
+    public void setAvroField(String field) {
+        properties.put(MappingConst.FIELD_NAME.getName(), field);
+    }
+
+    String setAvroField() {
+        return properties.get(MappingConst.FIELD_NAME.getName());
+    }
+
+    public void setAvroColumns(String columns) {
+        properties.put(MappingConst.COLUMNS.getName(), columns);
+    }
+
+    public String getAvroColumns() {
+        return properties.get(MappingConst.COLUMNS.getName());
+    }
+
+    public void setStorageDataType(String dataType) {
+        properties.put(MappingConst.STORAGE_DATA_TYPE.getName(), dataType);
+    }
+
+    public String getStorageDataType() {
+        return properties.get(MappingConst.STORAGE_DATA_TYPE.getName());
+    }
+
+    public boolean isValid(IngestionMapping.IngestionMappingKind mappingKind)
+    {
+        switch (mappingKind)
+        {
+            case Csv:
+                return !StringUtils.isEmpty(this.columnName);
+            case Json:
+            case Parquet:
+            case Orc:
+                TransformationMethod transformationMethod = getJsonTransform();
+                return !StringUtils.isEmpty(this.columnName) && (!StringUtils.isEmpty(getJsonPath())
+                        || transformationMethod == TransformationMethod.SourceLineNumber
+                        || transformationMethod == TransformationMethod.SourceLocation);
+            case Avro:
+                return !StringUtils.isEmpty(this.columnName) &&
+                        !StringUtils.isEmpty(getAvroColumns());
+            default:
+                return false;
+        }
     }
 }
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ColumnMapping.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ColumnMapping.java
@@ -1,20 +1,30 @@
 package com.microsoft.azure.kusto.ingest;
 
 import java.io.Serializable;
-import java.util.HashMap;
 import java.util.Map;
 
 /**
- * Abstract class to column mapping.
+ * Data class that describes the Mapping setting for .ingest command
  */
-public abstract class ColumnMapping implements Serializable {
-    private String columnName;
-    private String cslDataType;
-    private Map Properties;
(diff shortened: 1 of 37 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
```diff
@@ -10,15 +10,16 @@
 // This class describes a pre-define ingestion mapping by its name- mapping reference and its kind.
 /// </summary>
 public class IngestionMapping {
+    private ColumnMapping[] ingestionMapping;
     private IngestionMappingKind ingestionMappingKind;
```
vladikbr: Do we still need the ingestion mapping kind? Isn't it obsolete by the format?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
```diff
@@ -10,15 +10,16 @@
 // This class describes a pre-define ingestion mapping by its name- mapping reference and its kind.
 /// </summary>
 public class IngestionMapping {
+    private ColumnMapping[] ingestionMapping;
```
vladikbr: naming: columnMappings

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
```diff
@@ -10,15 +10,16 @@
...
     private IngestionMappingKind ingestionMappingKind;
     private String ingestionMappingReference;
-    public final static List<String> mappingRequiredFormats = Arrays.asList("json", "singlejson", "avro", "parquet", "orc");
+    public final static List<String> mappingRequiredFormats = Arrays.asList("Json", "singlejson", "avro", "parquet", "orc");
```
vladikbr: Why capital 'J'

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/IngestionErrorCode.java
```diff
@@ -172,7 +172,7 @@
     BadRequest_MappingReferenceWasNotFound,
```
vladikbr: Missing docstring

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -154,7 +154,7 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
     private String getFormat(IngestionProperties ingestionProperties) {
         String format = ingestionProperties.getDataFormat();
         if (format == null) {
-            return "csv";
+            return "Csv";
```
vladikbr: Why uppercase?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
vladikbr: Ah, ok, I see this is what we have in .NET as well.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/IngestionErrorCode.java
ohadbitt: whats missing docstring

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ColumnMapping.java
```diff
@@ -0,0 +1,30 @@
...
+    public String columnType;
+    public Map<String, String> properties;
+
+    public ColumnMapping(String columnName, String cslDataType, Map properties) {
+        this.columnName = columnName;
+        this.columnType = cslDataType;
+        this.properties = properties;
+    }
+
+    public void addProperty(MappingConst propertyName, String value){
```
danield137: shouldn't we have specific setters and getters for specific keys? it's not like ColumnMapping can hold any random key

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
```diff
@@ -33,16 +34,38 @@ public IngestionMapping(String ingestionMappingReference, IngestionMappingKind i
     }
 
     /**
-     * Sets the ingestion mapping parameters
+     * Please use setIngestionMappingReference for production as passing the mapping every time is wasteful
+     * @param ingestionMapping          Array of columnMappings of the same kind.
+     * @param ingestionMappingKind      IngestionMappingKind: the format of the source data to map from.
+     */
+    public IngestionMapping(ColumnMapping[] ingestionMapping, IngestionMappingKind ingestionMappingKind) {
```
danield137: naming here is strange. parameter name should probably be columnsMapping instead

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
```diff
@@ -33,16 +34,38 @@ public IngestionMapping(String ingestionMappingReference, IngestionMappingKind i
...
+    }
+
+    /**
+     * Sets the ingestion mapping reference parameters
      *
      * @param ingestionMappingReference String: the name of the pre-defined ingestion mapping.
      * @param ingestionMappingKind      IngestionMappingKind: the format of the source data to map from.
      */
-    public void setIngestionMapping(String ingestionMappingReference, IngestionMappingKind ingestionMappingKind) {
+    public void setIngestionMappingReference(String ingestionMappingReference, IngestionMappingKind ingestionMappingKind) {
```
danield137: I'm not sure I would mix Mapping reference into this class. but rather have a separate one instead as it makes it unclear what are mandatory fields and what aren't (also would not allow setters , just setters to enforce correct usage)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -173,8 +175,15 @@ public void setIngestIfNotExists(ArrayList<String> ingestIfNotExists) {
...
+        if (StringUtils.isNotBlank(mappingReference)) {
             fullAdditionalProperties.put("ingestionMappingReference", mappingReference);
+            fullAdditionalProperties.put("ingestionMappingType", ingestionMapping.getIngestionMappingKind().toString());
+        } else if (ingestionMapping.getIngestionMapping() != null) {
+            ObjectMapper objectMapper = new ObjectMapper();
+
+            String mapping = objectMapper.writeValueAsString(ingestionMapping.getIngestionMapping());
+            fullAdditionalProperties.put("ingestionMapping", mapping);
+            fullAdditionalProperties.put("ingestionMappingType", ingestionMapping.getIngestionMappingKind().toString());
         }
```
danield137: else - raise a validation error

ingest/src/main/java/com/microsoft/azure/kusto/ingest/MappingConst.java
```diff
@@ -0,0 +1,25 @@
+package com.microsoft.azure.kusto.ingest;
+
+public enum MappingConst {
+    // Json Mapping consts
+    PATH("Path"),
```
danield137: Like I said above, you could just expose these as methods on the class

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
danield137: Unfortunately these are case sensitive ATM

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientImplTest.java
```diff
@@ -274,10 +274,10 @@ void generateName() {
                     shouldCompress ? CompressionType.gz : compression);
         };
         String csvNoCompression = genName.apply(DATA_FORMAT.csv, null);
-        assert(csvNoCompression.endsWith("fileName.csv.gz"));
+        assert(csvNoCompression.endsWith("fileName.Csv.gz"));
```
danield137: this makes little sense. usually file names are lower case

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
ohadbitt: it is valid to not supply mapping

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ColumnMapping.java
ohadbitt: it's not. propertyName is of type MappingConst .. 
the .Net implementation doesnt even force the key to be of specific type ...

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
ohadbitt: decided to leave it lower case - this will work on next kusto versions

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ColumnMapping.java
danield137: I mean what you are doing is basically achieving the same result with less readability. we should improve the .Net version as well.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
ohadbitt: again just wanted to do same things as in the .Net - don't think this should be different here ..

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
ohadbitt: This here acutally is just a mistake of the intelij refactoring

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
ohadbitt: Vladik agreed that this should be consistent with the .Net

ingest/src/main/java/com/microsoft/azure/kusto/ingest/MappingConst.java
ohadbitt: please see if your comment is still relevant after last code iteration

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ColumnMapping.java
```diff
@@ -0,0 +1,106 @@
...
+        this(columnName, cslDataType, new HashMap<>());
+    }
+
+    public ColumnMapping(String columnName, String cslDataType, Map<String, String> properties) {
+        this.columnName = columnName;
+        this.columnType = cslDataType;
+        this.properties = properties;
+    }
+
+    public void setJsonPath(String path) {
```
vladikbr: Remove format name from all these methods.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/MappingConst.java
```diff
@@ -0,0 +1,25 @@
...
+    TRANSFORMATION_METHOD("Transform"),
+    // csv Mapping consts
+    ORDINAL("Ordinal"),
+    CONST_VALUE("ConstValue"),
+    // Avro Mapping consts
+    FIELD_NAME("Field"),
+    COLUMNS("Columns"),
+    STORAGE_DATA_TYPE("StorageDataType");
```
vladikbr: This is a general transform

ingest/src/main/java/com/microsoft/azure/kusto/ingest/MappingConst.java
```diff
@@ -0,0 +1,25 @@
+package com.microsoft.azure.kusto.ingest;
+
+public enum MappingConst {
```
vladikbr: class and file name: MappingConsts

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/TransformationMethod.java
```diff
@@ -0,0 +1,59 @@
+package com.microsoft.azure.kusto.ingest.source;
```
vladikbr: this file should go one level up, in ingest/src/main/java/com/microsoft/azure/kusto/ingest

PR #127, opened 2020-04-12
Fix command result parsing

ohadbitt:
Fix command result parsing - there are two options for the column type json property key

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
@@ -29,6 +29,7 @@ public class KustoResultSetTable implements ResultSet {
     private static final String columnsPropertyName = "Columns";
     private static final String columnNamePropertyName = "ColumnName";
     private static final String columnTypePropertyName = "ColumnType";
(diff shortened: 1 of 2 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -29,6 +29,7 @@
     private static final String columnsPropertyName = "Columns";
     private static final String columnNamePropertyName = "ColumnName";
     private static final String columnTypePropertyName = "ColumnType";
+    private static final String columnTypeSecondPropertyName = "ColumnType";
```
vladikbr, line 32: This is the same as columnTypePropertyName

vladikbr: Also, naming. I suggest columnTypeAlternateName

ohadbitt: fixed

amshalev: "DataType"

PR #148, opened 2020-08-27
Support authentication via callable

ohadbitt:
## Future Release Comment
New constructor for connection string builder that accepts a callable 

**Breaking Changes:**
- None

**Features:**
- Authentication via callable 

**Fixes:**
- None

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
@@ -39,13 +39,15 @@ class AadAuthenticationHelper {
     private AuthenticationResult lastAuthenticationResult;
     private Lock lastAuthenticationResultLock = new ReentrantLock();
     private String applicationClientId;
+    private Callable<String> tokenProvider;
 
     private enum AuthenticationType {
         AAD_USERNAME_PASSWORD,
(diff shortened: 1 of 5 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -39,13 +39,15 @@
...
+    private Callable<String> tokenProvider;
 
     private enum AuthenticationType {
         AAD_USERNAME_PASSWORD,
         AAD_APPLICATION_KEY,
         AAD_DEVICE_LOGIN,
         AAD_APPLICATION_CERTIFICATE,
-        AAD_ACCESS_TOKEN
+        AAD_ACCESS_TOKEN,
+        CALLABLE;
```
noam-shapira, line 50: Callable is implementation, please change to AAD_ACCESS_TOKEN_PROVIDER

noam-shapira (approved):

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
ohadbitt: was just about to do that

vladikbr (approved):

PR #150, opened 2020-09-06
Accept json response + fix maven build warnings

ohadbitt:
**Breaking Changes:**
- DataServiceException.cause will contain a json message from the response with more details then before
- Fasterxml.jackson library version is changed to [2.11.0,2.12.0)

### DIFF
--- a/data/pom.xml
+++ b/data/pom.xml
@@ -117,7 +117,12 @@
     </build>
 
     <dependencies>
-
+        <!-- Added as version 9.0.0 breaks adal - remove after moving to msal -->
+        <dependency>
+            <groupId>com.nimbusds</groupId>
+            <artifactId>nimbus-jose-jwt</artifactId>
+            <version>[6.0.1,9.0.0)</version>
+        </dependency>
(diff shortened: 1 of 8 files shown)


data/pom.xml
```diff
@@ -117,7 +122,12 @@
     </build>
 
     <dependencies>
-
+        <!-- Added as version 9.0.0 breaks adal - remove after moving to msalde -->
```
vladikbr, line 125:
```suggestion
        <!-- Added as version 9.0.0 breaks adal - remove after moving to msal -->
```

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -23,6 +23,9 @@
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
+//import com.microsoft.azure.kusto.data.ClientImpl.DEFAULT_DATABASE_NAME;
```
vladikbr, line 26: If these are not needed, why not just delete them?

vladikbr (approved):

PR #151, opened 2020-09-21
ManagedStreamingIngestClient

AsafMah:
This pull request adds ManagedStreamingIngestClient, which uses both StreamingIngestClient and QueuedIngestClient to get the advantages of both.

Right now it just tries the Streaming client 10 times, and if all of them fails it calls the Queued client.
The dotnet version also has a mechanism to concatenate streams to improve performance, I'm still working on that, but I think what I have here is still worth reviewing.

Mostly I'm concerned about the tests, there are some TODO items with open questions, and I probably didn't write enough of them, so please suggest more.

(I'm adding Vladik and Amit since this is my first PR so just in case I'm doing something wrong)


ingest/src/test/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClientTest.java
```diff
@@ -0,0 +1,344 @@
...
+        OperationStatus status = managedStreamingIngestClient.ingestFromFile(fileSourceInfo, ingestionProperties).getIngestionStatusCollection().get(0).status;
+        assertEquals(status, OperationStatus.Succeeded);
+        verify(streamingClientMock, atLeastOnce()).executeStreamingIngest(any(String.class), any(String.class), any(InputStream.class),
+                isNull(), any(String.class), any(String.class), any(boolean.class));
+    }
+
+    @Test
+    void IngestFromFile_CompressedJson() throws Exception {
+        String resourcesDirectory = System.getProperty("user.dir") + "/src/test/resources/";
+        String path = resourcesDirectory + "testdata.json";
```
amshalev, line 257: It's the same path you use for the prev test. although from the method name you should test here a compressed json

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -0,0 +1,132 @@
...
+        ingestionProperties.validate();
+
+        if (streamSourceInfo.isLeaveOpen()) {
+            throw new UnsupportedOperationException("Stream can't be Leave Open in ManagedStreamingIngestClient");
+        }
+
+        for (int i = 0; i < MAX_RETRY_CALLS; i++) {
+            try {
+                return streamingIngestClient.ingestFromStream(streamSourceInfo, ingestionProperties);
+            } catch (Exception e) {
```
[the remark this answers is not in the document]
boazsha: Are there any exceptions that will cause us to go straight to queued?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public IngestionResult ingestFromBlob(BlobSourceInfo blobSourceInfo, IngestionProperties ingestionProperties) throws IngestionClientException, IngestionServiceException {
+        Ensure.argIsNotNull(blobSourceInfo, "blobSourceInfo");
+        Ensure.argIsNotNull(ingestionProperties, "ingestionProperties");
+
+        blobSourceInfo.validate();
+        ingestionProperties.validate();
+
+        //if it's a blob we ingest using the queued client
```
boazsha, line 77: Space + "If"

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -0,0 +1,132 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.net.URISyntaxException;
+
+/**
+ * <p>ManagedStreamingIngestClient</p>
+ * <p>
+ * This class combines a managed streaming client with a queued streaming client, to create an optimized experience.
+ * Since the streaming client is communicating directly with the engine, it's more prone to failure, so this class
```
boazsha, line 25:
```suggestion
 * Since the streaming client communicates directly with the engine, it's more prone to failure, so this class
```

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -0,0 +1,132 @@
...
+ * holds both a streaming client and a queued client.
+ * At first it tries using the streaming client, and if it fails retries it {@value MAX_RETRY_CALLS} times, and if that
+ * still fails it then uses the queued streaming client.
```
boazsha, line 28:
```suggestion
 * It tries {@value MAX_RETRY_CALLS} times using the streaming client, after which it falls back to the queued streaming client in case of failure.
```

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -0,0 +1,132 @@
...
+    @Override
+    public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, IngestionProperties ingestionProperties) throws IngestionClientException, IngestionServiceException {
+        Ensure.argIsNotNull(streamSourceInfo, "streamSourceInfo");
+        Ensure.argIsNotNull(ingestionProperties, "ingestionProperties");
+
+        streamSourceInfo.validate();
+        ingestionProperties.validate();
+
+        if (streamSourceInfo.isLeaveOpen()) {
+            throw new UnsupportedOperationException("Stream can't be Leave Open in ManagedStreamingIngestClient");
```
boazsha, line 107: Instead of "Leave Open", I suggest to write the meaning (alternatively: "LeaveOpen can't be true in ManagedStreamingIngestClient").

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -0,0 +1,132 @@
...
+            return ingestFromStream(streamSourceInfo, ingestionProperties);
+        } catch (FileNotFoundException e) {
+            log.error("File not found when ingesting a file.", e);
+            throw new IngestionClientException("IO exception - check file path.", e);
+        }
+
+    }
+
+    @Override
+    public IngestionResult ingestFromBlob(BlobSourceInfo blobSourceInfo, IngestionProperties ingestionProperties) throws IngestionClientException, IngestionServiceException {
```
boazsha, line 70: I suggest documenting the fact that this method doesn't follow the class behavior (as stated in its documentation)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -0,0 +1,132 @@
...
+ * This class combines a managed streaming client with a queued streaming client, to create an optimized experience.
+ * Since the streaming client is communicating directly with the engine, it's more prone to failure, so this class
+ * holds both a streaming client and a queued client.
+ * At first it tries using the streaming client, and if it fails retries it {@value MAX_RETRY_CALLS} times, and if that
+ * still fails it then uses the queued streaming client.
+ */
+public class ManagedStreamingIngestClient implements IngestClient {
+
+    private final static Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+    public static final int MAX_RETRY_CALLS = 10;
```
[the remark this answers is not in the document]
AsafMah: You're right, I'll make it 3 like in dotnet

ingest/src/test/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClientTest.java
AsafMah: You're right, since that's also a problem in StreamingIngestClientTest.java I fixed it in both now

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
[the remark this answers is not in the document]
AsafMah: In C# we have the method IsPermanent (which just calls not on IsTransient) and seems to have some heuristics about what exceptions are permanent or not.
In case they are permanent it doesn't go to the queue, it just rethrows it.
None of the heuristics in the dotnet file seemed relevant here so I decided to skip it, but if you have some idea when we might want to go to the queue instantly or bail entirely I'd love to hear.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: I'm going with the second suggestion since it indicates the field name directly, which is clearer to me.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: Do you think I also should document it on the method, or just the class?

ingest/pom.xml
```diff
@@ -191,6 +191,12 @@
             <artifactId>json</artifactId>
             <version>${json.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>${annotations.version}</version>
+            <scope>compile</scope>
```
ohadbitt, line 198: compile is default - so no need

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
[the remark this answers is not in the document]
amshalev: I don't familiar with all the streaming ingestion errors but I think there are some errors that worth to check... you should consult about that with @ArielYehezkely  and\or Valentina from the Engine team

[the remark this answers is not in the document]
ohadbitt: I'm not sure if thats the case in streamingIngestion but for commands and querying - since version 2.3.0 of the SDK the server response returns errors as JSON with isPermanent field in them, pull latest and check if it has it here too or ask Valentina for that as well

ingest/src/test/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClientTest.java
```diff
@@ -0,0 +1,352 @@
...
+        ingestionProperties.setReportMethod(IngestionProperties.IngestionReportMethod.Table);
+        ArgumentCaptor<TableServiceEntity> captur = ArgumentCaptor.forClass(TableServiceEntity.class);
+
+        managedStreamingIngestClient.ingestFromBlob(blobSourceInfo, ingestionProperties);
+
+        verify(azureStorageClientMock, atLeast(1)).azureTableInsertEntity(anyString(), captur.capture());
+        assert (((IngestionStatus) captur.getValue()).getIngestionSourcePath()).equals("https://storage.table.core.windows.net/ingestionsstatus20190505");
+    }
+
+    //todo: all of these null/empty argument checks are also repetitive, but maybe that's just java
```
ohadbitt, line 151: Please change to  -> // TODO

Its not really repetitive - its just easier for debugging as separate methods but it doesn't have to be - the asserts are one liners ..

ingest/src/test/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClientTest.java
```diff
@@ -0,0 +1,352 @@
...
+            when(streamingClientMock.executeStreamingIngest(any(String.class), any(String.class), argumentCaptor.capture(),
+                    isNull(), any(String.class), eq("mappingName"), any(boolean.class)))
+                    .thenReturn(null);
+        }
+    }
+
+    @Test
+    void IngestFromStream_FailFewTimes() throws Exception {
+        int failCount = 2;
+        final int[] times = {0}; //it's an array so we can safely modify it in the lambda
```
ohadbitt, line 327: Here and all comments: we actually comment here like in the .Net code- move comments above the line, space after the `//` and first letter is capital letter . Actually I don't think you need to explain these stuff on comments ..

ingest/src/test/java/com/microsoft/azure/kusto/ingest/StreamingIngestClientTest.java
```diff
@@ -301,25 +305,34 @@ void IngestFromFile_Json() throws Exception {
...
         String path = resourcesDirectory + "testdata.json";
         FileSourceInfo fileSourceInfo = new FileSourceInfo(path, new File(path).length());
+        String contents = new String(Files.readAllBytes(Paths.get(path)), StandardCharsets.UTF_8).trim();
+
         ingestionProperties.setDataFormat(IngestionProperties.DATA_FORMAT.json);
         ingestionProperties.setIngestionMapping("JsonMapping", IngestionMapping.IngestionMappingKind.Json);
         OperationStatus status = streamingIngestClient.ingestFromFile(fileSourceInfo, ingestionProperties).getIngestionStatusCollection().get(0).status;
         assertEquals(status, OperationStatus.Succeeded);
-        verify(streamingClientMock, atLeastOnce()).executeStreamingIngest(any(String.class), any(String.class), any(InputStream.class),
+        verify(streamingClientMock, atLeastOnce()).executeStreamingIngest(any(String.class), any(String.class), argumentCaptor.capture(),
```
ohadbitt, line 314: Just wondering - why did you change to argementCaptor?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * <p>ManagedStreamingIngestClient</p>
+ * <p>
+ * This class combines a managed streaming client with a queued streaming client, to create an optimized experience.
+ * Since the streaming client communicates directly with the engine, it's more prone to failure, so this class
+ * holds both a streaming client and a queued client.
+ * It tries {@value MAX_RETRY_CALLS} times using the streaming client, after which it falls back to the queued streaming client in case of failure.
+ * <p>
+ * Note that {@code ingestFromBlob} acts differently from the other functions - since if you already got a blob it's faster to
```
ohadbitt, line 29: This is not faster - it just makes more sense .. we might want to suggest the user to set the flushImmediately flat to true in the ingestionProperties to get latency similar to streamingIngest

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionUtils.java
```diff
@@ -0,0 +1,53 @@
...
+        ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+        new CsvRoutines().write(resultSetSourceInfo.getResultSet(), byteArrayOutputStream);
+        byteArrayOutputStream.flush();
+        if (byteArrayOutputStream.size() <= 0) {
+            String message = "Empty ResultSet.";
+            log.error(message);
+            throw new IngestionClientException(message);
+        }
+        ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(byteArrayOutputStream.toByteArray());
+        return new StreamSourceInfo(byteArrayInputStream);
```
ohadbitt, line 51: you can pass to the id of the resultSet to the `StreamSourceInfo` ctor,
also for the regular streamingIngestClient

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
boazsha: FlushImmediately is not recommended for customers (@vladikbr to keep me honest).

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -0,0 +1,134 @@
...
+ * queue it rather to download it and stream it, ManagedStreamingIngestClient sends it directly to the queued client.
```
boazsha, line 30:
```suggestion
 * Note that {@code ingestFromBlob} behaves differently from the other methods - since a blob already exists it makes more sense to enqueue it rather than downloading and streaming it, thus ManagedStreamingIngestClient skips the streaming retries and sends it directly to the queued client.
```

boazsha: Btw, I'm not entirely sure that it "makes more sense", given that the user decided to use managed *streaming* client.
I understand we need to override it, but in another dimension I would drop this method altogether in this managed client.
Either way, we need to make it perfectly clear that this method doesn't utilize streaming ingest.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClientTest.java
AsafMah: Do we have a style guide somewhere?

And as for the explanation, IDK if I refactored this code someday I might think "why isn't this an int" and forget about this 'gotcha', so I think this comment is nice.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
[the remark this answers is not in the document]
AsafMah: Ok I added support for the JSON permanent field, anything else?

data/src/main/java/com/microsoft/azure/kusto/data/UncloseableStream.java
```diff
@@ -0,0 +1,70 @@
+package com.microsoft.azure.kusto.data;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * This class exists to handle outside dependencies which close a stream when we don't want to.
+ * It takes a stream, and simply forwards to all of its methods, except for the {@code close()} method - which does nothing.
```
ohadbitt, line 10:
```suggestion
 * It takes a stream, and simply forwards the call to all of its methods, except for the {@code close()} method - which does nothing.
```

ohadbitt (approved):

PR #152, opened 2020-09-22
Remove federated url path if set

ohadbitt:
#### Pull Request Description

**Features:**
- Strip Fed=true from URL


data/src/main/java/com/microsoft/azure/kusto/data/ConnectionStringBuilder.java
```diff
@@ -42,6 +42,10 @@ public String getClusterUrl() {
         return clusterUri;
     }
 
+    void setClusterUrl(String clusterUri) {
```
yihezkel, line 45: This isn't relevant to your change, but I just wanted to say that I suggest we make public all methods here that are part of its API (which doesn't include setClusterUrl). I think we want the other methods to be accessible even to other packages (even though Java SDK isn't currently organized).

data/src/test/java/com/microsoft/azure/kusto/data/ConnectionStringBuilderTest.java
```diff
@@ -158,4 +159,13 @@ void createWithAadTokenProviderAuthentication(){
         Assertions.assertDoesNotThrow( () -> ConnectionStringBuilder
                 .createWithAadTokenProviderAuthentication("resource.uri", () -> "token"));
     }
+
+    @Test
+    @DisplayName("validate ClientImpl strips fed=true")
+    void stripFedereatedAuthFromCSB() throws URISyntaxException {
```
yihezkel, line 165: Since we're trying to get all our code more into testing, I suggest also adding a negative test for this field.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -33,7 +35,16 @@
     private final String applicationNameForTracing;
 
     public ClientImpl(ConnectionStringBuilder csb) throws URISyntaxException {
-        clusterUrl = csb.getClusterUrl();
+        String url = csb.getClusterUrl();
+        URI clusterUri = new URI(url);
+        String host = clusterUri.getHost();
+        String auth = clusterUri.getAuthority().toLowerCase();
+        if (host == null && auth.endsWith(";fed=true")) {
+            url = new URIBuilder().setScheme(clusterUri.getScheme()).setHost(auth.substring(0, clusterUri.getAuthority().indexOf(";fed=true"))).toString();
```
yihezkel, line 43: Need null checks, though I'd personally instead add validation to the CSB constructor for that which is required.

Also, I like how you did this but just to point out that there's a method URIBuilder.fromUri() and then you could just set the truncated host.

yihezkel (changes requested):

data/src/test/java/com/microsoft/azure/kusto/data/ConnectionStringBuilderTest.java
ohadbitt: what could be a negative test here?

data/src/main/java/com/microsoft/azure/kusto/data/ConnectionStringBuilder.java
ohadbitt: we should consider this but yea not for this PR

data/src/test/java/com/microsoft/azure/kusto/data/ConnectionStringBuilderTest.java
yihezkel: void dontChangeUrlIfNotFederated() throws URISyntaxException {
        String clusterUrl = "https://service.uri";
        ConnectionStringBuilder csb = ConnectionStringBuilder.createWithAadApplicationCredentials(clusterUrl, "id", "appKey");
        ClientImpl client = new ClientImpl(csb);
        Assertions.assertEquals(clusterUrl, csb.getClusterUrl());
    }

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -33,7 +35,16 @@
     private final String applicationNameForTracing;
 
     public ClientImpl(ConnectionStringBuilder csb) throws URISyntaxException {
-        clusterUrl = csb.getClusterUrl();
+        String url = csb.getClusterUrl();
+        URI clusterUri = new URI(url);
+        String host = clusterUri.getHost();
+        String auth = clusterUri.getAuthority().toLowerCase();
```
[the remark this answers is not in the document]
yihezkel: My mistake, I thought we could create a CSB using a constructor and bypass the helper method (createWithAadApplicationCredentials()). IIRC that's valid in the .NET SDK.
I see here that the constructor is private.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
yihezkel: You're right, CSB can only be created via helper method, as above.

yihezkel (approved):

PR #153, opened 2020-09-30
Expose authority id for all authenticaiton methods

ohadbitt:
**Fixes:**
- Device authentication and certificate authentication did not expose a way to set authority id

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
@@ -197,7 +197,7 @@ class AadAuthenticationHelper {
 
     private AuthenticationResult waitAndAcquireTokenByDeviceCode(DeviceCode deviceCode, AuthenticationContext context)
             throws InterruptedException {
-        int timeout = 20 * 1000;
+        int timeout = 60 * 1000;
         AuthenticationResult result = null;
         while (timeout > 0) {
             try {
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ConnectionStringBuilder.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ConnectionStringBuilder.java
@@ -147,17 +147,31 @@ public class ConnectionStringBuilder {
         return createWithAadApplicationCredentials(resourceUri, applicationClientId, applicationKey, null);
     }
 
-    public static ConnectionStringBuilder createWithDeviceCodeCredentials(String resourceUri) {
+    public static ConnectionStringBuilder createWithDeviceCodeCredentials(String resourceUri, String authorityId) {
         if (StringUtils.isEmpty(resourceUri)) {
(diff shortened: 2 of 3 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/ConnectionStringBuilder.java
```diff
@@ -147,17 +147,31 @@ public static ConnectionStringBuilder createWithAadApplicationCredentials(String
...
 
-    public static ConnectionStringBuilder createWithDeviceCodeCredentials(String resourceUri) {
+    public static ConnectionStringBuilder createWithDeviceCodeCredentials(String resourceUri, String authorityId) {
         if (StringUtils.isEmpty(resourceUri)) {
             throw new IllegalArgumentException("resourceUri cannot be null or empty");
         }
-        return new ConnectionStringBuilder(resourceUri);
+
+        ConnectionStringBuilder csb = new ConnectionStringBuilder(resourceUri);
+        csb.aadAuthorityId = authorityId;
```
vladikbr, line 156: What happens if authorityId is empty?
To what will it default?

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
```diff
@@ -197,7 +197,7 @@ private AuthenticationResult acquireAccessTokenUsingDeviceCodeFlow() throws Exce
 
     private AuthenticationResult waitAndAcquireTokenByDeviceCode(DeviceCode deviceCode, AuthenticationContext context)
             throws InterruptedException {
-        int timeout = 20 * 1000;
+        int timeout = 60 * 1000;
```
vladikbr, line 200: I think it's a bit too much

vladikbr (changes requested):
Pls see comments

data/src/main/java/com/microsoft/azure/kusto/data/ConnectionStringBuilder.java
ohadbitt: default is microsoft.com

data/src/main/java/com/microsoft/azure/kusto/data/AadAuthenticationHelper.java
ohadbitt: it is a race to make it work - its not the first time i see people struggle to make it in 20 seconds ..  
how much do you think

vladikbr: Fair enough

data/src/main/java/com/microsoft/azure/kusto/data/ConnectionStringBuilder.java
vladikbr: "common" is better, but check with Ran that it is correct and check that it works for you correctly

PR #159, opened 2020-12-07
added classes for generating pojos from KustoResultSetTable instances

maltalex:
This commit includes three new classes:
1. KustoType which is a generic class for representing different Kusto
column types and bind them to different Java types

2. KustoColumn which represents a Kusto column - it has a name,
nullability information, and a type (KustoType)

3. KustoResultMapper which can be set up to parse KustoResultSetTable
instances and return a list of pojos. The setup phase is done using a
builder that takes the pojo constructor, as well as zero or more
instances of KustoColumn along with setter functions to set pojo's
fields based on the column values.

This mechanism can potentially be extended to use reflection based on
the target pojo class instead of the manual KustoResultMapper setup

Example usage - methods that implement a couple of simple queries (`take` and `sample`) on `MyTable` with various column types, returning a list of POJOs:

```
public class Example {

	public static enum SomeEnum {
		A, B, C;
	}
	
	public static class MyPojo {
		int id;
		String name;
		ZonedDateTime time;
		SomeEnum someEnum;
		UUID uuid;
		
		public void setId(int id) {
			this.id = id;
		}
		
		public void setName(String name) {
			this.name = name;
		}

		public void setTime(ZonedDateTime time) {
			this.time = time;
		}
		
		public void setUuid(UUID uuid) {
			this.uuid = uuid;
		}

		public void setSomeEnum(SomeEnum someEnum) {
			this.someEnum = someEnum;
		}
	}
	
	static final KustoResultMapper<MyPojo> mapper = KustoResultMapper.newBuilder(MyPojo::new)
			.addNonNullableColumn(KustoType.INTEGER, "Id", MyPojo::setId)
			.addNullableColumn(KustoType.STRING, "Name", MyPojo::setName)
			.addNullableColumn(KustoType.DATETIME_ZONED_DATE_TIME, "TimeColumnName", MyPojo::setTime)
			.addNullableColumn(KustoType.GUID_UUID, "Guid", MyPojo::setUuid)
			.addNullableColumn(KustoType.STRING, "Enum", (r,t) -> r.setSomeEnum(SomeEnum.valueOf(t))) //set enum from kusto string
			.build();

	public static List<MyPojo> take(ClientImpl client, int count) throws DataServiceException, DataClientException {
		return mapper.extractList(client.execute("MyTable | take " + count).getPrimaryResults());
	}

	public static List<MyPojo> sample(ClientImpl client, int count) throws DataServiceException, DataClientException {
		return mapper.extractList(client.execute("MyTable | sample " + count).getPrimaryResults());
	}
}
```


data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Represents a single Kusto result column with a type, name and nullability information
+ * 
+ * @param <C>
+ *            the Java class returned by the ADX column type
+ * @param <T>
+ *            type of ADX column
+ */
+public class KustoColumn<C, T extends KustoType<C>> {
```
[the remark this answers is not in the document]
ohadbitt: oalso if you would have had the 'ordinal' and could invoke getObject(ordinal) instead and save all the map lookups

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
```diff
@@ -0,0 +1,55 @@
...
+	public static final KustoType<UUID>				GUID_UUID					= new KustoType<>(UUID.class, o -> UUID.fromString(o.toString()));
+	public static final KustoType<String>			GUID_STRING					= new KustoType<>(String.class);
+	public static final KustoType<String>			STRING						= new KustoType<>(String.class);
+	public static final KustoType<Double>			REAL_DOUBLE					= new KustoType<>(Double.class);
+	public static final KustoType<Float>			REAL_FLOAT					= new KustoType<>(Float.class,
+			o -> Float.valueOf(Double.class.cast(o).floatValue()));
+	
+	public static final KustoType<ZonedDateTime>	DATETIME_ZONED_DATE_TIME	= new KustoType<>(ZonedDateTime.class, o -> ZonedDateTime.parse(o.toString()));
+	public static final KustoType<Instant>			DATETIME_INSTANT			= new KustoType<>(Instant.class, o -> Instant.parse(o.toString()));
+	public static final KustoType<Long>				DATETIME_LONG				= new KustoType<>(Long.class, o -> Instant.parse(o.toString()).toEpochMilli());
```
ohadbitt, line 32: why do we need this as part of the kusto types?
I think we should keep the list here short so that users can find easily the respective Converter for their KQL data type

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
```diff
@@ -0,0 +1,55 @@
...
+			o -> Long.valueOf(Number.class.cast(o).longValue()));
+	public static final KustoType<UUID>				GUID_UUID					= new KustoType<>(UUID.class, o -> UUID.fromString(o.toString()));
+	public static final KustoType<String>			GUID_STRING					= new KustoType<>(String.class);
+	public static final KustoType<String>			STRING						= new KustoType<>(String.class);
+	public static final KustoType<Double>			REAL_DOUBLE					= new KustoType<>(Double.class);
+	public static final KustoType<Float>			REAL_FLOAT					= new KustoType<>(Float.class,
+			o -> Float.valueOf(Double.class.cast(o).floatValue()));
+	
+	public static final KustoType<ZonedDateTime>	DATETIME_ZONED_DATE_TIME	= new KustoType<>(ZonedDateTime.class, o -> ZonedDateTime.parse(o.toString()));
+	public static final KustoType<Instant>			DATETIME_INSTANT			= new KustoType<>(Instant.class, o -> Instant.parse(o.toString()));
```
ohadbitt, line 31: Please add the following:
Duration and call it Timespan
Decimal 
Dynamic 
Boolean

data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
```diff
@@ -0,0 +1,48 @@
+package com.microsoft.azure.kusto.data;
+
+/**
+ * Represents a single Kusto result column with a type, name and nullability information
+ * 
+ * @param <C>
```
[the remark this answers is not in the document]
maltalex: Is that the only style issue?  It would probably be simpler to just run the code through your formatter before merging. Do you have a formatter you can share?

data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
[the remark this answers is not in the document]
maltalex: 1. Where's the "KustoColumn" type? I didn't see anything on a GitHub search except this PR (https://github.com/Azure/azure-kusto-java/search?q=KustoColumn)

2. How do you propose we get the ordinal information? If we detect it automatically and cache it, the mapper instance will only work correctly with queries that return the columns in the same exact order. That's a foot gun. Or do you mean we add a constructor that takes an ordinal instead of a column name so it's provided by the user?

data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
```diff
@@ -0,0 +1,48 @@
...
+		if (o == null) {
+			if (!this.isNullable) {
+				throw new NullPointerException("Column " + this.name + " is not nullable");
+			}
+			return null;
+		}
+		try {
+			return this.type.type(o);
+		} catch (Exception e) {
+			throw new IllegalArgumentException("Column " + this.name + " is of type " + o.getClass() + " but I expected type " + this.type.clazz, e);
```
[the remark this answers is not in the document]
maltalex: Okay, but any specific reason for preferring `String.format` ? It's a lot slower than string concatenation.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * A class for mapping Kusto results to a list of pojos
+ * 
+ * @param <R>
+ *            pojo type returned by the mapping
+ */
+public class KustoResultMapper<R> {
+	
+	private static class QueryResultColumnInfo<R, C> {
```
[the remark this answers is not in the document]
maltalex: It maps `ResultSet` instances to pojos, that's why it makes sense to me.

I'm definitely not married to the current name, but I fee like `ColumnPopulator` is worse. What's being populated are the different fields in the pojos. Not columns.

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
maltalex: We don't _need_ this. In fact, we don't _need_ any of these classes. They're a convenience we've written internally to make parsing Kusto results more straightforward and type safe. Since our mappers needed to map datetime to several java types, we placed them all here to avoid writing the conversion code separately for every mapper. I can never remember the right way to transform `Strings` into the various data types. Also, the parsing code is conceptually part of Kusto since it depends on the exact representation of `datetime` values returned by the Json library.

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
maltalex: Sure, but since I don't have tables with these column types, could you tell me what Java types they currently return so I can write the conversion code?

data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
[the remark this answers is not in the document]
ohadbitt: no

data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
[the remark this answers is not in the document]
ohadbitt: 1. https://github.com/Azure/azure-kusto-java/blob/a1417109a264ca56ac7d56adaae0095ea460d7b0/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultColumn.java
2. You cache it per KustoResultSetTable which is anyway different for each query you run ...

data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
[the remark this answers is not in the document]
ohadbitt: if its too hard to follow you can leave it and we fix it later

data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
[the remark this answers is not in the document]
ohadbitt: its just more readable - and this is not a code that should run a lot

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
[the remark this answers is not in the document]
ohadbitt: its always easier for the writer of the code to say that the variables names make sense to him but when i first read it it was hard for me to understand what is the purpose of the class.
You could call it DataPopulator or something like that - when something is called info it usually means it is some readonly fields and that the class doesnt do the work itself

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
ohadbitt: You should check it anyway but 
Timespan: String
decimal: String
dynamic: string or JSONObject 
boolean: Boolean

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
ohadbitt: this is public not internal i think the better way to go would be to leave here just the 10 kusto types and add another class called KustoExtendedTypes with all the others 

bool 
datetime 
dynamic 
guid (and uuid)
int 
long
real (which is .Net double - dont think we need float too)  
string 
timespan
decimal

data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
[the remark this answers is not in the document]
maltalex: 1. Done, I merged KustoColumn and QueryResultColumnInfo into a new "ObjectPopulator" class
2. Done. `extractList` now caches the column's ordinals per invocation. And users can manually specify ordinals in the mappers instead (or on top of) column names.

data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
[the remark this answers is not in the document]
maltalex: done.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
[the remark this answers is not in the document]
maltalex: Done, I merged KustoColumn and QueryResultColumnInfo into a new "ObjectPopulator" class

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
maltalex: What would be the benefit of that? I think it'll just make it less clear for users - they'd have to search in two different classes for their types instead of just one.

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
maltalex: These types make no sense to me. A timespan as a `String`? A decimal as a `String`?

Also, returning JSONObject is a terrible idea in my opinion since that's the type of the underlying JSON library. Why would you want to expose it? That would make user code depend on an internal dependency of yours, so you'll be stuck with it forever.

data/src/main/java/com/microsoft/azure/kusto/data/KustoColumn.java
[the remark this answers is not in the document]
maltalex: Note about point 2 - currently `KustoResultSetTable` doesn't implement the `ResultSet` correctly in ways that affect this change (see issue #161). For example, `ResultSet` ordinals are 1-based, while `KustoResultSetTable` uses 0-based numbering.

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
ohadbitt: I think you missunderstood me - i answered as to which types are iferered by the JSON library from these kusto types ... 
Kusto returns Timespan decimals and dynamic as strings because it is the most convenient to consume it by anyone calling the http query.
We should convert them like that:
Timespan: String -> Duration
decimal: String -> BigDecimal
dynamic : STring /JSONObject -> JSONObject / JSONArray (dynamic type is representing a .NET json object)
boolean : boolean -> boolean

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
ohadbitt: i think its better then looking for them and having so many options - 
If we have a main class showing the 10 kusto types it will be easy for a new user to find the type he needs as it will be the type of the column 
other - more complex casting could be mad using the Extended class 
This is a suggestion, please say if you think its still wrong

ohadbitt (changes requested):

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
maltalex: In that case, the timestamp, decimal and boolean make sense.

However, why would you like to expose `org.json.JSONObject ` directly to users? From my point of view, it's an implementation detail. If you ever decide to switch to some other JSON parsing library you'd be breaking user's code.

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
maltalex: Everything we're talking about is a suggestion :)
Anyway, I don't see the benefit of dividing the types into two classes. It would make discovering the API using an IDE's autocomplete functionality harder. How about using a double underscore or some other "delimiter" to divide the Kusto type and the Java types?

`DATETIME__INSTANT`
`DATETIME__ZONED_DATE_TIME`

Another idea I have is to use static inner classes for the various kusto types. One inner class for `datetime`, one for `string` etc. In each class, we can have the various java types the Kusto type can be converted into. Pseudocode:

```
class KustoType<..>{
    static class DateTime<...>{
        STRING = ...
        INSTANT = ...
        ZONED_DATE_TIME = ...
    }
    static class Guid<...>{
        STRING = ...
        UUID = ....
    }
}
```

Users will then specify the type using something like `KustoType.DateTime.INSTANT`. Maybe this would make the distinction between the Kusto type and the Java type clearer, and the amount of types less overwhelming.

data/src/main/java/com/microsoft/azure/kusto/data/ObjectPopulator.java
```diff
@@ -0,0 +1,74 @@
...
+		this.type = type;
+		this.isNullable = isNullable;
+		this.valueSetter = valueSetter;
+	}
+	
+	void populateFrom(R r, KustoResultSetTable resultSet) {
+		populateFrom(r, resultSet, this.presetOrdinal);
+	}
+	
+	void populateFrom(R r, KustoResultSetTable resultSet, int ordinal) {
```
yihezkel, line 52: Better to use a descriptive variable name

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
```diff
@@ -0,0 +1,55 @@
...
+ * 
+ * @author alex
+ *
+ * @see https://docs.microsoft.com/azure/data-explorer/kusto/query/scalar-data-types/
+ *
+ * @param <C>
+ */
+public class KustoType<C> {
+	
+	public static final KustoType<Integer>			INTEGER						= new KustoType<>(Integer.class);
```
yihezkel, line 19: We use spaces instead of tabs (tell me about it :-) )
So besides the left indentation, we suggest using a single space between variables' signature, name and what the variable is set to.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
```diff
@@ -0,0 +1,106 @@
...
+		
+		public <C> Builder<R> addNullableColumn(KustoType<C> type, String name, BiConsumer<R, C> setter) {
+			return addColumn(type, name, true, setter);
+		}
+		
+		public <C> Builder<R> addNullableColumn(KustoType<C> type, String name, int ordinal, BiConsumer<R, C> setter) {
+			return addColumn(type, name, ordinal, true, setter);
+		}
+		
+		public <C> Builder<R> addNullableColumn(KustoType<C> type, int ordinal, BiConsumer<R, C> setter) {
```
yihezkel, line 59: These 6 helper methods seem unnecessarily divided by whether the column is nullable. Why not accept "isNullable" as a parameter?

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+
+/**
+ * A class for mapping Kusto results to a list of pojos
+ * 
+ * @param <R>
+ *            pojo type returned by the mapping
+ */
+public class KustoResultMapper<R> {
```
yihezkel, line 14: I like the example code you included in the PR's description. Would you please include that as a few unit tests?

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
maltalex: Hi @yihezkel 

Like I've already mentioned to to @ohadbitt in this PR, I can fix some of the styling issues here and there, but something tells me that you guys have your own formatter and you'd end up using it on the PR anyway, so I don't see the point in fixing individual formatting issues.
I can change the tabs to spaces, no problem, but then we'd just run into some other formatting issue that goes against your code conventions. Therefore, I think that the better approach would be to format the code using your formatter (either before or after merging the PR), or somehow give me access to your formatter so I can do it myself.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
maltalex: > Why not accept "isNullable" as a parameter?

Just a personal preference. It makes the API more readable compared to a boolean flag that's easier to miss or just forget what it means. Consider the following two snippets:

1.

	static final KustoResultMapper<MyPojo> mapper = KustoResultMapper.newBuilder(MyPojo::new)
			.addNonNullableColumn(KustoType.INTEGER, "Id", MyPojo::setId)
			.addNullableColumn(KustoType.STRING, "Name", MyPojo::setName)
			.addNullableColumn(KustoType.DATETIME_ZONED_DATE_TIME, "TimeColumnName", MyPojo::setTime)
			.addNullableColumn(KustoType.GUID_UUID, "Guid", MyPojo::setUuid)
			.addNullableColumn(KustoType.STRING, "Enum", (r,t) -> r.setSomeEnum(SomeEnum.valueOf(t)))
			.build();

2.

	static final KustoResultMapper<MyPojo> mapper = KustoResultMapper.newBuilder(MyPojo::new)
			.addColumn(KustoType.INTEGER, "Id", MyPojo::setId, false)
			.addColumn(KustoType.STRING, "Name", MyPojo::setName, true)
			.addColumn(KustoType.DATETIME_ZONED_DATE_TIME, "TimeColumnName", MyPojo::setTime, true)
			.addColumn(KustoType.GUID_UUID, "Guid", MyPojo::setUuid, true)
			.addColumn(KustoType.STRING, "Enum", (r,t) -> r.setSomeEnum(SomeEnum.valueOf(t)), true)
			.build();

I tend to prefer (1), but have no strong feelings about it. I can change it to (2) if you prefer.

data/src/main/java/com/microsoft/azure/kusto/data/ObjectPopulator.java
maltalex: done.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
maltalex: done. See `KustoResultMapperTest`.

ohadbitt (commented):
Hey 
sorry we couldnt get to this 
anyway please correct the tests then we can finish with the PR

maltalex:
@ohadbitt 

> 
> Hey
> sorry we couldnt get to this
> anyway please correct the tests then we can finish with the PR

I'm sorry, I don't follow.
What tests do you mean? I added the `KustoResultMapperTest` and it passes on my machine. Other, per-existing tests also pass. I don't have access to your CI environment to see what's wrong there.

ohadbitt (approved):

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
yihezkel: You have a fair point, but I definitely prefer passing it as a mandatory (so it won't be missed) parameter - modern IDEs will show the fieldname. I think adding so many extra methods makes it more complicated.

data/src/main/java/com/microsoft/azure/kusto/data/ObjectPopulator.java
```diff
@@ -0,0 +1,73 @@
...
+ * 
+ * @param <R>
+ *            the pojo this column needs to set its value to
+ * 			
+ * @param <C>
+ *            the Java class returned by the ADX column type
+ * @param <T>
+ *            type of ADX column
+ */
+class ObjectPopulator<R, C, T extends KustoType<C>> {
```
yihezkel, line 17: I believe you already changed the name of this class, but doesn't it represent a KustoResultColumnPopulator? If so, I'd think that name makes most sense.

data/src/main/java/com/microsoft/azure/kusto/data/ObjectPopulator.java
```diff
@@ -0,0 +1,73 @@
...
+		this.isNullable = isNullable;
+		this.valueSetter = valueSetter;
+	}
+	
+	void populateFrom(R objToPopulate, KustoResultSetTable resultSet) {
+		populateFrom(objToPopulate, resultSet, this.presetOrdinal);
+	}
+	
+	void populateFrom(R objToPopulate, KustoResultSetTable resultSet, int ordinal) {
+		Object o = resultSet.getObject(columnIndexInResultSet(resultSet));
```
yihezkel, line 52: It seems calling columnIndexInResultSet() here is an unnecessary duplicate call. In the flow coming from KustoResultMapper.extractList(), we already executed this method and passed it in as the "ordinal" parameter.

yihezkel, line 52: I'd rename "o" to "resultValue"

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
```diff
@@ -0,0 +1,107 @@
...
+	}
+	
+	public List<R> extractList(KustoResultSetTable resultSet) {
+		List<R> ret = new ArrayList<>(resultSet.count());
+		int[] columnOrdinals = new int[this.columns.size()];
+		for (int i = 0; i < this.columns.size(); i++) {
+			columnOrdinals[i] = this.columns.get(i).columnIndexInResultSet(resultSet);
+		}
+		while (resultSet.next()) {
+			R r = this.objConstructor.get();
```
yihezkel, line 98: I'd rename "r" to KustoResultValue

data/src/test/java/com/microsoft/azure/kusto/data/KustoResultMapperTest.java
```diff
@@ -0,0 +1,246 @@
...
+	public void prepareResultSet() {
+		when(this.resultSet.findColumn("Id")).thenReturn(1);
+		when(this.resultSet.findColumn("Name")).thenReturn(2);
+		when(this.resultSet.findColumn("TimeColumn")).thenReturn(3);
+		when(this.resultSet.findColumn("Guid")).thenReturn(4);
+		when(this.resultSet.findColumn("Enum")).thenReturn(5);
+		when(this.resultSet.findColumn("Instant")).thenReturn(6);
+	}
+	
+	void testNonNull(KustoResultMapper<TestPojo> mapper) throws Exception {
```
yihezkel, line 89: testSingleNonNullable

data/src/test/java/com/microsoft/azure/kusto/data/KustoResultMapperTest.java
```diff
@@ -0,0 +1,246 @@
...
+		assertNotNull(pojo);
+		assertEquals(1, pojo.id);
+		assertEquals("MyName", pojo.name);
+		assertEquals(1, pojo.zonedTime.toInstant().toEpochMilli());
+		assertEquals(UUID.fromString("e091cf92-6195-4005-bad5-82af80ff1939"), pojo.uuid);
+		assertEquals(TestEnum.C, pojo.testEnum);
+		assertEquals(2, pojo.instantTime.toEpochMilli());
+	}
+	
+	void testNullable(KustoResultMapper<TestPojo> mapper) throws Exception {
```
yihezkel, line 108: testSingleNullable

yihezkel (changes requested):
This is a great, efficient, flexible and clean solution. Thanks for your contribution!

I requested a few minor changes, and 1 small change to the flow. This is very close.

I apologize for the delay in reviewing until now. I should be able to comment/approve much faster in the next iteration.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
maltalex: Done.

data/src/main/java/com/microsoft/azure/kusto/data/ObjectPopulator.java
maltalex: ok.

data/src/main/java/com/microsoft/azure/kusto/data/ObjectPopulator.java
maltalex: Right, I probably mixed it with the flow from `extractSingle()`. It turns out I wasn't actually using the ordinal passed to `populateFrom` at all except for in the exception text. Fixed it now, please take a look.

maltalex: done.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
maltalex: renamed to `resultObject`

data/src/test/java/com/microsoft/azure/kusto/data/KustoResultMapperTest.java
maltalex: renamed all tests that test "extractSingle" - added "single" to all of them.

data/src/test/java/com/microsoft/azure/kusto/data/KustoResultMapperTest.java
maltalex: renamed all tests that test "extractSingle" - added "single" to all of them.

yihezkel (approved):
As I mentioned elsewhere, this is a clean, flexible and elegant solution. Thank you for your contribution!

maltalex:
Thanks, @yihezkel
Hope others will find it useful.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
```diff
@@ -0,0 +1,132 @@
...
+		final List<KustoResultColumnPopulator<R, ?, ?>> queryResultColumns = new ArrayList<>();
+		final Supplier<R> objConstructor;
+		
+		public Builder(Supplier<R> objConstructor) {
+			this.objConstructor = objConstructor;
+		}
+		
+		/**
+		 * Add a column by name. Using this function will cause the KustoResultMapper to lookup the column by name once per column per call to
+		 * {@link KustoResultMapper##extractList(KustoResultSetTable)} or {@link KustoResultMapper##extractSingle(KustoResultSetTable)}
```
yihezkel, line 26: Javadoc validation is failing - I believe you want:
{@link KustoResultMapper#extractList(KustoResultSetTable)} or {@link KustoResultMapper#extractSingle(KustoResultSetTable)}

yihezkel (commented):
Javadoc validation is failing. We can merge this once all issues identified by the CI are addressed (it's green).

data/src/main/java/com/microsoft/azure/kusto/data/KustoType.java
```diff
@@ -0,0 +1,53 @@
...
+import java.time.ZonedDateTime;
+import java.util.UUID;
+import java.util.function.Function;
+
+/**
+ * Represents various ADX data types and their bindings to Java classes
+ * 
+ * @author alex
+ *
+ * @see https://docs.microsoft.com/azure/data-explorer/kusto/query/scalar-data-types/
```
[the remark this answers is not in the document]
maltalex: done.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultMapper.java
maltalex: okay, changed it.

PR #160, opened 2020-12-22
Rearchitect the AAD authentication code to be object oriented

yihezkel:
#### Pull Request Description

Rearchitect the AAD authentication code to be object oriented and follow best-practices (Java and otherwise)

---

#### Future Release Comment

**Breaking Changes:**
- Moved auth-related classes to new auth package


ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -70,19 +70,33 @@ public static void setUp() {
...
         } catch (Exception ex) {
             Assertions.fail("Failed to drop and create new table", ex);
         }
 
         resourcesPath = Paths.get(System.getProperty("user.dir"), "src", "test", "resources").toString();
         try {
+            Thread.sleep(2000);
             String mappingAsString = new String(Files.readAllBytes(Paths.get(resourcesPath, "dataset_mapping.json")));
-            queryClient.execute(databaseName, String.format(".create table %s ingestion json mapping '%s' '%s'",
+            result = queryClient.execute(databaseName, String.format(".create table %s ingestion json mapping '%s' '%s'",
```
yihezkel, line 99: I added a few of these in some system tests. They're useful for debugging, and it's NBD to leave them in a system test class instead of having to constantly remove and re-add.

data/src/main/java/com/microsoft/azure/kusto/data/auth/AccessTokenTokenProvider.java
```diff
@@ -0,0 +1,24 @@
...
+import com.microsoft.azure.kusto.data.exceptions.DataClientException;
+import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
+import org.jetbrains.annotations.NotNull;
+
+import java.net.URISyntaxException;
+
+public class AccessTokenTokenProvider extends TokenProviderBase {
+    private final String accessToken;
+
+    AccessTokenTokenProvider(@NotNull String accessToken, @NotNull String clusterUrl, String authorityId) throws URISyntaxException {
```
yogilad, line 15: The authId here is redundent.
You may want to be tolerant of authId being null.

The same goes for clusterURL.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.microsoft.aad.msal4j.ConfidentialClientApplication;
+import com.microsoft.aad.msal4j.IClientCertificate;
+import org.jetbrains.annotations.NotNull;
+
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+
+public class ApplicationCertificateTokenProvider extends ConfidentialAppTokenProviderBase {
+    ApplicationCertificateTokenProvider(@NotNull String applicationClientId, @NotNull IClientCertificate clientCertificate, @NotNull String clusterUrl, String authorityId) throws URISyntaxException {
```
yogilad, line 14: Does Java support SNI auth?

data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationKeyTokenProvider.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.microsoft.aad.msal4j.ConfidentialClientApplication;
+import com.microsoft.aad.msal4j.IClientSecret;
+import org.jetbrains.annotations.NotNull;
+
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+
+public class ApplicationKeyTokenProvider extends ConfidentialAppTokenProviderBase {
+    ApplicationKeyTokenProvider(@NotNull String clientId, @NotNull IClientSecret clientSecret, @NotNull String clusterUrl, String authorityId) throws URISyntaxException {
```
yogilad, line 14: applicationClientId

data/src/main/java/com/microsoft/azure/kusto/data/auth/PublicAppTokenProviderBase.java
```diff
@@ -0,0 +1,31 @@
...
+    IPublicClientApplication clientApplication;
+
+    PublicAppTokenProviderBase(@NotNull String clusterUrl, String authorityId) throws URISyntaxException {
+        super(clusterUrl, authorityId);
+    }
+
+    @Override
+    protected IAuthenticationResult acquireAccessTokenSilentlyMsal() throws MalformedURLException, InterruptedException, ExecutionException, TimeoutException {
+        CompletableFuture<Set<IAccount>> accounts = clientApplication.getAccounts();
+        return clientApplication.acquireTokenSilently(getSilentParameters(accounts.join())).get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
yogilad, line 29: I think you need to either user a user hint or make sure you only have one account only

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
```diff
@@ -0,0 +1,104 @@
...
+        if (accessTokenResult == null) {
+            accessTokenResult = acquireNewAccessToken();
+        }
+        return accessTokenResult.accessToken();
+    }
+
+    protected IAuthenticationResult acquireAccessTokenSilently() throws DataServiceException, DataClientException {
+        try {
+            // TODO: MSAL docs say this would throw MsalUiRequiredException, but it doesn't in Java
+            return acquireAccessTokenSilentlyMsal();
```
yogilad, line 69: This design is odd.
You make the base class MSAL aware, then override it out where it is not needed.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
```diff
@@ -0,0 +1,104 @@
...
+        } catch (TimeoutException | ExecutionException e) {
+            return null; // Legitimate outcome, in which case a new token will be acquired
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            return null; // Legitimate outcome, in which case a new token will be acquired
+        }
+    }
+
+    protected IAuthenticationResult acquireAccessTokenSilentlyMsal() throws MalformedURLException, InterruptedException, ExecutionException, TimeoutException, DataServiceException {
+        throw new DataServiceException("Cannot obtain a refresh token for Authentication type '" + this.getClass().getSimpleName() + "'");
```
yogilad, line 81: Ha?
refresh token?
Isn't there a way to abstract this away so that it must be implemented by the sub class

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    protected IAuthenticationResult acquireNewAccessToken() throws DataServiceException, DataClientException {
+        throw new DataServiceException("Cannot obtain a new token for Authentication type '" + this.getClass().getSimpleName() + "'");
```
yogilad, line 85: Same "Ha?"

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
```diff
@@ -0,0 +1,104 @@
...
+
+    SilentParameters getSilentParameters(Set<IAccount> accountSet) {
+        IAccount account = getAccount(accountSet);
+        if (account != null) {
+            return SilentParameters.builder(scopes).account(account).authorityUrl(aadAuthorityUrl).build();
+        }
+        return SilentParameters.builder(scopes).authorityUrl(aadAuthorityUrl).build();
+    }
+
+    IAccount getAccount(Set<IAccount> accountSet) {
```
yogilad, line 96: hint or only one for public

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
```diff
@@ -0,0 +1,104 @@
...
+
+    protected IAuthenticationResult acquireAccessTokenSilentlyMsal() throws MalformedURLException, InterruptedException, ExecutionException, TimeoutException, DataServiceException {
+        throw new DataServiceException("Cannot obtain a refresh token for Authentication type '" + this.getClass().getSimpleName() + "'");
+    }
+
+    protected IAuthenticationResult acquireNewAccessToken() throws DataServiceException, DataClientException {
+        throw new DataServiceException("Cannot obtain a new token for Authentication type '" + this.getClass().getSimpleName() + "'");
+    }
+
+    SilentParameters getSilentParameters(Set<IAccount> accountSet) {
```
yogilad, line 88: Can you unify this and the below functions into one?

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderTokenProvider.java
```diff
@@ -0,0 +1,29 @@
...
+package com.microsoft.azure.kusto.data.auth;
+
+import com.microsoft.azure.kusto.data.exceptions.DataClientException;
+import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
+import org.jetbrains.annotations.NotNull;
+
+import java.net.URISyntaxException;
+import java.util.concurrent.Callable;
+
+public class TokenProviderTokenProvider extends TokenProviderBase {
```
yogilad, line 13: CallbackTokenProvider

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderTokenProvider.java
```diff
@@ -0,0 +1,29 @@
...
+    private final Callable<String> tokenProvider;
+
+    TokenProviderTokenProvider(@NotNull Callable<String> tokenProvider, @NotNull String clusterUrl, String authorityId) throws URISyntaxException {
```
yogilad, line 16: Remove redundant parameters

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public static TokenProviderBase createTokenProvider(@NotNull ConnectionStringBuilder csb) throws URISyntaxException {
+        String clusterUrl = csb.getClusterUrl();
+        String authorityId = csb.getAuthorityId();
+        if (StringUtils.isNotBlank(csb.getApplicationClientId()) && StringUtils.isNotBlank(csb.getApplicationKey())) {
+            String clientId = csb.getApplicationClientId();
+            IClientSecret clientSecret = ClientCredentialFactory.createFromSecret(csb.getApplicationKey());
+            return new ApplicationKeyTokenProvider(clientId, clientSecret, clusterUrl, authorityId);
+        } else if (csb.getX509Certificate() != null && csb.getPrivateKey() != null && StringUtils.isNotBlank(csb.getApplicationClientId())) {
```
yogilad, line 27: else is redundant here and below

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderFactory.java
```diff
@@ -0,0 +1,45 @@
...
+public class TokenProviderFactory {
+    private TokenProviderFactory() {
+        // Hide constructor as this is a Factory
+    }
+
+    public static TokenProviderBase createTokenProvider(@NotNull ConnectionStringBuilder csb) throws URISyntaxException {
+        String clusterUrl = csb.getClusterUrl();
+        String authorityId = csb.getAuthorityId();
+        if (StringUtils.isNotBlank(csb.getApplicationClientId()) && StringUtils.isNotBlank(csb.getApplicationKey())) {
+            String clientId = csb.getApplicationClientId();
```
yogilad, line 24: nit:
Since you do not need else clauses due to return on success,
```
load params 
if (checks)
{
 return
}

load next parameter set
...
```

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
```diff
@@ -0,0 +1,90 @@
...
+            tmp = new URI("http://localhost");
+        } catch (URISyntaxException e) { // This cannot happen, but is necessary to allow for a static final variable whose instantiation can throw an exception
+            tmp = null;
+        }
+        redirectUri = tmp;
+    }
+
+    private final String usernameHint;
+
+    UserPromptTokenProvider(@NotNull String clusterUrl, String authorityId) throws URISyntaxException {
```
yogilad, line 40: Must I provide a an auth id for user prompt?
Is this enforced by kcsb?

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    @Override
+    protected IAuthenticationResult acquireNewAccessToken() throws DataServiceException, DataClientException {
+        IAuthenticationResult result;
+        try {
+            // This is the only auth method that allows the same application to be used for multiple distinct accounts, so reset account cache between sign-ins
+            clientApplication = PublicClientApplication.builder(CLIENT_ID).authority(aadAuthorityUrl).build();
+            CompletableFuture<IAuthenticationResult> future = clientApplication.acquireToken(InteractiveRequestParameters.builder(redirectUri).scopes(scopes).loginHint(usernameHint).build());
```
yogilad, line 61: are you sure this is ok if the hint is not initialized?

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    IAccount getAccount(Set<IAccount> accountSet) {
+        if (StringUtils.isNotBlank(usernameHint)) {
+            return accountSet.stream().filter(u -> usernameHint.equalsIgnoreCase(u.username())).findAny().orElse(null);
+        } else {
+            if (accountSet.isEmpty()) {
+                return null;
+            } else {
+                // Normally we would filter accounts by the user authenticating, but there's only 1 per AadAuthenticationHelper instance
+                return accountSet.iterator().next();
```
yogilad, line 86: One or None.
First is not a good choice as it is random.

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
```diff
@@ -0,0 +1,90 @@
...
+            throw new DataServiceException(clusterUrl, ERROR_ACQUIRING_APPLICATION_ACCESS_TOKEN, e);
+        }
+        if (result == null) {
+            throw new DataServiceException(clusterUrl, "acquireWithUserPrompt got 'null' authentication result");
+        }
+        return result;
+    }
+
+    @Override
+    IAccount getAccount(Set<IAccount> accountSet) {
```
yogilad, line 78: Merge with the base API and accept a name that can be defaulted to null

samples/src/main/java/TableStatus.java
```diff
@@ -19,23 +19,23 @@
...
+            IngestionResult ingestionResult;
+            try (IngestClient client = IngestClientFactory.createClient(csb)) {
+                IngestionProperties ingestionProperties = new IngestionProperties(System.getProperty("dbName"),
+                        System.getProperty("tableName"));
+                ingestionProperties.setIngestionMapping(System.getProperty("dataMappingName"), IngestionMapping.IngestionMappingKind.Json);
+                ingestionProperties.setReportMethod(QueueAndTable);
+                ingestionProperties.setReportLevel(IngestionProperties.IngestionReportLevel.FailuresAndSuccesses);
+                FileSourceInfo fileSourceInfo = new FileSourceInfo(System.getProperty("filePath"), 0);
+                ingestionResult = client.ingestFromFile(fileSourceInfo, ingestionProperties);
+            }
```
yogilad, line 38: Catch?

yogilad (changes requested):

data/src/main/java/com/microsoft/azure/kusto/data/auth/AccessTokenTokenProvider.java
yihezkel: Good catch with authId

I think the clusterUrl should be required in a CSB (even though it's only somewhat enforced currently - I brought this up in another thread elsewhere, I'll look that up). Further, in this MSAL implementation, it's useful to always have it available to include for example in exception handling (even though it's not currently used in all classes' flows).

data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
yihezkel: When I started this, I made a list of what was available in the library and then whittled that list down with Ran and Vladik to decide which ones we wanted. SNI auth either didn't make the first cut (not supported by Java MSAL API), or we didn't want it included.
I'll look up shortly which it was.

data/src/main/java/com/microsoft/azure/kusto/data/auth/PublicAppTokenProviderBase.java
yihezkel: I'm not sure. I'll ask MSAL about this, since I couldn't discern proper behavior from their documentation.

I just looked for another example implementation to see what they do, and at https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java#L721 they take the first one for PublicApps.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
yihezkel: Everything here, from TokenProviderBase and below, is MSAL-aware.

I figured when I did this that it might look strange at first glance, but let's look at it more deeply to understand why I believe it makes more sense.

In all cases where we acquire a token silently, this is the appropriate flow. It's never "overrided out". Now some flows don't acquire tokens silently, in which case this never gets called. And I think it's a design decision whether to get this benefit:
a) in this way in which case some inheritors to whom it's not applicable will get this method, or
b) to add another layer of abstraction so that not only will the flows that don't do silent token retrieval not call this method but also not inherit this method.
I chose (a) because adding another layer of abstraction for only this benefit seemed to make it unnecessarily less maintainable.

By the way, I just want to point out that this method mostly exists to organize exception handling centrally. We want to do this once in the base class, as it's identical in all inheritors.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
yihezkel: Good call, we're not getting a refresh token but following a silent flow. Reworded.

We don't want this to be executed by some subclasses. Per the discussion above, if I add another layer of abstraction, I can remove this from the Base. Not sure that's a good idea, though.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
yihezkel: Same response as above (except nothing to reword)

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
yihezkel: They used to be unified, but I split them in order to allow just getAccount() in UserPromptTokenProvider to be overriden, while reducing code duplication.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
yihezkel: See override in UserPromptTokenProvider.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderFactory.java
yihezkel: Different strokes for different folks -
I always cringe when I see that design because I think the "else"s add significantly to readability (similar to  https://stackoverflow.com/a/3261899).

My philosophy: The code is more for the programmer to read than the compiler.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderFactory.java
yihezkel: Same comment as above

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderTokenProvider.java
yihezkel: Comment duplicates https://github.com/Azure/azure-kusto-java/pull/160/files#r548539978

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yihezkel: The MSAL Java API examples always provide it, and the relevant code (com.microsoft.aad.msal4j.InteractiveRequest.java) requires one to exist - but not necessarily provided. This is because if one is not provided, it defaults to "https://login.microsoftonline.com/common/".

It is not enforced by the KCSB, so null can be provided.

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yihezkel: Yes - I tested it.

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yihezkel: The username is indeed a property only relevant to UserPromptAuth. It's a different implementation.

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yihezkel: There can only be 0 or 1 accounts in the UserPromptTokenProvider. More isn't possible by design.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
yihezkel: I don't think it's supported in the MSAL Java API

yihezkel: When I originally replaced ADL with MSAL in the Java SDK, the latest Java MSAL API (1.7.0) did not support SNI auth.

However, the very next version (1.7.1) added support for it (see https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/219).

I suppose we should add it - good catch! (But in a separate PR to keep this focused.)
Tagging @vladikbr to confirm

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderFactory.java
yihezkel: Here's another interesting discussion about this, which I very much agree with: https://blog.mozilla.org/nnethercote/2009/08/31/no-else-after-return-considered-harmful/

data/src/main/java/com/microsoft/azure/kusto/data/auth/PublicAppTokenProviderBase.java
yogilad: What would happen if I were to use a different identity when prompted for the second time?

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
yogilad: UserToken and Callback do not use MSAL.
MSI would also not use MSAL if and when you support it.

I'd separate this so that so that you have base => MSAL base => MSAL Impls.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
yogilad: See? 
You need base and MSAL Base, and then you will not need to jump over hoops to make the design work

yogilad: i.e. you should implement functionality in rather then override functionality out

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderFactory.java
yogilad: Your call, 
just note you make double calls to property getters and your code is much more verbose then it should be.
Readability-wise you loose points here

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderFactory.java
yogilad: Same response as above :)

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yogilad: So, default = null or comment

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yogilad: Default = null or comment then

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yogilad: I still vote for same impl. with option for null hint

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yogilad: A) you can be prompted again and provide a different user 
B) security is all about everyone making sure their code is as safe as possible, so there's no harm in checking there's only one account

yogilad (changes requested):

data/src/main/java/com/microsoft/azure/kusto/data/auth/PublicAppTokenProviderBase.java
yihezkel: As a precaution, I reset the PublicClientApplication each time. There can only be one account/identity.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
yihezkel: I don't disagree with adding an additional layer of abstraction, but just to point out that this isn't an example of overriding functionality out, since getSilentParameters isn't overrided out. It's just not *used* in the TokenProvider and Token cases.

But I agree with the pros for another layer of abstraction, and though I hadn't done so originally to avoid making the design overly complex, I think you're right that it's worth it.

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yihezkel: Per my comment in UserPromptTokenProvider, I reset the PublicCLientApp to be sure in terms of security.

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yihezkel: I think it's clear because clusterUrl is annotated as @NotNull, and authorityId isn't.

Note that Java doesn't support default method parameter values.

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yihezkel: Same as comment above

samples/src/main/java/TableStatus.java
yihezkel: In Java, a try-with-resources block is effectively a cleaner and more concise way to include a finally block. The catch is optional, and superfluous if unnecessary: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
yogilad: I chose to avoid a 3 level hierarchy as well, but at the cost of repeating myself on the different MSAL implementations.
i.e. my base impl is very very simple and contains abstract functions for get token and get token silent.
If you want to go by the book though, use 3 levels (base, msal base, actual impl of base or msal base)

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yihezkel: I'm a subscriber to the approach of almost no comments to explain "what" (i.e. such explanations are via the code itself), and comments should instead be primarily to explain "why".

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yogilad: API description strings are not comments :)

yihezkel: Sorry I don't follow. I believe it's an actual comment.

yogilad: Many IDE's treat specific comments as an ad-hoc documentation if they have the right format and / or are in the right location, e.g. the /// Summary of C++

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
```diff
@@ -0,0 +1,83 @@
...
+        String aadAuthorityFromEnv = System.getenv("AadAuthorityUri");
+        if (aadAuthorityFromEnv == null) {
+            authorityUrl = String.format("https://login.microsoftonline.com/%s/", authorityId);
+        } else {
+            authorityUrl = String.format("%s%s%s/", aadAuthorityFromEnv, aadAuthorityFromEnv.endsWith("/") ? "" : "/", authorityId);
+        }
+        return authorityUrl;
+    }
+
+    public String acquireAccessToken() throws DataServiceException, DataClientException {
```
yogilad, line 42: Is @override required here as well?

yogilad (approved):

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yihezkel: Sorry, I'm still not following.

I wrote: "Per my comment..."
You responded: "API description strings are not comments", which implies to me that you think I made a Javadoc API description.
So I responded: "I believe it's an actual comment."

My comment is as follows:
// This is the only auth method that allows the same application to be used for multiple distinct accounts, so reset account cache between sign-ins

Are you saying you *want* it to be a Javadoc?

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yihezkel: Do you agree/comfortable with my approach, @yogilad ?

data/src/main/java/com/microsoft/azure/kusto/data/auth/AccessTokenTokenProvider.java
yihezkel: Do you agree/comfortable with my approach, @yogilad ?

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
yihezkel: Good catch

data/src/main/java/com/microsoft/azure/kusto/data/auth/AccessTokenTokenProvider.java
yogilad: I don't think you should request parameters you do not use in practice

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yogilad: I'll let you have this one :)

data/src/main/java/com/microsoft/azure/kusto/data/auth/UserPromptTokenProvider.java
yogilad: Mixed up two conversations.
You can close this one

PR #166, opened 2021-02-25
Add new API to query returning a json result

yihezkel:
#### Pull Request Description

Add new API to query returning a json result. When the caller doesn't need the Java object, this saves time converting to a Java object in the SDK and from the caller converting from the Java object.

---

#### Future Release Comment
**Features:**
- New Query API that returns a json instead of a KustoOperationResult

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Client.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Client.java
@@ -7,10 +7,15 @@ import com.microsoft.azure.kusto.data.exceptions.DataClientException;
 import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
 
 public interface Client {
-
     KustoOperationResult execute(String command) throws DataServiceException, DataClientException;
 
     KustoOperationResult execute(String database, String command) throws DataServiceException, DataClientException;
 
     KustoOperationResult execute(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException;
+
+    String executeForJsonResult(String database) throws DataServiceException, DataClientException;
+
+    String executeForJsonResult(String database, String command) throws DataServiceException, DataClientException;
+
+    String executeForJsonResult(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException;
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Client.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Client.java
@@ -13,9 +13,9 @@ public interface Client {
 
     KustoOperationResult execute(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException;
 
-    String executeForJsonResult(String database) throws DataServiceException, DataClientException;
+    String executeToJsonResult(String database) throws DataServiceException, DataClientException;
 
-    String executeForJsonResult(String database, String command) throws DataServiceException, DataClientException;
+    String executeToJsonResult(String database, String command) throws DataServiceException, DataClientException;
 
-    String executeForJsonResult(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException;
+    String executeToJsonResult(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException;
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -8,6 +8,7 @@ import com.microsoft.azure.kusto.data.auth.TokenProviderBase;
 import com.microsoft.azure.kusto.data.auth.TokenProviderFactory;
 import com.microsoft.azure.kusto.data.exceptions.DataClientException;
 import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
+import com.microsoft.azure.kusto.data.exceptions.KustoServiceError;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.client.utils.URIBuilder;
 import org.json.JSONException;
@@ -50,7 +51,7 @@ public class ClientImpl implements Client, StreamingClient {
         clusterUrl = url;
         aadAuthenticationHelper = TokenProviderFactory.createTokenProvider(csb);
         clientVersionForTracing = "Kusto.Java.Client";
-        String version = Utils.GetPackageVersion();
+        String version = Utils.getPackageVersion();
         if (StringUtils.isNotBlank(version)) {
             clientVersionForTracing += ":" + version;
         }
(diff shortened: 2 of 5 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/Client.java
```diff
@@ -7,10 +7,15 @@
...
 public interface Client {
-
     KustoOperationResult execute(String command) throws DataServiceException, DataClientException;
 
     KustoOperationResult execute(String database, String command) throws DataServiceException, DataClientException;
 
     KustoOperationResult execute(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException;
-}
+
+    String executeForJsonResult(String database) throws DataServiceException, DataClientException;
```
yihezkel, line 16: Naming this was tough. On the one hand it's uncommon to name a method to describe its return format. Worse, here we both do and have no choice but to provide a name that's inconsistent with the other names.

An alternative was "executeReturnJson".

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -73,21 +78,20 @@ public static void setUp() {
     @AfterAll
     public static void tearDown() {
         try {
-            queryClient.execute(databaseName, String.format(".drop table %s ifexists", tableName));
+            queryClient.executeForJsonResult(databaseName, String.format(".drop table %s ifexists", tableName));
```
yihezkel, line 81: For performance, the new default should be to use "executeForJsonResult" for any query where we don't care to see the result.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -173,13 +176,28 @@ private void assertRowCount(int expectedRowsCount) {
                 Thread.sleep(5000);
                 timeoutInSec -= 5;
 
-                result = queryClient.execute(databaseName, String.format("%s | count", tableName));
+                if (checkViaJson) {
+                    String result = queryClient.executeForJsonResult(databaseName, String.format("%s | count", tableName));
+                    JSONArray jsonArray = new JSONArray(result);
+                    JSONObject primaryResult = null;
+                    for (Object o : jsonArray) {
+                        if (o.toString() != null && o.toString().matches(".*\"TableKind\"\\s*:\\s*\"PrimaryResult\".*")) {
```
yihezkel, line 184: This is basically what result.getPrimaryResults() does

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -221,7 +239,7 @@ void testIngestFromFile() {
             } catch (Exception ex) {
                 Assertions.fail(ex);
             }
-            assertRowCount(item.rows);
+            assertRowCount(item.rows, false);
```
yihezkel, line 242: This gives us the chance to add to existing E2E tests checks of both the json approach and the KustoOperationResult approach.

data/src/main/java/com/microsoft/azure/kusto/data/Client.java
yihezkel: @vladikbr suggested "executeWithJsonResult"

vladikbr (commented):
Suggest naming: executeWithJsonResult

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -155,16 +174,31 @@ public KustoOperationResult executeStreamingIngest(String database, String table
         if (timeoutMs == null) {
             timeoutMs = STREAMING_INGEST_TIMEOUT_IN_MILLISECS;
         }
-        return Utils.post(clusterEndpoint, null, stream, timeoutMs.intValue() + CLIENT_SERVER_DELTA_IN_MILLISECS, headers, leaveOpen);
+        String response = Utils.post(clusterEndpoint, null, stream, timeoutMs.intValue() + CLIENT_SERVER_DELTA_IN_MILLISECS, headers, leaveOpen);
+        try {
+            return new KustoOperationResult(response, clusterEndpoint.endsWith("v2/rest/query") ? "v2" : "v1");
```
AsafMah, line 179: This line has some logic in it, and appears both here and in execute. 
Maybe it should be a method, even a public facing one - just in case someone wants to turn their json into KustoOperationResult.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -72,21 +73,39 @@ public KustoOperationResult execute(String database, String command) throws Data
...
         }
         Long timeoutMs = properties == null ? null : properties.getTimeoutInMilliSec();
 
-        String clusterEndpoint;
-        if (command.startsWith(ADMIN_COMMANDS_PREFIX)) {
-            clusterEndpoint = String.format("%s/%s/rest/mgmt", clusterUrl, MGMT_ENDPOINT_VERSION);
-            if (timeoutMs == null) {
+        String clusterEndpoint = determineClusterEndpoint(command);
+        if (timeoutMs == null) {
+            if (command.startsWith(ADMIN_COMMANDS_PREFIX)) {
```
AsafMah, line 106: Maybe out of scope for this PR - but maybe we want to trim the command first? 
If someone starts an admin command with a space or tab they would get the wrong results

AsafMah, line 106: One more thing, but I don't have a good solution for this.

It's a great thing that you exported the logic to determineClusterEndpoint, but then you still have to call command.startsWith(ADMIN_COMMANDS_PREFIX) in your code, which makes it a bit redundant

data/src/main/java/com/microsoft/azure/kusto/data/Client.java
AsafMah: I think that executeWithJsonResult is the best option here, with executeForJsonResult at close second

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
ohadbitt: Actually its always false here ..

data/src/main/java/com/microsoft/azure/kusto/data/Client.java
yihezkel: I think that "with" implies the json is an input parameter, and that "for" implies the json is an output parameter (which is the case).

I'm outvoted, so I'll change before submitting unless someone agrees with this logic first.

yihezkel: How about "executeToJsonResult"?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
yihezkel: You can probably see I tried to get rid of that redundancy, but ran into the same issue you're calling out.

The root cause is that Java on principle demands that a method return 1 thing. Therefore if the same logic affects multiple parameters, we'd either create a new object or, if the logic is non-expensive, repeat the logic. I chose the latter.

We could add a 1-line method called, "isAdminCommand()" that obscures this repeated line, but all that does is add (little, IMHO) clarity. The logic would still be duplicated.

WDYT?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
yihezkel: I really wanted to add a new KustoOperationResult constructor that accepts a URL so that this logic would be delegated to there. I think that conceptually makes most sense.
But I saw adding a new KustoOperationResult constructor as out of scope for this PR.

@ohadbitt if it's necessarily false, should I remove it altogether?

data/src/main/java/com/microsoft/azure/kusto/data/Client.java
yihezkel: @vladikbr @AsafMah @ohadbitt 
Per Sheldon's confirmation a couple hours ago that he'd like this API as it is here, I'd like to conclude on the proper naming so we can merge this.

My order of preference is:
1) executeToJsonResult
2) executeForJsonResult
3) executeWithJsonResult

AsafMah: To is fine

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
yihezkel: Sorry I missed that, I see now that the cluster endpoint never ends in "v2/rest/query" and is always "v1".

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
AsafMah: I meant having an isAdminCommand method

AsafMah (approved):

PR #167, opened 2021-03-02
Optimize and upgrade dependency versions

yihezkel:
#### Pull Request Description
Mostly transparent optimizations

---

#### Future Release Comment
**Features:**
- Optimize and upgrade dependency versions
- Minor code optimizations

**Fixes:**
- Resolve mvn warnings
- Remove dependency duplicates/conflicts

### DIFF
--- a/data/pom.xml
+++ b/data/pom.xml
@@ -94,9 +94,6 @@
                 <version>${maven-dependency-plugin.version}</version>
                 <configuration>
                     <failOnWarning>true</failOnWarning>
-                    <ignoredUnusedDeclaredDependencies>
-                        <dependency>org.junit.jupiter:junit-jupiter-engine</dependency>
-                    </ignoredUnusedDeclaredDependencies>
                 </configuration>
                 <executions>
                     <execution>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -35,32 +35,33 @@
...
-        <httpcore.version>4.4.11</httpcore.version>
-        <json.version>20190722</json.version>
-        <fasterxml.jackson.core.version>[2.11.0,2.12.0)</fasterxml.jackson.core.version>
-        <azure-storage.version>8.3.0</azure-storage.version>
+        <commons-lang3.version>3.11</commons-lang3.version>
+        <msal4j.version>1.9.1</msal4j.version>
+        <httpclient.version>4.5.13</httpclient.version>
+        <httpcore.version>4.4.13</httpcore.version>
+        <json.version>20201115</json.version>
+        <fasterxml.jackson.core.version>2.12.1</fasterxml.jackson.core.version>
```
yihezkel, line 45: Previously was [2.11.0,2.12.0) - why?

ingest/pom.xml
```diff
@@ -88,8 +88,6 @@
                     <ignoredUnusedDeclaredDependencies>
                         <dependency>org.xerial:sqlite-jdbc</dependency>
                         <dependency>org.slf4j:slf4j-simple</dependency>
-                        <dependency>org.junit.jupiter:junit-jupiter-engine</dependency>
```
yihezkel, line 91: Why do we need this? The first is already scoped for tests only, and if the second is just for test scope why not scope for tests only?

data/pom.xml
```diff
@@ -94,9 +94,6 @@
                 <version>${maven-dependency-plugin.version}</version>
                 <configuration>
                     <failOnWarning>true</failOnWarning>
-                    <ignoredUnusedDeclaredDependencies>
-                        <dependency>org.junit.jupiter:junit-jupiter-engine</dependency>
```
yihezkel, line 98: Why do we need this? It's already scoped for tests only

pom.xml
ohadbitt: i dont know but i do know that all these jackson libraries are a big mess of versions

ohadbitt (approved):

PR #170, opened 2021-04-05
Add streaming query API

yihezkel:
#### Pull Request Description

Add Query API that streams the raw json (not the KustoOperationResult) for maximum performance in the use case of significant data.

---

#### Future Release Comment
**Features:**
- Add Query API that streams the raw json (not the KustoOperationResult) for maximum performance in the use case of significant data.

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -106,13 +106,13 @@ public class ClientImpl implements Client, StreamingClient {
         command = command.trim();
 
         String clusterEndpoint = determineClusterEndpoint(command);
-        Long timeoutMs = determineTimeout(properties, command);
+        long timeoutMs = determineTimeout(properties, command);
 
         Map<String, String> headers = generateIngestAndCommandHeaders(properties, "KJC.execute");
         addCommandHeaders(headers);
         String jsonPayload = generateCommandPayload(database, command, properties, clusterEndpoint);
 
-        return Utils.post(clusterEndpoint, jsonPayload, null, timeoutMs.intValue() + CLIENT_SERVER_DELTA_IN_MILLISECS, headers, false);
+        return Utils.post(clusterEndpoint, jsonPayload, null, timeoutMs + CLIENT_SERVER_DELTA_IN_MILLISECS, headers, false);
     }
 
     @Override
@@ -151,7 +151,7 @@ public class ClientImpl implements Client, StreamingClient {
         if (timeoutMs == null) {
             timeoutMs = STREAMING_INGEST_TIMEOUT_IN_MILLISECS;
         }
-        String response = Utils.post(clusterEndpoint, null, stream, timeoutMs.intValue() + CLIENT_SERVER_DELTA_IN_MILLISECS, headers, leaveOpen);
+        String response = Utils.post(clusterEndpoint, null, stream, timeoutMs + CLIENT_SERVER_DELTA_IN_MILLISECS, headers, leaveOpen);
         try {
             return new KustoOperationResult(response, "v1");
         } catch (KustoServiceError e) {
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -10,6 +10,7 @@ import com.microsoft.azure.kusto.data.exceptions.DataClientException;
 import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
 import com.microsoft.azure.kusto.data.exceptions.KustoServiceError;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.http.HttpHeaders;
 import org.apache.http.client.utils.URIBuilder;
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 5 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -81,12 +79,116 @@ static String post(String url, String payload, InputStream stream, Integer timeo
...
                 }
             }
-        } catch (JSONException | IOException | URISyntaxException e) {
+        } catch (JSONException | IOException e) {
             throw new DataClientException(url, "Error in post request:" + e.getMessage(), e);
         }
         return null;
     }
 
+    static InputStream postToStreamingOutput(String url, String payload, long timeoutMs, Map<String, String> headers) throws DataServiceException, DataClientException {
```
yihezkel, line 88: Note that I intentionally don't allow the query to be streamed in via InputStream. It also doesn't make sense to offer a "leaveOpen" flag for the response's InputStream.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -104,40 +105,14 @@ public String executeToJsonResult(String database, String command, ClientRequest
         }
         command = command.trim();
 
-        Long timeoutMs = properties == null ? null : properties.getTimeoutInMilliSec();
-
         String clusterEndpoint = determineClusterEndpoint(command);
-        if (timeoutMs == null) {
-            if (command.startsWith(ADMIN_COMMANDS_PREFIX)) {
```
amitof, line 111: Can the command start with a comment (rather than a dot) ?

data/src/main/java/com/microsoft/azure/kusto/data/CloseParentResourcesStream.java
```diff
@@ -0,0 +1,78 @@
...
+ * It takes a stream, and simply forwards the call to all of its methods, except for the {@code close()} method - which does nothing.
+ */
+public class CloseParentResourcesStream extends InputStream {
+    private final InputStream innerStream;
+    private final CloseableHttpClient httpClient;
+    private final CloseableHttpResponse httpResponse;
+
+    public CloseParentResourcesStream(InputStream innerStream, CloseableHttpClient httpClient, CloseableHttpResponse httpResponse) {
+        this.innerStream = innerStream;
+        this.httpClient = httpClient;
```
amitof, line 25: Does this mean that this is creating a new HttpClient on every request? would it be better to re-use the same client if possible (if no concurrent requests are being made,  like HttpClientFactory in Kusto.Data)

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -81,12 +79,116 @@ static String post(String url, String payload, InputStream stream, Integer timeo
                     throw new DataServiceException(url, message, ex);
                 }
             }
-        } catch (JSONException | IOException | URISyntaxException e) {
+        } catch (JSONException | IOException e) {
             throw new DataClientException(url, "Error in post request:" + e.getMessage(), e);
```
amitof, line 83: Should the HttpClient and HttpResponse be disposed here?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -81,12 +79,116 @@ static String post(String url, String payload, InputStream stream, Integer timeo
...
+                        return new DeflaterInputStream(contentStream);
+                    }
+                }
+                // Though the server responds with a gzip/deflate Content-Encoding header, we reach here because httpclient uses LazyDecompressingStream which handles the above logic
+                return contentStream;
+            } else if ((status == HttpStatus.SC_MOVED_TEMPORARILY || status == HttpStatus.SC_TEMPORARY_REDIRECT)
+                    && redirectCount + 1 <= MAX_REDIRECT_COUNT) {
+                Optional<Header> redirectLocation = Arrays.stream(httpResponse.getHeaders(HttpHeaders.LOCATION)).findFirst();
+                if (redirectLocation.isPresent() && !redirectLocation.get().getValue().equals(url)) {
+                    return postToStreamingOutput(redirectLocation.get().getValue(), payload, timeoutMs, headers, redirectCount + 1);
```
amitof, line 126: Should httpClient be closed here?

amitof (approved):

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -206,4 +198,56 @@ private String determineClusterEndpoint(String command) {
         }
         return clusterEndpoint;
     }
+
+    private long determineTimeout(ClientRequestProperties properties, String command) {
+        Long timeoutMs = properties == null ? null : properties.getTimeoutInMilliSec();
+        if (timeoutMs == null) {
+            if (command.startsWith(ADMIN_COMMANDS_PREFIX)) {
```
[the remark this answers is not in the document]
yihezkel: I don't think that's supported in any use case. A comment would cause the entire command to be commented out. Same in Kusto.Explorer

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
yihezkel: Yes, it ideally should, but we're doing something unique here.

Since we're passing the InputStream externally for the user to consume, I had 2 options:
1. After calling async "httpClient.execute", we could get the CompletableFuture, on which we can call get(). This is a blocking call, and would first read the entire stream from the endpoint, at which point we'd close the httpClient and httpResponse, and only then pass the InputStream. By this point, the InputStream is no longer a true stream from the endpoint. This negates the performance improvement the customer is seeking, as they would like to stream the data as it's received from their ADX cluster.
2. Pass the InputStream as it's streaming from the ADX cluster. If we close the httpClient or the httpResponse in the Utils POST call (before returning the InputStream to the caller and before they consume all its data), reading from the stream will result in an exception that it cannot read from the stream because it's closed. Yet we MUST close those 2 resources. So my solution was to have the resources close AFTER the caller finishes consuming the data from the InputStream. To do this, we should only have to tell the caller they're expected to close the InputStream. Per https://hc.apache.org/httpcomponents-client-4.5.x/current/tutorial/html/fundamentals.html closing the stream closes the other resources: "In order to ensure proper release of system resources one must close either the content stream associated with the entity or the response itself.". However, to be extra sure, I wrap the InputStream in an EofSensorInputStream, which closes the stream on its own when all of the data is consumed. Lastly, to be extra sure, I created a new InputStream object to wrap that which includes the HttpClient and HttpResponse objects, and closes those as well explicitly when it's closed.

I included this in my comment at https://github.com/Azure/azure-kusto-java/pull/170/files#diff-9b40b1e0eb4f281e2b999bcc168625e2979a8052a985db6a172c255b0768a1f1R12

data/src/main/java/com/microsoft/azure/kusto/data/CloseParentResourcesStream.java
yihezkel: I agree, and even saw in the Apache HttpClient docs: "Generally it is recommended to have a single instance of HttpClient per communication component or even per application. However, if the application makes use of HttpClient only very infrequently, and keeping an idle instance of HttpClient in memory is not warranted, it is highly recommended to explicitly shut down the multithreaded connection manager prior to disposing the HttpClient instance. This will ensure proper closure of all HTTP connections in the connection pool."

However, the existing POST is implemented this way, and since this feature was requested ASAP, I considered fixing that to be out of scope.

I'll add as an issue for a future enhancement that both of these should reuse the HttpClient. It indeed supports concurrent requests, but we'll still need a Factory because it can be created with a specified timeout, and we'd need to create an HttpClient per-timeout.

WDYT?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
yihezkel: Yes, thanks!

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
yihezkel: Sorry, I misunderstood your comment. Yes, we need to close the resources in the Exception.

data/src/main/java/com/microsoft/azure/kusto/data/CloseParentResourcesStream.java
yihezkel: Per @ohadbitt 's suggestion, I'll also add a TODO to this effect.

ohadbitt:
in regular execute - you should look for the streaming option in the ClientRequestProperties and if set throw an exception telling the user to use the Streaming API

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -81,12 +76,156 @@ static String post(String url, String payload, InputStream stream, Integer timeo
...
+                        DeflaterInputStream deflaterInputStream = new DeflaterInputStream(contentStream);
+                        returnInputStream = true;
+                        return deflaterInputStream;
+                    }
+                }
+                // Though the server responds with a gzip/deflate Content-Encoding header, we reach here because httpclient uses LazyDecompressingStream which handles the above logic
+                returnInputStream = true;
+                return contentStream;
+            }
+
```
ohadbitt, line 127: As talked - we should expose the internal exception 
Add here 
            DataWebException exc = new DataWebException(EntityUtils.toString(httpResponse.getEntity()), httpResponse);
and throw it later

ohadbitt (approved):

PR #173, opened 2021-04-25
Pom repo url

ohadbitt:
Found out we dont publish well our repo url
 https://search.maven.org/artifact/com.microsoft.azure.kusto/kusto-data/2.4.1/jar


data/pom.xml
```diff
@@ -6,7 +6,11 @@
     <artifactId>kusto-data</artifactId>
     <name>Kusto data client library</name>
     <description>Kusto client library for executing queries and retrieving data</description>
-
+    <url>https://github.com/Azure/azure-kusto-java/tree/master/data</url>
+    <scm>
+        <url>scm:git:https://github.com/Azure/azure-kusto-java</url>
```
yihezkel, line 11: This is an interesting use case. We have 2 artifacts, both within the same git repo.

- What use case does this change come to address?
- Does this git URL achieve that intended goal, considering that it doesn't quite point to the repo specific to the artifact?

yihezkel (approved):

data/pom.xml
ohadbitt: urls here are wrong
https://search.maven.org/artifact/com.microsoft.azure.kusto/kusto-data/2.4.1/jar

PR #174, opened 2021-04-26
Exceptions handling

ohadbitt:
Handling OneApiErrors occurring in the middle of rows reading .
Make sure DataServiceError is thrown in case of transient error and reflect this permanency

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -80,7 +80,12 @@ public class ClientImpl implements Client, StreamingClient {
         try {
             return new KustoOperationResult(response, clusterEndpoint.endsWith("v2/rest/query") ? "v2" : "v1");
         } catch (KustoServiceError e) {
-            throw new DataClientException(clusterEndpoint, "Error converting json response to KustoOperationResult:" + e.getMessage(), e);
+            if (e.getExceptions().isEmpty()){
+                throw new DataClientException(clusterEndpoint,
+                        "Error parsing json response as KustoOperationResult:" + e.getMessage(), e);
+            } else {
+                throw new DataServiceException(e.getMessage(), e);
+            }
         }
     }
 
--- a/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
@@ -36,6 +36,7 @@ public class KustoResultSetTable implements ResultSet {
     private static final String COLUMN_TYPE_SECOND_PROPERTY_NAME = "DataType";
     private static final String ROWS_PROPERTY_NAME = "Rows";
     private static final String EXCEPTIONS_PROPERTY_NAME = "Exceptions";
+    private static final String EXCEPTIONS_MESSAGE = "Query execution failed with inner exceptions";
 
     private final List<List<Object>> rows;
     private String tableName;
@@ -107,11 +108,11 @@ public class KustoResultSetTable implements ResultSet {
                             String message = exceptions.getString(0);
                             throw new KustoServiceError(message);
                         } else {
-                            throw new KustoServiceError(exceptions, false);
+                            throw new KustoServiceError(exceptions, false, EXCEPTIONS_MESSAGE);
                         }
                     } else {
                         throw new KustoServiceError(((JSONObject) row).getJSONArray(
-                                "OneApiErrors"),true);
+                                "OneApiErrors"),true, EXCEPTIONS_MESSAGE);
                     }
                 }
                 JSONArray rowAsJsonArray = jsonRows.getJSONArray(i);
--- a/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
@@ -3,7 +3,9 @@
 
 package com.microsoft.azure.kusto.data;
 
+import com.microsoft.azure.kusto.data.exceptions.DataWebException;
 import com.microsoft.azure.kusto.data.exceptions.KustoServiceError;
+import com.microsoft.azure.kusto.data.exceptions.OneApiError;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.time.FastDateFormat;
 import org.json.JSONArray;
@@ -105,8 +107,11 @@ public class KustoResultSetTable implements ResultSet {
                             String message = exceptions.getString(0);
                             throw new KustoServiceError(message);
                         } else {
-                            throw new KustoServiceError(exceptions);
+                            throw new KustoServiceError(exceptions, false);
                         }
+                    } else {
+                        throw new KustoServiceError(((JSONObject) row).getJSONArray(
+                                "OneApiErrors"),true);
                     }
                 }
                 JSONArray rowAsJsonArray = jsonRows.getJSONArray(i);
--- a/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataClientException.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataClientException.java
@@ -3,6 +3,9 @@
 
 package com.microsoft.azure.kusto.data.exceptions;
 
+/*
+  This class represents an error that happened on the client side and is therefore considered permanent
+ */
 public class DataClientException extends Exception {
     private String ingestionSource;
 
@@ -21,5 +24,9 @@ public class DataClientException extends Exception {
     public DataClientException(String ingestionSource, String message, Exception exception) {
         super(message, exception);
         this.ingestionSource = ingestionSource;
+    }
+
+    public Boolean isPermanent(){
+        return true;
     }
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataWebException.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataWebException.java
@@ -9,7 +9,7 @@ import org.json.JSONObject;
 public class DataWebException extends Exception{
 
     private String message;
-    private HttpResponse httpResponse;
+    private final HttpResponse httpResponse;
     private OneApiError apiError;
 
     public String getMessage() { return message; }
@@ -24,22 +24,22 @@ public class DataWebException extends Exception{
 
     public OneApiError getApiError() {
         if (apiError == null) {
-            JSONObject jsonObject = new JSONObject(getMessage());
-            apiError = createApiError(jsonObject);
+            JSONObject jsonObject = new JSONObject(getMessage()).getJSONObject("error");
+            apiError = new OneApiError(
+                    jsonObject.getString("code"),
+                    jsonObject.getString("message"),
+                    jsonObject.getString("@message"),
+                    jsonObject.getString("@type"),
+                    jsonObject.getJSONObject("@context"),
+                    jsonObject.getBoolean("@permanent")
+            );
         }
 
         return apiError;
     }
 
-    static public OneApiError createApiError(JSONObject obj){
-        JSONObject jsonObject = obj.getJSONObject("error");
-        return new OneApiError(
-                jsonObject.getString("code"),
-                jsonObject.getString("message"),
-                jsonObject.getString("@message"),
-                jsonObject.getString("@type"),
-                jsonObject.getJSONObject("@context"),
-                jsonObject.getBoolean("@permanent")
-        );
+    @Override
+    public String toString() {
+        return message;
     }
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataWebException.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataWebException.java
@@ -24,17 +24,22 @@ public class DataWebException extends Exception{
 
     public OneApiError getApiError() {
         if (apiError == null) {
-            JSONObject jsonObject = new JSONObject(getMessage()).getJSONObject("error");
-            apiError = new OneApiError(
-                    jsonObject.getString("code"),
-                    jsonObject.getString("message"),
-                    jsonObject.getString("@message"),
-                    jsonObject.getString("@type"),
-                    jsonObject.getJSONObject("@context"),
-                    jsonObject.getBoolean("@permanent")
-            );
+            JSONObject jsonObject = new JSONObject(getMessage());
+            apiError = createApiError(jsonObject);
         }
 
         return apiError;
+    }
+
+    static public OneApiError createApiError(JSONObject obj){
+        JSONObject jsonObject = obj.getJSONObject("error");
+        return new OneApiError(
+                jsonObject.getString("code"),
+                jsonObject.getString("message"),
+                jsonObject.getString("@message"),
+                jsonObject.getString("@type"),
+                jsonObject.getJSONObject("@context"),
+                jsonObject.getBoolean("@permanent")
+        );
     }
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
@@ -12,7 +12,8 @@ import java.util.List;
 public class KustoServiceError extends Exception {
     private ArrayList<Exception> exceptions = null;
 
-    public KustoServiceError(JSONArray exceptions, boolean isOneApi) throws JSONException {
+    public KustoServiceError(JSONArray exceptions, boolean isOneApi, String message) throws JSONException {
+        this(message);
         this.exceptions = new ArrayList<>();
         if (isOneApi){
             for (int j = 0; j < exceptions.length(); j++) {
@@ -31,5 +32,10 @@ public class KustoServiceError extends Exception {
 
     public List<Exception> getExceptions() {
         return exceptions;
+    }
+
+    @Override
+    public String toString() {
+        return "{\"exceptions\":" + exceptions + "}";
     }
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
@@ -9,6 +9,9 @@ import org.json.JSONException;
 import java.util.ArrayList;
 import java.util.List;
 
+/*
+  This class represents an error that returned from the service
+ */
 public class KustoServiceError extends Exception {
     private ArrayList<Exception> exceptions = null;
 
@@ -37,5 +40,16 @@ public class KustoServiceError extends Exception {
     @Override
     public String toString() {
         return "{\"exceptions\":" + exceptions + "}";
+    }
+
+    /*
+      Can return null if permanency is not known
+     */
+    public Boolean isPermanent(){
+        if (exceptions != null && exceptions.size() > 0 && exceptions.get(0) instanceof DataWebException) {
+            return ((DataWebException) exceptions.get(0)).getApiError().isPermanent();
+        }
+
+        return null;
     }
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
@@ -12,13 +12,19 @@ import java.util.List;
 public class KustoServiceError extends Exception {
     private ArrayList<Exception> exceptions = null;
 
-    public KustoServiceError(JSONArray exceptions) throws JSONException {
+    public KustoServiceError(JSONArray exceptions, boolean isOneApi) throws JSONException {
         this.exceptions = new ArrayList<>();
-        for (int j = 0; j < exceptions.length(); j++) {
-            this.exceptions.add(new Exception(exceptions.getString(j)));
+        if (isOneApi){
+            for (int j = 0; j < exceptions.length(); j++) {
+                this.exceptions.add(new DataWebException(exceptions.getJSONObject(j).toString(), null));
+            }
+        } else {
+            for (int j = 0; j < exceptions.length(); j++) {
+                this.exceptions.add(new Exception(exceptions.getString(j)));
+            }
         }
-
     }
+
     public KustoServiceError(String message) {
         super(message);
     }


data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -34,6 +36,7 @@
     private static final String COLUMN_TYPE_SECOND_PROPERTY_NAME = "DataType";
     private static final String ROWS_PROPERTY_NAME = "Rows";
     private static final String EXCEPTIONS_PROPERTY_NAME = "Exceptions";
+    private static final String EXCEPTIONS_MESSAGE = "Query execution failed with inner exceptions";
```
yihezkel, line 39: "Query execution failed with **multiple** inner exceptions"

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataClientException.java
```diff
@@ -22,4 +25,8 @@ public DataClientException(String ingestionSource, String message, Exception exc
         super(message, exception);
         this.ingestionSource = ingestionSource;
     }
+
+    public Boolean isPermanent(){
```
yihezkel, line 29: Not in scope for this change, but ideally we'd have a parent interface that defines isPermanent, and the Client and Service exceptions would both inherit and provide their implementations.

yihezkel (changes requested):

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -80,7 +80,12 @@ public KustoOperationResult execute(String database, String command, ClientReque
         try {
             return new KustoOperationResult(response, clusterEndpoint.endsWith("v2/rest/query") ? "v2" : "v1");
         } catch (KustoServiceError e) {
-            throw new DataClientException(clusterEndpoint, "Error converting json response to KustoOperationResult:" + e.getMessage(), e);
+            if (e.getExceptions().isEmpty()){
+                throw new DataClientException(clusterEndpoint,
```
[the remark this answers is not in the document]
ohadbitt: yea this seems wrong - dont recall my logic here- this should be in a catch clause for any other exception

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoClientException.java
```diff
@@ -0,0 +1,16 @@
...
+    public KustoClientException(String message) {
+        super(message);
+    }
+
+    public KustoClientException(String message, Exception exception) {
+        super(message, exception);
+    }
+
+    /*
+      Can return null if permanency is not known
```
AsafMah, line 13: Not a big fan of a Boolean that is also a tristate,
maybe this should be an enum/constants

samples/src/main/java/Query.java
```diff
@@ -14,6 +11,16 @@
...
+
+            ConnectionStringBuilder engineKcsb = ConnectionStringBuilder.createWithUserPrompt("https://kuskusprod" +
+                    ".kusto.windows.net");
+            Client client1 = ClientFactory.createClient(engineKcsb);
+//            client1.execute("KusKus", "KustoLogs | ");
+
+                    client1.execute("KusKus", "KustoLogs\n| where Timestamp between (datetime(2021-04-21T11:40:28)" +
+                    "..2h)\n| where Source !has \"MANAGE\" | where SourceId == \"6343AB81\" \n| extend queues = " +
+                    "todynamic(extract(\"TracePushQueueDetails:([^\\\\]]+\\\\])\", 1, EventText))");
+
```
AsafMah, line 23: Why was this added?

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
```diff
@@ -9,21 +9,47 @@
...
+    public String toString() {
+        return "{\"exceptions\":" + exceptions + "}";
+    }
+
+    /*
+      Can return null if permanency is not known
+     */
+    public Boolean isPermanent(){
+        if (exceptions != null && exceptions.size() > 0 && exceptions.get(0) instanceof DataWebException) {
+            return ((DataWebException) exceptions.get(0)).getApiError().isPermanent();
```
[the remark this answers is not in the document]
ohadbitt: i dont think - in the .Net client they only take the first

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
```diff
@@ -9,21 +9,47 @@
...
         super(message);
     }
 
     public List<Exception> getExceptions() {
         return exceptions;
     }
+
+    @Override
+    public String toString() {
+        return "{\"exceptions\":" + exceptions + "}";
```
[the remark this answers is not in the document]
ohadbitt: now it shouldnt be

samples/src/main/java/Query.java
ohadbitt: reverted

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/TriState.java
```diff
@@ -0,0 +1,7 @@
+package com.microsoft.azure.kusto.data.exceptions;
+
+public enum TriState {
+    TRUE,
+    FALSE,
+    DONTKNOW
```
AsafMah, line 6: DONT_KNOW or UNKNOWN

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoClientException.java
```diff
@@ -0,0 +1,17 @@
...
+        super(message);
+    }
+
+    public KustoClientException(String message, Exception exception) {
+        super(message, exception);
+    }
+
+    public abstract TriState isPermanent();
+
+    public static TriState triStateFromBool(boolean bool){
```
AsafMah, line 14: Move this method to inside `TriState`, and then you can remove the prefix and name it `fromBool`.

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/TriState.java
yihezkel: +1 for UNKNOWN

ohadbitt: DONT_KNOW is from the tristate in our .Net

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
```diff
@@ -9,21 +9,47 @@
 import java.util.ArrayList;
 import java.util.List;
 
+/*
+  This class represents an error that returned from the service
```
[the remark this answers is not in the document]
yihezkel: I still don't see the need for this class.

The KustoServiceError is our own, and only thrown in 1 block of code, in KustoResultSetTable. It's never used anywhere else, except to be caught, also in only 1 place - ClientImpl.execute(). And there, all we do is just rethrow with a DataServiceException!

So I'm not understanding a reason for KustoServiceError to exist. If this was me, I'd allow DataServiceException to accept multiple errors. Also, I'd change DataServiceException to have the "ingestionSource" as a settable property. This way, ClientImpl can set this value and then rethrow. It seems much cleaner to me (and also saves some resource creation).

What do you see as the purpose of a KustoServiceError as opposed to a DataServiceException?

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
```diff
@@ -9,21 +9,46 @@
...
+ */
 public class KustoServiceError extends Exception {
-    private ArrayList<Exception> exceptions = null;
+    private ArrayList<Exception> exceptions;
```
yihezkel, line 16: ArrayList -> List

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
```diff
@@ -9,21 +9,46 @@
...
     }
+
+    @Override
+    public String toString() {
+        return exceptions.size() == 0 ? getMessage() : "exceptions\":" + exceptions + "}";
+    }
+
+
+    public TriState isPermanent(){
+        if (exceptions.size() > 0 && exceptions.get(0) instanceof DataWebException) {
```
yihezkel, line 48: CollectionUtils.isEmpty()

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
```diff
@@ -9,21 +9,46 @@
...
-    private ArrayList<Exception> exceptions = null;
+    private ArrayList<Exception> exceptions;
 
-    public KustoServiceError(JSONArray exceptions) throws JSONException {
+    public KustoServiceError(JSONArray jsonExceptions, boolean isOneApi, String message) throws JSONException {
+        this(message);
         this.exceptions = new ArrayList<>();
-        for (int j = 0; j < exceptions.length(); j++) {
-            this.exceptions.add(new Exception(exceptions.getString(j)));
+        if (isOneApi){
```
yihezkel, line 21: Nit: I would move this if to *inside* the for loop, to eliminate the duplicated for loop for readability.

Isn't this less performant? JIT compiler to the rescue! When it sees loop-invariant conditions, it optimizes them to outside of the loop. So we get to have the best of both worlds - code that's more readable, and more performant.

pom.xml
```diff
@@ -12,8 +12,12 @@
     <scm>
         <url>scm:git:https://github.com/Azure/azure-kusto-java</url>
         <connection>scm:git:https://github.com/Azure/azure-kusto-java</connection>
+
```
yihezkel, line 15: Nit: why the empty space? Same for line 19

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
```diff
@@ -9,21 +9,46 @@
...
         return exceptions;
     }
+
+    @Override
+    public String toString() {
+        return exceptions.size() == 0 ? getMessage() : "exceptions\":" + exceptions + "}";
+    }
+
+
+    public TriState isPermanent(){
```
yihezkel, line 47: I think this is another support for why this class is unnecessary. It doesn't inherit from KustoClientException, so it duplicates this "isPermanent()" concept. Yet if we merge this with DataServiceException, this is included, and we won't have this conceptual duplication of this concept of permanence in separate Kusto exception classes.

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoClientException.java
```diff
@@ -0,0 +1,13 @@
+package com.microsoft.azure.kusto.data.exceptions;
+
+public abstract class KustoClientException extends Exception {
```
yihezkel, line 3: DataServiceException inherits from KustoClientException. I assume the motivation is to indicate that this Exception is part of the Java SDK (client). However, when its 2 concrete implementations are either DataServiceException or DataClientException, I think the ambiguous term "Client" can be confusing.

I think it would be better if this class was called DataException (or KustoDataException?).

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoClientException.java
```diff
@@ -0,0 +1,13 @@
...
+    public KustoClientException(String message) {
+        super(message);
+    }
+
+    public KustoClientException(String message, Exception exception) {
+        super(message, exception);
+    }
+
+    public abstract TriState isPermanent();
```
yihezkel, line 12: I think permanence is conceptually a property of this class ("permanent"), as each instance has its own value.

I would therefore make this class concrete instead  of abstract, include a property called "permanent", and this would be a concrete implementation of isPermanent() and simply return the value of the property.

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataServiceException.java
```diff
@@ -3,26 +3,30 @@
 
 package com.microsoft.azure.kusto.data.exceptions;
 
-public class DataServiceException extends Exception {
+public class DataServiceException extends KustoClientException {
     private String ingestionSource;
-
+    private TriState isPermanent;
```
yihezkel, line 8: I'd remove this, and use the inherited property.

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataServiceException.java
```diff
@@ -3,26 +3,30 @@
...
     public String getIngestionSource() { return ingestionSource; }
 
     public DataServiceException(String message) {
```
yihezkel, line 11: Nit: This method is never used in the code. Remove?

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataServiceException.java
```diff
@@ -3,26 +3,30 @@
...
-        super(message);
+        this(null, message, TriState.DONT_KNOW);
     }
 
-    public DataServiceException(String message, Exception exception) {
-        super(message, exception);
+    public DataServiceException(String message, Exception exception, TriState isPermanent) {
```
yihezkel, line 15: Nit: This method is never used in the code. Remove?

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataServiceException.java
```diff
@@ -3,26 +3,30 @@
...
+    public DataServiceException(String ingestionSource, String message, Exception exception, TriState isPermanent) {
         super(message, exception);
         this.ingestionSource = ingestionSource;
+        this.isPermanent = isPermanent;
     }
 
-    public DataServiceException(String ingestionSource, String message) {
-        super(message);
-        this.ingestionSource = ingestionSource;
+    public TriState isPermanent(){
```
yihezkel, line 29: This can inherit from its parent if it has the concrete implementation I suggested.

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataWebException.java
```diff
@@ -9,7 +9,7 @@
 public class DataWebException extends Exception{
 
     private String message;
-    private HttpResponse httpResponse;
+    private final HttpResponse httpResponse;
     private OneApiError apiError;
 
     public String getMessage() { return message; }
```
yihezkel, line 15: This method is unnecessary, as its parent class, Exception, already has this. Just call super(message) in the constructors below.

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataWebException.java
```diff
@@ -9,7 +9,7 @@
 public class DataWebException extends Exception{
 
     private String message;
```
yihezkel, line 11: This property is unnecessary, as its parent class, Exception, already has this. Just call super(message) in the constructors below.

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataWebException.java
```diff
@@ -9,7 +9,7 @@
...
-    private HttpResponse httpResponse;
+    private final HttpResponse httpResponse;
```
yihezkel, line 12: This field is never used, so I suggest we remove it altogether.

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/TriState.java
```diff
@@ -0,0 +1,12 @@
+package com.microsoft.azure.kusto.data.exceptions;
```
yihezkel, line 1: This class is quite generic. Do you intend to use this class elsewhere?

- On the one hand, this enum is only currently used for exception permanence. In that case, I think it would be clearer to name this enum "Permanence". It would also improve readability in new exception creation. Compare:
`throw new DataServiceException(clusterUrl, ERROR_ACQUIRING_APPLICATION_ACCESS_TOKEN, e, TriState.FALSE);`
vs
`throw new DataServiceException(clusterUrl, ERROR_ACQUIRING_APPLICATION_ACCESS_TOKEN, e, Permanence.FALSE);`
- On the other hand, if you envision using this in other cases (I personally don't), I suggest it should remain "TriState", but moved to a more generic package than "exceptions". I'd use "com.microsoft.azure.kusto.data.common" (which I think we should have anyway, for including anything that could be used by either the data or ingest projects).

- Lastly, an option I really like is to make use of existing Java functionality: Optional. Here's a sample usage:
`
public Optional<Boolean> isPermanent() {
    return Optional.of(true);
}
`

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -105,8 +106,11 @@ protected KustoResultSetTable(JSONObject jsonTable) throws KustoServiceError {
                             String message = exceptions.getString(0);
                             throw new KustoServiceError(message);
                         } else {
-                            throw new KustoServiceError(exceptions);
+                            throw new KustoServiceError(exceptions, false, EXCEPTIONS_MESSAGE);
```
yihezkel, line 109: I would change this entire block to:
`
if (exceptions != null) {
    throw new KustoServiceError(exceptions, false, EXCEPTIONS_MESSAGE);
}
`

I think it's the exception class' job to take the exceptions and set up the object accordingly, and this implementation convolutes the business logic. And of course this disregards that I don't even like the existence of the KustoServiceError class :-)
(If doing this, we'd also have to revert the EXCEPTIONS_MESSAGE to say it "failed were inner exception(s)")

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceError.java
```diff
@@ -9,21 +9,46 @@
...
+  This class represents an error that returned from the service
+ */
 public class KustoServiceError extends Exception {
-    private ArrayList<Exception> exceptions = null;
+    private ArrayList<Exception> exceptions;
 
-    public KustoServiceError(JSONArray exceptions) throws JSONException {
+    public KustoServiceError(JSONArray jsonExceptions, boolean isOneApi, String message) throws JSONException {
+        this(message);
         this.exceptions = new ArrayList<>();
```
yihezkel, line 20: This line is unnecessary, as the other constructor is called which already does this.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -141,7 +140,8 @@ static InputStream postToStreamingOutput(String url, String payload, long timeou
 
     private static DataServiceException createExceptionFromResponse(String url, HttpResponse httpResponse, Exception thrownException, String errorFromResponse) {
         if (httpResponse == null) {
-            return new DataServiceException(url, "POST failed to send request", thrownException);
+            return new DataServiceException(url, "POST failed to send request", thrownException,
+                    TriState.DONT_KNOW);
```
yihezkel, line 144: I'm not sure about this DONT_KNOW option. The way I see it, if we don't know that it's permanent, we generally assume transient. And over time we refine which errors are considered permanent/transient.

In this case, if `httpClient.execute(request)` returns null, I'd consider that to be transient. But more generally, I can't think of many times when DONT_KNOW should be used instead of FALSE.

In the .NET code, I see:
`
public static bool IsPermanent(this IngestionErrorCode ingestionErrorCode)
`
The property Kusto.Data.Net.Http.OneApiError.BaseError.permanent is a simple bool.

In Kusto.Cloud.Platform.Utils.ExtendedException.IsTransient, DONTKNOW is an option, but it's only returned if the exception doesn't match any of the if/elses.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConfidentialAppTokenProviderBase.java
```diff
@@ -29,14 +30,15 @@ protected IAuthenticationResult acquireNewAccessToken() throws DataServiceExcept
...
+            throw new DataServiceException(clusterUrl, ERROR_ACQUIRING_APPLICATION_ACCESS_TOKEN, e, TriState.FALSE);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
-            throw new DataServiceException(clusterUrl, ERROR_ACQUIRING_APPLICATION_ACCESS_TOKEN, e);
+            throw new DataServiceException(clusterUrl, ERROR_ACQUIRING_APPLICATION_ACCESS_TOKEN, e, TriState.FALSE);
         }
 
         if (result == null) {
-            throw new DataServiceException(clusterUrl, "acquireNewAccessToken got 'null' authentication result");
+            throw new DataServiceException(clusterUrl, "acquireNewAccessToken got 'null' authentication result",
```
yihezkel, line 40: I'd assume this isn't permanent

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoClientException.java
ohadbitt: sounds good

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataServiceException.java
ohadbitt: Its for users to use

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataServiceException.java
ohadbitt: it is used - by tests

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataWebException.java
ohadbitt: again its for the usage of customers

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/TriState.java
ohadbitt: will use Permanence
i started with an optional boolean but was asked to replace - i then tried to do something similar to the .Net but i guess we wont be using this elsewhere really

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
ohadbitt: so its the same as i answered in the other one - i wanted the message of the exception to be the exception if there is only one exception

yihezkel (approved):

PR #177, opened 2021-06-20
Add metadata endpoint

AsafMah:
#### Pull Request Description

_[Add a description of your pull request here]_

---

#### Future Release Comment
_[Add description of your change, to include in the next release]_
_[Delete any or all irrelevant sections, e.g. if your change does not warrant a release comment at all]_

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -9,6 +9,7 @@ import org.apache.http.*;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.conn.EofSensorInputStream;
 import org.apache.http.entity.ContentType;
@@ -33,12 +34,14 @@ import java.util.Arrays;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
 import java.util.zip.DeflaterInputStream;
 import java.util.zip.GZIPInputStream;
 
 class Utils {
     private static final int MAX_REDIRECT_COUNT = 1;
     private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+    private static final Long GET_TIMEOUT_IN_MILLISECS = TimeUnit.MINUTES.toMillis(4);
 
     private Utils() {
         // Hide constructor, as this is a static utility class
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -41,7 +41,6 @@ import java.util.zip.GZIPInputStream;
 class Utils {
     private static final int MAX_REDIRECT_COUNT = 1;
     private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-    private static final Long GET_TIMEOUT_IN_MILLISECS = TimeUnit.MINUTES.toMillis(4);
 
     private Utils() {
         // Hide constructor, as this is a static utility class
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -9,7 +9,6 @@ import org.apache.http.*;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
-import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.conn.EofSensorInputStream;
 import org.apache.http.entity.ContentType;
@@ -34,7 +33,6 @@ import java.util.Arrays;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
-import java.util.concurrent.TimeUnit;
 import java.util.zip.DeflaterInputStream;
 import java.util.zip.GZIPInputStream;
 
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
@@ -5,18 +5,29 @@ package com.microsoft.azure.kusto.data.auth;
 
 import com.microsoft.aad.msal4j.ConfidentialClientApplication;
 import com.microsoft.aad.msal4j.IClientCertificate;
+import com.microsoft.azure.kusto.data.exceptions.DataClientException;
+
 import org.jetbrains.annotations.NotNull;
 
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 
 public class ApplicationCertificateTokenProvider extends ConfidentialAppTokenProviderBase {
+    private final String applicationClientId;
+    private final IClientCertificate clientCertificate;
+
     ApplicationCertificateTokenProvider(@NotNull String applicationClientId, @NotNull IClientCertificate clientCertificate, @NotNull String clusterUrl, String authorityId) throws URISyntaxException {
         super(clusterUrl, authorityId);
+        this.applicationClientId = applicationClientId;
+        this.clientCertificate = clientCertificate;
+    }
+
+    @Override
+    protected void onCloudInit() throws DataClientException {
         try {
             clientApplication = ConfidentialClientApplication.builder(applicationClientId, clientCertificate).authority(aadAuthorityUrl).validateAuthority(false).build();
         } catch (MalformedURLException e) {
-            throw new URISyntaxException(aadAuthorityUrl, ERROR_INVALID_AUTHORITY_URL);
+            throw new DataClientException(clusterUrl, ERROR_INVALID_AUTHORITY_URL, e);
         }
     }
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
@@ -15,6 +15,7 @@ import java.io.IOException;
 import java.net.URISyntaxException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 
 public class CloudInfo {
     public static final String METADATA_ENDPOINT = "v1/rest/auth/metadata";
@@ -51,7 +52,7 @@ public class CloudInfo {
         this.firstPartyAuthorityUrl = firstPartyAuthorityUrl;
     }
 
-    public static void manuallyAddToCache(String clusterUrl, CloudInfo cloudInfo){
+    public static void manuallyAddToCache(String clusterUrl, CloudInfo cloudInfo) {
         synchronized (cache) {
             cache.put(clusterUrl, cloudInfo);
         }
@@ -67,7 +68,7 @@ public class CloudInfo {
             CloudInfo result;
 
             try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
-                HttpGet request = new HttpGet(UriUtils.concatPathToUri(clusterUrl, METADATA_ENDPOINT));
+                HttpGet request = new HttpGet(UriUtils.setPathForUri(clusterUrl, METADATA_ENDPOINT));
                 request.addHeader(HttpHeaders.ACCEPT_ENCODING, "gzip,deflate");
                 request.addHeader(HttpHeaders.ACCEPT, "application/json");
                 try (CloseableHttpResponse response = httpClient.execute(request)) {
@@ -108,6 +109,29 @@ public class CloudInfo {
                 innerObject.getString("KustoServiceResourceId"),
                 innerObject.getString("FirstPartyAuthorityUrl")
         );
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        CloudInfo cloudInfo = (CloudInfo) o;
+
+        return loginMfaRequired == cloudInfo.loginMfaRequired
+                && Objects.equals(loginEndpoint, cloudInfo.loginEndpoint)
+                && Objects.equals(kustoClientAppId, cloudInfo.kustoClientAppId)
+                && Objects.equals(kustoClientRedirectUri, cloudInfo.kustoClientRedirectUri)
+                && Objects.equals(kustoServiceResourceId, cloudInfo.kustoServiceResourceId)
+                && Objects.equals(firstPartyAuthorityUrl, cloudInfo.firstPartyAuthorityUrl);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(loginMfaRequired, loginEndpoint, kustoClientAppId, kustoClientRedirectUri, kustoServiceResourceId, firstPartyAuthorityUrl);
     }
 
     public boolean isLoginMfaRequired() {
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
@@ -1,17 +1,18 @@
 package com.microsoft.azure.kusto.data.auth;
 
-import com.microsoft.azure.kusto.data.exceptions.DataClientException;
 import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
 
 import org.apache.http.HttpHeaders;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.utils.URIBuilder;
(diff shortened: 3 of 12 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -33,12 +34,14 @@
...
 import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
 import java.util.zip.DeflaterInputStream;
 import java.util.zip.GZIPInputStream;
 
 class Utils {
     private static final int MAX_REDIRECT_COUNT = 1;
     private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+    private static final Long GET_TIMEOUT_IN_MILLISECS = TimeUnit.MINUTES.toMillis(4);
```
yogilad, line 44: I can't find the use for this constant.
What do we wait 4 minutes for?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -0,0 +1,132 @@
...
+    private static final Map<String, CloudInfo> cache = new HashMap<>();
+
+    private final boolean loginMfaRequired;
+    private final String loginEndpoint;
+    private final String kustoClientAppId;
+    private final String kustoClientRedirectUri;
+    private final String kustoServiceResourceId;
+    private final String firstPartyAuthorityUrl;
+
+    public CloudInfo(boolean loginMfaRequired, String loginEndpoint, String kustoClientAppId, String kustoClientRedirectUri, String kustoServiceResourceId, String firstPartyAuthorityUrl) {
```
yogilad, line 44: Any reason for this to be public?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -0,0 +1,132 @@
...
+        synchronized (cache) {
+            CloudInfo cloudInfo = cache.get(clusterUrl);
+            if (cloudInfo != null) {
+                return cloudInfo;
+            }
+
+            CloudInfo result;
+
+            try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
+                HttpGet request = new HttpGet(clusterUrl + "/" + METADATA_ENDPOINT);
```
yogilad, line 69: What if clusterUrl ends with a '/'

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -0,0 +1,132 @@
...
+                    } else {
+                        String errorFromResponse = EntityUtils.toString(response.getEntity());
+                        throw new DataServiceException(clusterUrl,
+                                "Error in metadata endpoint, got code: " + statusCode + "\nWith error: " + errorFromResponse, true);
+                    }
+                }
+            } catch (IOException ex) {
+                throw new DataServiceException(clusterUrl, "IOError when trying to retrieve CloudInfo", ex, true);
+            }
+            cache.put(clusterUrl, result);
```
yogilad, line 91: Does this require sync?

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
```diff
@@ -41,13 +48,38 @@ private String determineAadAuthorityUrl(String authorityId) {
...
     }
 
+    protected void initializeCloudInfo() throws DataServiceException, DataClientException {
+        if (cloudInfo != null) {
+            return;
+        }
+        synchronized (this) {
+            if (cloudInfo != null) {
+                return;
+            }
```
yogilad, line 66: This should enclose the entire below section

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
```diff
@@ -41,13 +48,38 @@ private String determineAadAuthorityUrl(String authorityId) {
...
+
+            cloudInfo = CloudInfo.retrieveCloudInfoForCluster(clusterUrl);
```
yogilad, line 68: This can lead to race conditions.
Set it last or use a different property to determine completion and set it last

yogilad (changes requested):
Preliminary review (haven't read it all)

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
AsafMah: Well it's mainly for the `manuallyAddToCache` function.
I needed to create it for the tests, and I thought there is no harm in making in public.
I can make both private/protected/package-private if you prefer

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
AsafMah: Thanks!

Technically you could say the same comments about when I concatenate to URIs I get from the metadata query,
but for them even c# relies on them not having a '/'. But maybe I should do it anyway for consistency? WDYT?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
AsafMah: Yes - 
1. HashMaps aren't really necessarily thread safe, so protecting `put` calls is a good idea
2. Since all of the function is synced, this means that only one thread will actually make the web reqeust

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
AsafMah: I don't understand? 
The rest of the function is within the sync, or did you mean the if? I don't understand

yogilad: Ignore this.
I blame the morning and not enough coffee.

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
AsafMah: I don't see how, the whole thing is synchronized.
Another thread will have to wait until all of the fields are initialized, and then will enter and immediately exit on `cloudInfo != null`. Did I miss something?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
yogilad: Tests are important, so keep it :)
Just wanted to make sure

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
yogilad: Aren't there any Java URL composition APIs which do this without having to manually parse the last char?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
yogilad: I read the scope wrong again.
Sorry

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
yogilad: first two lines are out of the sync scope
```
protected void initializeCloudInfo() throws DataServiceException, DataClientException {
        if (cloudInfo != null) {
            return;
        }
        synchronized (this) {
```

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
AsafMah: I really hate how github only displays the outdated diffs, you'll see in a recent commit that I changed it to 
`HttpGet request = new HttpGet(new URIBuilder(clusterUrl).setPath(METADATA_ENDPOINT).build().toString());`

Which is more verbose but is the "correct" way to do it.

The question is should I do the same for other URI concatinations we have, even if we know for sure they don't end with '/'?

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
AsafMah: That's on purpose, it's the Double-Checked locking pattern.
So if cloudInfo is initialized it saves us entering the synchronized block.
I also used it in python in TokenProviderBase._init_once and CloudSettings.get_cloud_info_for_cluster

Maybe it neesd a comment?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
ohadbitt: yea while you are on it

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
ohadbitt: its clear enough

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -0,0 +1,132 @@
...
+    public CloudInfo(boolean loginMfaRequired, String loginEndpoint, String kustoClientAppId, String kustoClientRedirectUri, String kustoServiceResourceId, String firstPartyAuthorityUrl) {
+        this.loginMfaRequired = loginMfaRequired;
+        this.loginEndpoint = loginEndpoint;
+        this.kustoClientAppId = kustoClientAppId;
+        this.kustoClientRedirectUri = kustoClientRedirectUri;
+        this.kustoServiceResourceId = kustoServiceResourceId;
+        this.firstPartyAuthorityUrl = firstPartyAuthorityUrl;
+    }
+
+    public static void manuallyAddToCache(String clusterUrl, CloudInfo cloudInfo){
```
ohadbitt, line 54: It seems like this is intended for users - but the class is not public

ohadbitt (approved):

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -0,0 +1,133 @@
...
+            } catch (IOException | URISyntaxException ex) {
+                throw new DataServiceException(clusterUrl, "IOError when trying to retrieve CloudInfo", ex, true);
+            }
+            cache.put(clusterUrl, result);
+            return result;
+        }
+    }
+
+    private static CloudInfo parseCloudInfo(String content) {
+        JSONObject jsonObject = new JSONObject(content);
```
ohadbitt, line 98: Ran this with local Engine and got result
{"dSTS":null,"AzureAD":null}
which fails on getJSONObject("AzureAD)
Might want to check this on other SDKs too (return here null and use default)

ohadbitt (approved):

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
yogilad: Check out the section named "The Visibility Issue"
https://mortoray.com/2012/02/28/double-checked-locking-why-doesnt-it-work-and-how-to-fix-it/

data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
```diff
@@ -5,18 +5,29 @@
...
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 
 public class ApplicationCertificateTokenProvider extends ConfidentialAppTokenProviderBase {
+    private final String applicationClientId;
+    private final IClientCertificate clientCertificate;
+
     ApplicationCertificateTokenProvider(@NotNull String applicationClientId, @NotNull IClientCertificate clientCertificate, @NotNull String clusterUrl, String authorityId) throws URISyntaxException {
         super(clusterUrl, authorityId);
+        this.applicationClientId = applicationClientId;
```
yogilad, line 21: Asaf, I'm not sure how this code ties up to the current PR, or if you are fixing auth issues on the way (where are those used?).

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
AsafMah: According to this: https://stackoverflow.com/questions/3578604/how-to-solve-the-double-checked-locking-is-broken-declaration-in-java
The simplest way to solve this is to declare the field as volatile, which I now did

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
AsafMah: LMK if the code looks good now

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
AsafMah: It's public now

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
AsafMah: Good catch, also fixed in python

data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
AsafMah: It used to initalize the `clientApplication` object in the ctor, but now we can't (because we don't want a web request in a ctor), so the logic moved to `onCloudInit`.

That meant that I need to save the params that are later passed to the clientApplication ctor.

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
yogilad: There are tons of problems with double check lock 😄 
The one I refer to in particular is this:
* You use the state of cloudInfo as an indicator to skip init
* Inside the initialization block you set cloud info first
* You follow to initialize other members
* Other threads may see that cloudInfo is not null and assume that scopes and aadAuthorityUrl are also initialized 
* Hence a race condition

AsafMah: Ohhhh!
Great catch

I replaced it now with a boolean `isCloudInfoInit`, that should prevent the race condition

ohadbitt: I think Assaf current implemenation i good although in the above stackoverflow it is discussed that the alternative to that is to use singleton which this thing really is  - then maybe we should do it like that 
If you ask me i don't think this is race is relevant here - we used to have the same methodology with AAD auth before Msal with no problems .

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
ohadbitt: fine with me

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
```diff
@@ -18,14 +18,10 @@
     protected static final String ERROR_ACQUIRING_APPLICATION_ACCESS_TOKEN = "Error acquiring ApplicationAccessToken";
     protected final Logger logger = LoggerFactory.getLogger(getClass());
     protected final String clusterUrl;
-    protected final Set<String> scopes;
 
     TokenProviderBase(@NotNull String clusterUrl) throws URISyntaxException {
         URI clusterUri = new URI(clusterUrl);
         this.clusterUrl = String.format("%s://%s", clusterUri.getScheme(), clusterUri.getHost());
```
ohadbitt, line 24: I think the right way for that is UriBuilder

ohadbitt (approved):

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -345,6 +346,20 @@ void testCreateWithAadTokenProviderAuthentication() {
         assertTrue(canAuthenticate(engineCsb));
     }
 
+    @Test
+    void testCloudInfoWithCluster() throws DataServiceException {
```
yogilad, line 350: Yep.
In theory the content of both default and retrieved settings is the same so you can compare it as well

data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
AsafMah: It took me until Yochai's comment to see this
The methodology was fine, the problem was with `cloudInfo` as the sentinel variable.
We initialize other things after cloudInfo, so a thread that will exit in the second line of the function with a non-null cloudInfo might not have working `scopes` or  `aadAuthorityUrl`

That's why I changed it to a boolean that only gets set at the end of the function.

yogilad (approved):

ohadbitt (approved):

PR #180, opened 2021-07-11
Remove resultSet

ohadbitt:
**Breaking Changes:**
- Remove ResultSet interface

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
@@ -566,7 +566,7 @@ public class KustoResultSetTable {
         return rows.size();
     }
 
-    public boolean isNullAt(int columnIndex){
+    public boolean isNull(int columnIndex){
         return get(columnIndex) == null;
     }
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
@@ -111,7 +111,7 @@ public class KustoResultSetTable {
                         }
                     } else {
                         throw new KustoServiceQueryError(((JSONObject) row).getJSONArray(
-                                "OneApiErrors"),true, EXCEPTIONS_MESSAGE);
+                                "OneApiErrors"), true, EXCEPTIONS_MESSAGE);
                     }
                 }
                 JSONArray rowAsJsonArray = jsonRows.getJSONArray(i);
@@ -235,7 +235,7 @@ public class KustoResultSetTable {
         return (Double) get(columnIndex);
     }
 
-   public byte[] getBytes(int columnIndex) {
+    public byte[] getBytes(int columnIndex) {
         return (byte[]) get(columnIndex);
     }
 
@@ -479,7 +479,6 @@ public class KustoResultSetTable {
     /**
      * This will cut the date up to yyyy-MM-dd'T'HH:mm:ss.SSS
      */
-
     public Date getDate(int columnIndex, Calendar calendar) throws SQLException {
         if (calendar == null) {
             return getDate(columnIndex);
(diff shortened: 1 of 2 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -139,7 +139,7 @@ protected KustoResultSetTable(JSONObject jsonTable) throws KustoServiceQueryErro
         return currentRow;
     }
 
-    @Override
+
```
AsafMah, line 142: I think the file needs re-formatting
Sometimes the `@Override` is replaced with a blank line, other times it doesn't
Make sure the space between methods stays consistent

AsafMah:
Since this is a proper breaking change we will need to bump the major, right?

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -506,42 +415,26 @@ public BigDecimal getBigDecimal(int columnIndex) {
...
     }
 
-    @Override
     public void beforeFirst() {
         rowIterator = rows.iterator();
     }
 
-    @Override
-    public void afterLast() {
-        while (next()) ;
```
yogilad, line 541: Why was this removed (not deprecated it not implemented)?

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -1018,314 +577,11 @@ public URL getURL(String columnName) throws SQLException {
...
-    @Override
-    public boolean isWrapperFor(Class<?> aClass) {
-        return false;
-    }
-
     public int count() {
         return rows.size();
     }
+
+    public boolean isNullAt(int columnIndex){
```
yogilad, line 584: IsColumnNull or simply IsNull

yogilad (approved):
Some small comments

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
ohadbitt: I dont think its relevant to this SDK usage - i think the only reason to use this is if you provide two ways iterator - so you can go afterLast and do while(previous)

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
ohadbitt: took this from Spark - i think its more clear

ohadbitt: will take isNull

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -255,27 +235,14 @@ public Double getDoubleObject(int columnIndex) {
...
-        if (get(columnIndex) == null) {
-            return null;
-        }
-
-        return (BigDecimal) get(columnIndex);
-    }
-
-    @Override
-    public byte[] getBytes(int columnIndex) {
+   public byte[] getBytes(int columnIndex) {
```
AsafMah, line 238: nit: missing a space in indentation

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -928,7 +479,7 @@ public LocalDateTime getKustoDateTime(String columnName) {
     /**
      * This will cut the date up to yyyy-MM-dd'T'HH:mm:ss.SSS
      */
-    @Override
+
```
AsafMah, line 482: nit: line

AsafMah (approved):

PR #181, opened 2021-08-04
Add x-ms-activitycontext and x-ms-user-id headers

yihezkel:
#### Pull Request Description

Add x-ms-activitycontext and x-ms-user-id headers to service requests to improve telemetry

---

#### Future Release Comment
**Features:**
- Add x-ms-activitycontext and x-ms-user-id headers to service requests to improve telemetry

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoDataExceptionBase.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoDataExceptionBase.java
@@ -1,10 +1,10 @@
 package com.microsoft.azure.kusto.data.exceptions;
 
-public abstract class KustoDataException extends Exception {
+public class KustoDataExceptionBase extends Exception {
     private final String ingestionSource;
     private final boolean isPermanent;
 
-    public KustoDataException(String ingestionSource, String message, Exception exception, boolean isPermanent) {
+    public KustoDataExceptionBase(String ingestionSource, String message, Exception exception, boolean isPermanent) {
         super(message, exception);
         this.ingestionSource = ingestionSource;
         this.isPermanent = isPermanent;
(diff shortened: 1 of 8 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoDataExceptionBase.java
```diff
@@ -1,10 +1,10 @@
 package com.microsoft.azure.kusto.data.exceptions;
 
-public abstract class KustoDataException extends Exception {
```
ohadbitt, line 3: why not abstract (the new name even suggests its abstract)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -132,13 +132,13 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
 
         } catch (DataClientException | IOException e) {
-            log.error(e.getMessage(), e);
-            throw new IngestionClientException(e.getMessage(), e);
-        } catch (DataServiceException e) {
             log.error(e.getMessage(), e);
             if (e.getCause() instanceof DataWebException && "Error in post request".equals(e.getMessage())) {
                 validateEndpointServiceType(connectionDataSource, EXPECTED_SERVICE_TYPE);
             }
+            throw new IngestionClientException(e.getMessage(), e);
```
ohadbitt, line 139: maybe in another PR
IngestionClientException should also have isPermanent and could actually just pass the ispermanent of its exception

ohadbitt (approved):

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoDataExceptionBase.java
yihezkel: I vaguely remember having a reason, but I'm not sure what I was thinking :-D

Reverting back to abstract.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
yihezkel: Agreed. As discussed, one of us will do this in another PR. I'm making a note of it.

PR #185, opened 2021-08-30
Fix time out as timestamp

ohadbitt:
Timeout is expected to be a timespan not long

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
@@ -3,13 +3,19 @@
 
 package com.microsoft.azure.kusto.data;
 
+import com.microsoft.azure.kusto.data.exceptions.TimespanParseException;
 import org.apache.commons.lang3.StringUtils;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.time.Duration;
 import java.time.LocalTime;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.microsoft.azure.kusto.data.Utils.SECONDS_PER_DAY;
 
 /*
  * Kusto supports attaching various properties to client requests (such as queries and control commands).
@@ -24,9 +30,11 @@ public class ClientRequestProperties {
     public static final String OPTION_SERVER_TIMEOUT = "servertimeout";
     public static final String OPTION_CLIENT_REQUEST_ID = "ClientRequestId";
     private static final long NANOS_TO_MILLIS = 1000000L;
-    private HashMap<String, Object> parameters;
-    private HashMap<String, Object> options;
-
+    private final HashMap<String, Object> parameters;
+    private final HashMap<String, Object> options;
+    private static final Pattern PATTERN =
+            Pattern.compile("(\\d+)?\\.?([0-2]?\\d:(?:60|[0-5]\\d):(?:60|[0-5]\\d)(:?\\.\\d+))",
+                    Pattern.CASE_INSENSITIVE);
     public ClientRequestProperties() {
         parameters = new HashMap<>();
         options = new HashMap<>();
@@ -70,7 +78,7 @@ public class ClientRequestProperties {
         if (timeoutObj instanceof Long) {
             timeout = (Long) timeoutObj;
         } else if (timeoutObj instanceof String) {
-            timeout = LocalTime.parse((String) timeoutObj).toNanoOfDay() / NANOS_TO_MILLIS;
+            timeout = getMillisFromTimespanString((String) timeoutObj);
         } else if (timeoutObj instanceof Integer) {
             timeout = Long.valueOf((Integer) timeoutObj);
         }
@@ -78,8 +86,25 @@ public class ClientRequestProperties {
         return timeout;
     }
 
+    private long getMillisFromTimespanString(String str) throws TimespanParseException {
+        Matcher matcher = PATTERN.matcher(str);
+        if (!matcher.matches()) {
+            throw new TimespanParseException(str);
+        }
+
+        long millis = 0;
+        String days = matcher.group(1);
+        if (days != null) {
+            millis += Long.parseLong(days) * SECONDS_PER_DAY * 1000L;
+        }
+
+        millis += LocalTime.parse(matcher.group(1)).toNanoOfDay() / 1000L;
+        return millis;
+    }
+
     public void setTimeoutInMilliSec(Long timeoutInMs) {
-        options.put(OPTION_SERVER_TIMEOUT, timeoutInMs);
+        Duration duration = Duration.ofMillis(timeoutInMs);
+        options.put(OPTION_SERVER_TIMEOUT, Utils.formatDurationAsTimespan(duration));
     }
 
     JSONObject toJson() {
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
@@ -98,7 +98,7 @@ public class ClientRequestProperties {
             millis += Long.parseLong(days) * SECONDS_PER_DAY * 1000L;
         }
 
-        millis += LocalTime.parse(matcher.group(1)).toNanoOfDay() / 1000L;
+        millis += LocalTime.parse(matcher.group(2)).toNanoOfDay() / 1000000L;
         return millis;
     }
 
@@ -110,10 +110,18 @@ public class ClientRequestProperties {
     JSONObject toJson() {
         try {
             JSONObject optionsAsJSON = new JSONObject(this.options);
-            Long timeoutInMilliSec = getTimeoutInMilliSec();
-            if (timeoutInMilliSec != null) {
-                LocalTime localTime = LocalTime.ofNanoOfDay(timeoutInMilliSec * NANOS_TO_MILLIS);
-                optionsAsJSON.put(OPTION_SERVER_TIMEOUT, localTime.toString());
+            Object timeoutObj = getOption(OPTION_SERVER_TIMEOUT);
+
+            if (timeoutObj != null) {
+                String timeoutString = "";
+                if (timeoutObj instanceof Long) {
+                    timeoutString = LocalTime.ofNanoOfDay((Long)timeoutObj * NANOS_TO_MILLIS).toString();
+                } else if (timeoutObj instanceof String) {
+                    timeoutString = (String) timeoutObj;
+                } else if (timeoutObj instanceof Integer) {
+                    timeoutString = LocalTime.ofNanoOfDay((Integer)timeoutObj * NANOS_TO_MILLIS).toString();
+                }
+                optionsAsJSON.put(OPTION_SERVER_TIMEOUT, timeoutString);
             }
             JSONObject json = new JSONObject();
             json.put(OPTIONS_KEY, optionsAsJSON);
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -27,9 +27,6 @@ import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
 import java.net.*;
 import java.time.Duration;
-import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeFormatterBuilder;
-import java.time.format.ResolverStyle;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Optional;
@@ -37,8 +34,7 @@ import java.util.Properties;
 import java.util.zip.DeflaterInputStream;
 import java.util.zip.GZIPInputStream;
 
-import static java.time.temporal.ChronoField.*;
-import static java.time.temporal.ChronoField.NANO_OF_SECOND;
+import static com.microsoft.azure.kusto.data.auth.CloudInfo.LOCALHOST;
 
 class Utils {
     private static final int MAX_REDIRECT_COUNT = 1;
@@ -50,7 +46,9 @@ class Utils {
 
     static String post(String url, String payload, InputStream stream, long timeoutMs, Map<String, String> headers, boolean leaveOpen) throws DataServiceException, DataClientException {
         URI uri = parseUriFromUrlString(url);
-        HttpClient httpClient = getHttpClient(Math.toIntExact(timeoutMs));
+        HttpClient httpClient = getHttpClient(timeoutMs > Integer.MAX_VALUE ?
+                Integer.MAX_VALUE :
+                Math.toIntExact(Integer.MAX_VALUE));
 
         try (InputStream ignored = (stream != null && !leaveOpen) ? stream : null) {
             HttpPost request = setupHttpPostRequest(uri, payload, stream, headers);
@@ -236,11 +234,12 @@ class Utils {
     private static URI parseUriFromUrlString(String url) throws DataClientException {
         try {
             URL cleanUrl = new URL(url);
-            if ("https".equalsIgnoreCase(cleanUrl.getProtocol())) {
+            if ("https".equalsIgnoreCase(cleanUrl.getProtocol()) || url.equalsIgnoreCase(LOCALHOST)) {
                 return new URI(cleanUrl.getProtocol(), cleanUrl.getUserInfo(), cleanUrl.getHost(), cleanUrl.getPort(), cleanUrl.getPath(), cleanUrl.getQuery(), cleanUrl.getRef());
             } else {
-                throw new DataClientException(url, "Cannot forward security token to a remote service over insecure " +
-                        "channel (http://)");
+//                throw new DataClientException(url, "Cannot forward security token to a remote service over insecure " +
+//                        "channel (http://)");
+                return new URI(cleanUrl.getProtocol(), cleanUrl.getUserInfo(), cleanUrl.getHost(), cleanUrl.getPort(), cleanUrl.getPath(), cleanUrl.getQuery(), cleanUrl.getRef());
             }
         } catch (URISyntaxException | MalformedURLException e) {
             throw new DataClientException(url, "Error parsing target URL in post request:" + e.getMessage(), e);
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -25,16 +25,20 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
-import java.net.MalformedURLException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URL;
+import java.net.*;
+import java.time.Duration;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.ResolverStyle;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.zip.DeflaterInputStream;
 import java.util.zip.GZIPInputStream;
+
+import static java.time.temporal.ChronoField.*;
+import static java.time.temporal.ChronoField.NANO_OF_SECOND;
 
 class Utils {
     private static final int MAX_REDIRECT_COUNT = 1;
@@ -46,7 +50,6 @@ class Utils {
 
     static String post(String url, String payload, InputStream stream, long timeoutMs, Map<String, String> headers, boolean leaveOpen) throws DataServiceException, DataClientException {
         URI uri = parseUriFromUrlString(url);
-
         HttpClient httpClient = getHttpClient(Math.toIntExact(timeoutMs));
 
         try (InputStream ignored = (stream != null && !leaveOpen) ? stream : null) {
@@ -65,7 +68,11 @@ class Utils {
                     throw createExceptionFromResponse(url, response, null, responseContent);
                 }
             }
-        } catch (JSONException | IOException e) {
+        }
+        catch (SocketTimeoutException e) {
+            throw new DataServiceException(url, "Timed out in post request:" + e.getMessage(), false);
+        }
+        catch (JSONException | IOException e) {
             throw new DataClientException(url, "Error in post request:" + e.getMessage(), e);
         }
         return null;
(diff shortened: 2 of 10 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -23,10 +29,11 @@
...
     public static final String OPTION_SERVER_TIMEOUT = "servertimeout";
     public static final String OPTION_CLIENT_REQUEST_ID = "ClientRequestId";
-    private static final long NANOS_TO_MILLIS = 1000000L;
-    private HashMap<String, Object> parameters;
-    private HashMap<String, Object> options;
-
+    private final HashMap<String, Object> parameters;
+    private final HashMap<String, Object> options;
+    private static final Pattern PATTERN =
+            Pattern.compile("(?:(\\d+)\\.)?((?:[0-2]?\\d:)?(?:60|[0-5]\\d):(?:60|[0-5]\\d)(?:\\.\\d+)?)",
```
AsafMah, line 35: This regex is a lot, if you took it from somewhere you should link it, and if you made it then it needs comments.
I think it's subtly wrong though:

1. 60 isn't a valid value for minutes/seconds (unless there's a leap second, maybe?)
2. Since the hour is optional, but isn't dependent on the days, values like "1.53:00" are valid even though they seem not to


But I think that we can do it without the regex - you already validate the string with Long.parseLong and 
LocalTime.parse, so you don't need to do the work twice.

I assume you did this because LocalTime.parse doesn't support days, so instead of the regex just split on the first ".".

There is an edge case of when they don't have days, but do have millis (13:24:30.54), but in this case just check if the days have a ":", and if so don't split.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -70,25 +77,51 @@ public Long getTimeoutInMilliSec() {
...
     }
 
     JSONObject toJson() {
         try {
             JSONObject optionsAsJSON = new JSONObject(this.options);
-            Long timeoutInMilliSec = getTimeoutInMilliSec();
-            if (timeoutInMilliSec != null) {
-                LocalTime localTime = LocalTime.ofNanoOfDay(timeoutInMilliSec * NANOS_TO_MILLIS);
-                optionsAsJSON.put(OPTION_SERVER_TIMEOUT, localTime.toString());
+            Object timeoutObj = getOption(OPTION_SERVER_TIMEOUT);
```
AsafMah, line 111: Technically not yours, but isn't it weird that we support all of these types in the OPTION_SERVER_TIMEOUT, but we only officially provide one way to set it - setTimeoutInMilliSec?

It's not documented anywhere as far as I know, so we should either add other methods or not support it

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
```diff
@@ -8,15 +8,15 @@
...
- * When a CSV data source schema and the target schema doesn't match or when using JSON, AVRO, PARQUET or ORC formats,
+ * When a CSV data source schema and the target schema doesn't match or when using JSON, AVRO formats,
  * there is a need to define an ingestion mapping to map the source schema to the table schema.
  * This class describes a pre-define ingestion mapping by its name- mapping reference and its kind.
  */
 public class IngestionMapping {
     private ColumnMapping[] columnMappings;
     private IngestionMappingKind ingestionMappingKind;
     private String ingestionMappingReference;
-    public static final List<String> mappingRequiredFormats = Arrays.asList("json", "singlejson", "avro", "apacheavro", "parquet", "orc");
```
AsafMah, line 19: Why did you change this?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -46,14 +47,17 @@ public ClientImpl(ConnectionStringBuilder csb) throws URISyntaxException {
...
+
         String auth = clusterUri.getAuthority().toLowerCase();
         if (host == null && auth.endsWith(FEDERATED_SECURITY_POSTFIX)) {
             url = new URIBuilder().setScheme(clusterUri.getScheme()).setHost(auth.substring(0, clusterUri.getAuthority().indexOf(FEDERATED_SECURITY_POSTFIX))).toString();
             csb.setClusterUrl(url);
         }
 
         clusterUrl = url;
-        aadAuthenticationHelper = TokenProviderFactory.createTokenProvider(csb);
+        aadAuthenticationHelper = clusterUrl.toLowerCase().startsWith(CloudInfo.LOCALHOST) ?
```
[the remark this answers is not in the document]
yihezkel: Also, if localhost can run as https, creating a URL object would help for comparing just the hostname.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -23,10 +29,11 @@
     private static final String PARAMETERS_KEY = "Parameters";
     public static final String OPTION_SERVER_TIMEOUT = "servertimeout";
     public static final String OPTION_CLIENT_REQUEST_ID = "ClientRequestId";
-    private static final long NANOS_TO_MILLIS = 1000000L;
-    private HashMap<String, Object> parameters;
-    private HashMap<String, Object> options;
-
+    private final HashMap<String, Object> parameters;
```
yihezkel, line 32: Map instead of HashMap will be more flexible for future changes, and more importantly, for tests.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: I would name this MILLIS_FROM_TIMESPAN_REGEX instead of PATTERN.

In terms of Asaf's comments:
- I think the validity of "60" depends on the specs of the producer (C#'s System.Timespan). It might be correct - I'm sure it's documented somewhere.
- I'm surprised the hour component *can* be optional! And as Asaf said, especially if the days are present.
- I suspect the above are incorrect, per some regexes that others built without allowing "60" or independently-optional hours: https://stackoverflow.com/questions/12438861/timespan-regex-with-days-portion
- I personally would keep the Regex, but I would refactor `getMillisFromTimespanString` to create built-in Java methods, like Temporal and Duration, from the regex' 2 (or you might need more) groups. Getting millis from these standard objects would be trivial.

I'm personally less concerned with accepting an invalid C# Timespan (like hour==29), than with being sure we properly convert all valid Strings accurately to millis.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -70,25 +77,51 @@ public Long getTimeoutInMilliSec() {
         if (timeoutObj instanceof Long) {
             timeout = (Long) timeoutObj;
         } else if (timeoutObj instanceof String) {
-            timeout = LocalTime.parse((String) timeoutObj).toNanoOfDay() / NANOS_TO_MILLIS;
+            timeout = getMillisFromTimespanString((String) timeoutObj);
```
yihezkel, line 80: I'd rename this.
"get" implies you're only interested in the millis *component* (and I was confused as it led me to think that, initially).
But we instead are (parsing and) *converting* the Timespan string to millis. Or, if we use Java's Duration class, we really just need to parse it, and then converting to millis is trivial using the class' method.

Perhaps:
`Duration convertTimespanStringToDuration(String timespanStr)` or `Duration parseTimespanStringAsDuration(String timespanStr)`

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/TimespanParseException.java
```diff
@@ -0,0 +1,8 @@
+package com.microsoft.azure.kusto.data.exceptions;
+
+public class TimespanParseException extends RuntimeException {
```
yihezkel, line 3: I personally don't see the benefit of creating a new class of exceptions for each exception scenario.
I see exception classes as being categories, and their `String message` as providing more specific details.

Java already has a https://docs.oracle.com/javase/8/docs/api/java/text/ParseException.html?is-external=true for the general case, and if you end up using Duration (or anything from the Java Time library), has https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeParseException.html

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -70,25 +77,51 @@ public Long getTimeoutInMilliSec() {
...
+    private long getMillisFromTimespanString(String str) throws TimespanParseException {
+        Matcher matcher = PATTERN.matcher(str);
+        if (!matcher.matches()) {
+            throw new TimespanParseException(str);
+        }
+
+        long millis = 0;
+        String days = matcher.group(1);
+        if (days != null) {
+            millis += Long.parseLong(days) * SECONDS_PER_DAY * 1000L;
```
yihezkel, line 97: As above, I think Java's Duration class is better suited instead of reinventing the wheel.

Java's time library also already has all these constants covered. For example, `SECONDS_PER_DAY` would be:
`long seconds = TimeUnit.DAYS.toSeconds(1)`
And the millis you're looking for would be:
`long millis = TimeUnit.DAYS.toMillis(1)`

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -70,25 +77,51 @@ public Long getTimeoutInMilliSec() {
...
+        }
+
+        millis += LocalTime.parse(matcher.group(2)).toNanoOfDay() / 1000000L;
```
yihezkel, line 100: If working with millis (a duration of time and not a point in time), I'd use Duration over LocalTime.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: The documentation I see only mentions allowing a C# Timespan:
https://docs.microsoft.com/en-us/azure/data-explorer/kusto/api/netfx/request-properties
https://docs.microsoft.com/en-us/azure/data-explorer/kusto/concepts/querylimits

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: Also, I believe the servertimeout property may not be more than an hour!

If that's the case, we can both validate the property's input, and simplify this regex.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -23,10 +29,11 @@
     private static final String PARAMETERS_KEY = "Parameters";
     public static final String OPTION_SERVER_TIMEOUT = "servertimeout";
     public static final String OPTION_CLIENT_REQUEST_ID = "ClientRequestId";
-    private static final long NANOS_TO_MILLIS = 1000000L;
-    private HashMap<String, Object> parameters;
-    private HashMap<String, Object> options;
-
+    private final HashMap<String, Object> parameters;
+    private final HashMap<String, Object> options;
```
yihezkel, line 33: 1) HashMap -> Map
2) This naming conflicts with a local variable of type `JSONObject` in the method `fromString`

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -70,25 +77,51 @@ public Long getTimeoutInMilliSec() {
...
             JSONObject optionsAsJSON = new JSONObject(this.options);
-            Long timeoutInMilliSec = getTimeoutInMilliSec();
-            if (timeoutInMilliSec != null) {
-                LocalTime localTime = LocalTime.ofNanoOfDay(timeoutInMilliSec * NANOS_TO_MILLIS);
-                optionsAsJSON.put(OPTION_SERVER_TIMEOUT, localTime.toString());
+            Object timeoutObj = getOption(OPTION_SERVER_TIMEOUT);
+
+            if (timeoutObj != null) {
+                String timeoutString = "";
+                if (timeoutObj instanceof Long) {
```
yihezkel, line 115: I don't think this would work. Wouldn't every value in a JSON be interpreted as a String?

I think all 3 of:
- {"Options":{"servertimeout":"2"}}
- {"Options":{"servertimeout":"9999999999"}}
- {"Options":{"servertimeout":"445635.29:60:60.354654"}}
Would resolve as `instanceof String`

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -46,14 +47,17 @@ public ClientImpl(ConnectionStringBuilder csb) throws URISyntaxException {
         String url = csb.getClusterUrl();
         URI clusterUri = new URI(url);
         String host = clusterUri.getHost();
+        Ensure.argIsNotNull(clusterUri.getAuthority(), "clusterUri.authority");
```
yihezkel, line 50: This is nice in C#, but Java already gives us `Objects.requireNonNull()` and Apache Commons already gives us `Validate.notNull()`

I'd get rid of this Ensure class.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -148,19 +155,19 @@ private static DataServiceException createExceptionFromResponse(String url, Http
...
                 try {
                     message = String.format("%s, ActivityId='%s'", formattedException.getApiError().getDescription(), activityId);
+                    return new DataServiceException(url, message, formattedException,
+                            formattedException.getApiError().isPermanent());
                 } catch (Exception ignored) {
                 }
-                return new DataServiceException(url, message, formattedException,
-                        formattedException.getApiError().isPermanent());
             }
+            errorFromResponse = String.format("Http StatusCode='%s', ActivityId='%s'", httpResponse.getStatusLine().toString(), activityId);
```
yihezkel, line 168: This may overwrite errorFromResponse's value.

I'd put this 1 line in an else.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -222,14 +229,15 @@ private static HttpPost setupHttpPostRequest(URI uri, String payload, InputStrea
...
+
         return request;
     }
 
     @NotNull
     private static URI parseUriFromUrlString(String url) throws DataClientException {
         try {
             URL cleanUrl = new URL(url);
-            if ("https".equalsIgnoreCase(cleanUrl.getProtocol())) {
+            if ("https".equalsIgnoreCase(cleanUrl.getProtocol()) || url.equalsIgnoreCase(CloudInfo.LOCALHOST)) {
```
yihezkel, line 240: cleanUrl.getHost().equalsIgnoreCase(CloudInfo.LOCALHOST)

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -251,4 +259,28 @@ static String getPackageVersion() {
         }
         return "";
     }
+    static final int SECONDS_PER_MINUTE = 60;
+    static final int MINUTES_PER_HOUR = 60;
+    static final int HOURS_PER_DAY = 24;
+    static final int SECONDS_PER_HOUR = SECONDS_PER_MINUTE * MINUTES_PER_HOUR;
+    static final int SECONDS_PER_DAY = HOURS_PER_DAY * SECONDS_PER_HOUR;
```
yihezkel, line 266: All of these constants are already defined in the Java time library. See comment above.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -251,4 +259,28 @@ static String getPackageVersion() {
...
+
+    public static String formatDurationAsTimespan(Duration duration) {
+        long seconds = duration.getSeconds();
+        int nanos = duration.getNano();
+
+        long hours = (seconds / SECONDS_PER_HOUR) % HOURS_PER_DAY;
```
yihezkel, line 272: I think I'm missing something here, but as far as I can tell, the logic should be:
`long nanoSecondsRemaining = duration.getNano() + TimeUnit.SECONDS.toNanoseconds(duration.getSeconds());

long days = nanoSecondsRemaining/TimeUnit.DAYS.toNanoseconds(1);
nanoSecondsRemaining -= days * TimeUnit.DAYS.toNanoseconds(1);

int hours = nanoSecondsRemaining/TimeUnit.HOURS.toNanoseconds(1);
nanoSecondsRemaining -= hours * TimeUnit.HOURS.toNanoseconds(1);

int minutes = nanoSecondsRemaining/TimeUnit.MINUTES.toNanoseconds(1);
nanoSecondsRemaining -= minutes * TimeUnit.MINUTES.toNanoseconds(1);

int seconds = nanoSecondsRemaining/TimeUnit.SECONDS.toNanoseconds(1);
nanoSecondsRemaining -= seconds * TimeUnit.SECOND.toNanoseconds(1);
`

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -251,4 +259,28 @@ static String getPackageVersion() {
         }
         return "";
     }
+    static final int SECONDS_PER_MINUTE = 60;
+    static final int MINUTES_PER_HOUR = 60;
+    static final int HOURS_PER_DAY = 24;
+    static final int SECONDS_PER_HOUR = SECONDS_PER_MINUTE * MINUTES_PER_HOUR;
+    static final int SECONDS_PER_DAY = HOURS_PER_DAY * SECONDS_PER_HOUR;
+
+    public static String formatDurationAsTimespan(Duration duration) {
```
yihezkel, line 268: Nevermind the intro of Duration above, I see you're already familiar with it.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/AzureStorageClient.java
```diff
@@ -3,6 +3,7 @@
 
 package com.microsoft.azure.kusto.ingest;
 
+import com.microsoft.azure.kusto.data.Ensure;
```
yihezkel, line 6: Why?

(Also, not a fan of the Ensure class to begin with, as above)

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
yihezkel: I see you didn't create it, and it's pretty entrenched in our project. I guess we can't get rid of it altogether, but I personally would reduce our dependency on it and use the built-in Java method here.

data/src/test/java/com/microsoft/azure/kusto/data/UtilitiesTest.java
```diff
@@ -0,0 +1,23 @@
...
+    @DisplayName("Convert millis to .Net timespan")
+    void convertMillisToTimespan() {
+        Long timeout = 40 * 60 * 1000L + 2000L; // 40 minutes 2 seconds
+        ClientRequestProperties clientRequestProperties = new ClientRequestProperties();
+        clientRequestProperties.setTimeoutInMilliSec(timeout);
+        Assertions.assertEquals(timeout, clientRequestProperties.getTimeoutInMilliSec());
+        Assertions.assertEquals(timeout, clientRequestProperties.getOption(ClientRequestProperties.OPTION_SERVER_TIMEOUT));
+        clientRequestProperties.setOption(ClientRequestProperties.OPTION_SERVER_TIMEOUT, "1.01:40:02.1");
+        clientRequestProperties.getTimeoutInMilliSec();
+        Assertions.assertEquals(25 * (60 * 60 * 1000L) + 40 * 60 * 1000L + 2000L + 100L,
```
yihezkel, line 18: Suggest extracting what's common between the inputs and the asserts to its own variable, and comparing that instead of literals.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
ohadbitt: duration.getNano returns just the remaining nanos as duration is made of (nanos ,seconds)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/AzureStorageClient.java
ohadbitt: No reason for it to be on ingest anyway - data is common to both

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionMapping.java
ohadbitt: ITs the truth

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: I wrote a test to check those other values, and as far as I could tell, in all cases it interpreted the object as a String.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
yihezkel: That's why the first line I wrote was:
`duration.getNano() + TimeUnit.SECONDS.toNanoseconds(duration.getSeconds())`

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: The optional indicator (?) is after:
- (?:(\\d+)\\.)?
- (?:[0-2]?\\d:)?
- (?:\\.\\d+)?
Aren't the first 2 the days and hours?

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: I think it's relatively straightforward to do so.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: But why are we providing them so many ways to set the value? Shouldn't we require they stick to the documented approach?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
yihezkel: Sorry, I don't follow.

We can wrap just this 1 line in an else, and the following line - the return, will be called no matter what.

Also, note that the catch on line 165 above does nothing and the execution passes to line 168.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
ohadbitt: I didnt get your point - you can test the regex in some GUI

yihezkel: The RegEx is: `(?:(\\d+)\\.)?((?:[0-2]?\\d:)?(?:60|[0-5]\\d):(?:60|[0-5]\\d)(?:\\.\\d+)?)`
Its parts seem to be:
- (?:(\\d+)\\.)? == Days, which are optional
- ((?:[0-2]?\\d:)? == Hours, which are optional
- (?:60|[0-5]\\d): == Minutes
- (?:60|[0-5]\\d) == Seconds
- (?:\\.\\d+)?) == < Seconds, which are optional

Am I wrong?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
ohadbitt: i tested mysef

PR #191, opened 2021-10-25
Reactivate tests and fix non passed tests

ohadbitt:
Reactivate tests, with new E2E and fix non passed tests
** bug fix **
* extension was duplicated csv.gz.csv.gz
** breaking change **
* ingestion properties must define format


data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -56,9 +56,9 @@ public CloudInfo(boolean loginMfaRequired, String loginEndpoint, String kustoCli
         this.firstPartyAuthorityUrl = firstPartyAuthorityUrl;
     }
 
-    public static void manuallyAddToCache(String clusterUrl, CloudInfo cloudInfo) {
+    public static void manuallyAddToCache(String clusterUrl, CloudInfo cloudInfo) throws URISyntaxException {
         synchronized (cache) {
-            cache.put(clusterUrl, cloudInfo);
+            cache.put(UriUtils.setPathForUri(clusterUrl, ""), cloudInfo);
```
yihezkel, line 61: nit: I'd think (new Uri(clusterUri)).ToString() would've been cleaner.

ingest/pom.xml
```diff
@@ -203,6 +193,18 @@
...
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>${junit.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
```
yihezkel, line 204: Is this "api" package really necessary if the full package is already included (immediately above)?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -290,6 +290,17 @@ void validate() {
         }
     }
 
+    public void validateResultSetProperties() {
+        String givenDataFormat = this.getDataFormat();
+        if (givenDataFormat == null) {
+            this.setDataFormat(IngestionProperties.DataFormat.csv);
```
yihezkel, line 296: I assume this is documented that CSV will be set by default if the user doesn't set it

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -290,6 +290,17 @@ void validate() {
...
+        } else {
+            Ensure.isTrue(givenDataFormat.equals(IngestionProperties.DataFormat.csv.name()),
```
yihezkel, line 298: I must be missing something here - why does this seem to require that the user-provided DataFormat is CSV?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -290,6 +290,17 @@ void validate() {
...
+                    String.format("ResultSet translates into csv format but {%s} was given", givenDataFormat));
```
yihezkel, line 299: nit: We usually use single quotes around variables in traces

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -130,12 +131,12 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
 
         } catch (DataClientException | IOException e) {
             log.error(e.getMessage(), e);
-            if (e.getCause() instanceof DataWebException && "Error in post request".equals(e.getMessage())) {
-                validateEndpointServiceType(connectionDataSource, EXPECTED_SERVICE_TYPE);
-            }
             throw new IngestionClientException(e.getMessage(), e);
         } catch (DataServiceException e) {
             log.error(e.getMessage(), e);
+            if (e.getMessage() != null && e.getMessage().contains("404")) {
```
yihezkel, line 137: Better to match more than just 404. Maybe use a regex to ensure \s\d\d\d\s or similar.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
AsafMah: Is it the same functionality? This removes everything after the domain

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
AsafMah: We know it is now, right?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
AsafMah: It's explained in the next line - ResultSet translates into a csv stream, so for them specifically it has to be csv

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
AsafMah: Can we relay on more that?

ingest/pom.xml
AsafMah: We do need it, otherwise it complains about "implicit dependency"

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: Now it's in the Javadoc (or once the newest PR is merged)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
yihezkel: More would be ideal - it requires testing to see the range of possibilities, But I suspect a text match of "404" is insufficient.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
yihezkel: Sorry, I didn't mean to include the `ToString()`.

I meant to create a `URI` object, which provides all subcomponents for free instead of having to reinvent the wheel.

ingest/pom.xml
yihezkel: I vaguely remember already fixing this. In master, we don't have both dependencies - we have just the api.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: I see the comment, but is that accurate? I'm still feeling like I'm missing something obvious.

yihezkel: OK I think I understand - `ResultSet`s have to be of format `CSV`. This isn't documented in the SDK, though. The further we get from "toshb'a"/the business logic being documentation the better.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
AsafMah: We need to strip the path out of the urls and make sure they're in the same format.
The only way to change the path is with the URI builder which is what we do in the method.

Do you mean change cache so that it would map URIs rather than strings? we could do that, but we still need UriUtils.setPathForUri

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
AsafMah: Ok, I fixed it properly.
We now transfer the actual response object, and then we can see if it's actually 404 LMK what you think

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
AsafMah: I added a line of doc, WDYT?

ingest/pom.xml
AsafMah: Ok I removed it here to.
The order is different because here it's alphabetical

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
yihezkel: Not changing what's cached, but instead how we get just the protocol+hostname.
It's a "nit" anyway, but I just see this as harder to read than using the `URL` object as the easy and standard way to get its components:
`
URL urlToClean = (new URL(clusterUrl));
cache.put(urlToClean.getProtocol() + "://" + urlToClean.getHost()) // not sure if port is relevant
`
I like `URIBuilder` as well: `new URIBuilder(uri).setPath("").build().toString()` - which is what's effectively done here when `UriUtils.setPathForUri()` is called.

I guess I'm just saying that I think it's less clear to me that `UriUtils.setPathForUri(clusterUrl, "")` simplifies the URL to just the protocol+hostname, compared to the other 2 approaches above.

AsafMah: Maybe another method -  UriUtils.stripUriPath(clusterUrl) that will call UriUtils.setPathForUri(clusterUrl, "") ?

yihezkel: I'd prefer to directly use the code that use built-in (or Apache) Java methods from my previous comment (`urlToClean.getHost()` or `new URIBuilder(uri).setPath("")`), instead of calling a static method. I don't even think that static method should exist. I just think this adds unnecessary complexity.

But this has lived longer than an average nit comment's lifespan, so nothing further from me :-)

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataServiceException.java
```diff
@@ -11,4 +11,8 @@ public DataServiceException(String ingestionSource, String message, boolean isPe
     public DataServiceException(String ingestionSource, String message, Exception exception, boolean isPermanent) {
         super(ingestionSource, message, exception, isPermanent);
     }
+
+    public boolean is404Error() {
+        return getCause() != null && getCause() instanceof DataWebException && ((DataWebException) getCause()).getHttpResponse().getStatusLine().getStatusCode() == 404;
```
yihezkel, line 16: Can it be a 404 error and not be of type DataWebException?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: Looks great!

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataServiceException.java
AsafMah: It can not. Look at `createExceptionFromResponse`
DataWebException is our type, and we create it whenever we get an http response which is not 200

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -398,6 +382,7 @@ void testCloudInfoWithCluster() throws DataServiceException {
         String clusterUrl = System.getenv("ENGINE_CONNECTION_STRING");
         CloudInfo cloudInfo = CloudInfo.retrieveCloudInfoForCluster(clusterUrl);
         assertNotSame(CloudInfo.DEFAULT_CLOUD, cloudInfo);
+        assertEquals(CloudInfo.DEFAULT_CLOUD, cloudInfo);
```
yihezkel, line 385: This depends on where it's run.

yihezkel (approved):

PR #194, opened 2021-11-08
Support AAD Managed Identity authentication

yihezkel:
#### Pull Request Description

Support AAD Managed Identity authentication

---

#### Future Release Comment
**Features:**
- Support AAD Managed Identity authentication


data/src/main/java/com/microsoft/azure/kusto/data/auth/ManagedIdentityTokenProvider.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public ManagedIdentityTokenProvider(String managedIdentityClientId, @NotNull String clusterUrl) throws URISyntaxException {
+        super(clusterUrl);
+        if (StringUtils.isNotBlank(managedIdentityClientId)) {
+            this.managedIdentityCredential = new ManagedIdentityCredentialBuilder()
+                    .clientId(managedIdentityClientId) // only required for user assigned
+                    .build();
+        } else {
+            this.managedIdentityCredential = new ManagedIdentityCredentialBuilder().build();
+        }
```
AsafMah, line 25:
```suggestion
       ManagedIdentityCredentialBuilder builder = new ManagedIdentityCredentialBuilder();
        if (StringUtils.isNotBlank(managedIdentityClientId)) {
            builder = builder.clientId(managedIdentityClientId); // only required for user assigned
        }
        this.managedIdentityCredential = builder.build();
```
Cutting down on a bit of duplication

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -238,38 +255,53 @@ public static ConnectionStringBuilder createWithAadApplicationCertificateSubject
...
         }
 
-        ConnectionStringBuilder csb = new ConnectionStringBuilder(resourceUri);
+        ConnectionStringBuilder csb = new ConnectionStringBuilder(clusterUrl);
         csb.tokenProvider = tokenProviderCallable;
         return csb;
     }
-}
+
+    public static ConnectionStringBuilder createWithAadManagedIdentity(String clusterUrl) {
```
AsafMah, line 293: This re-order is confusing.
Is there any actual change to this class other than adding this method and the two members?

pom.xml
```diff
@@ -46,6 +46,9 @@
         <json.version>20201115</json.version>
         <fasterxml.jackson.core.version>[2.11.0,2.12.0)</fasterxml.jackson.core.version>
         <azure-storage.version>8.6.6</azure-storage.version>
+        <azure-identity.version>1.4.0</azure-identity.version>
+        <azure-core.version>1.21.0</azure-core.version>
+        <reactor-core.version>3.4.10</reactor-core.version>
```
AsafMah, line 51: Why do we need this dependency?

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
yihezkel: Yeah, sorry. Need to make reviewers suffer once when we organize it, and then it should make it easier to read in the future.

The changes are:
1) These 2 methods that I added
2) Introducing an approach to CSB that makes more sense - an explicit "use" boolean to determine which flow is desired, instead of the TokenFactory trying to guess based on which properties are set. This actually became necessary for the first time now, and I'll likely want to expand the concept in the future.

pom.xml
yihezkel: For the `block()`

AsafMah: How does that work? you don't import it in the file, and java doesn't have extension methods

data/pom.xml
```diff
@@ -174,6 +174,21 @@
             <artifactId>slf4j-api</artifactId>
             <version>${slf4j.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.azure</groupId>
+            <artifactId>azure-identity</artifactId>
+            <version>${azure-identity.version}</version>
```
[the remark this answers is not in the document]
yihezkel: I resolved all dependency conflicts except `jna`, which is caused entirely by `azure-identity`. There's already an issue open for it: https://github.com/Azure/azure-sdk-for-java/issues/25298
It shouldn't affect us as we don't have any other dependencies on it, and when we upgrade, the (non-)issue should go away.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ManagedIdentityTokenProvider.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.NotNull;
+
+import java.net.URISyntaxException;
+
+public class ManagedIdentityTokenProvider extends TokenProviderBase {
+    private final ManagedIdentityCredential managedIdentityCredential;
+    private TokenRequestContext tokenRequestContext;
+
+    public ManagedIdentityTokenProvider(String managedIdentityClientId, @NotNull String clusterUrl) throws URISyntaxException {
```
AsafMah, line 17: Usually clusterUrl is the first param

AsafMah (changes requested):
Overall it looks very good.
Two things:
1. We need tests. Even if it's like in python's msi that we don't run by default automatically, we do need tests for the provider.
2. Resolve the build pipeline issue

data/src/main/java/com/microsoft/azure/kusto/data/auth/ManagedIdentityTokenProvider.java
yihezkel: I don't think that's the case.

yogilad: That's commonly the case in other SDKs.
Not sure of Java

yihezkel:
@AsafMah 
I just saw your comment. Adding a new one, as I don't seem to be able to reply to that.

I mentioned over e-mail but maybe you weren't on the thread, but the pipeline is fixed.

The only way to test MI auth is in envs that have the MI, which we can only use on an Azure machine set up for that. I ran a manual test in such an env, but I don't see the rationale in adding as a `@Test` something that'll never run as a `@Test`.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
```diff
@@ -15,12 +15,42 @@
...
     protected final Logger logger = LoggerFactory.getLogger(getClass());
     protected final String clusterUrl;
+    protected CloudInfo cloudInfo = null;
 
     TokenProviderBase(@NotNull String clusterUrl) throws URISyntaxException {
         this.clusterUrl = UriUtils.setPathForUri(clusterUrl, "");
     }
 
+    protected void initializeCloudInfo() throws DataServiceException {
+        if (cloudInfo == null) {
```
yogilad, line 28: != ?

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
```diff
@@ -15,12 +15,42 @@
...
+            return;
+        }
+        synchronized (this) {
+            if (cloudInfo == null) {
```
yogilad, line 32: != ?

data/src/main/java/com/microsoft/azure/kusto/data/auth/ManagedIdentityTokenProvider.java
yihezkel: Not in Java

yogilad: Saw below and was horrified.
Let's make this one conform

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
yihezkel: Good catch. When I finally got around to running the E2E tests, this bug caused them **all** to fail :-D

data/src/main/java/com/microsoft/azure/kusto/data/auth/ManagedIdentityTokenProvider.java
yihezkel: What was horrifying? That clusterUrl isn't the 1st param?

yogilad: That it moves from place to place depending on your choice of auth method

yogilad (approved):
Assuming tests run.
please do consider placing the URI as the first argument.
(In a thick English Accent) BY GOD WE ARE NOT BARBARIENS!

data/src/main/java/com/microsoft/azure/kusto/data/auth/ManagedIdentityTokenProvider.java
yihezkel: I believe the authorityId was always at the end if applicable, and clusterUrl was always before that.

AsafMah (approved):

PR #195, opened 2021-11-14
Quickstart Application

yihezkel:
#### Pull Request Description

The quick start application is a **self-contained and runnable** example script that demonstrates authenticating, connecting to, administering, ingesting data into and querying Azure Data Explorer using the azure-kusto-python SDK.
You can use it as a baseline to write your own first kusto client application, altering the code as you go, or copy code sections out of it into your app.

**Tip:** The app includes comments with tips on recommendations, coding best practices, links to reference materials and recommended TODO changes when adapting the code to your needs.

---

#### Future Release Comment
**Features:**
- Quickstart app
- Improve conversion between DataFormat and IngestionMappingKind


yihezkel:
Note that a prerequisite for this PR is https://github.com/Azure/azure-kusto-java/pull/194

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,399 @@
...
+        ExecutionType(String name) {
+            this.name = name;
+        }
+
+        public String getName() {
+            return name;
+        }
+    }
+
+    public static void main(String[] args) throws InterruptedException {
```
yihezkel, line 70: In a future commit/PR, I will break `main` into submethods.
This both aligns with best practices, but also makes it easier to refactor this "query-ingest-query" approach into an additional "query-only" mode.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,399 @@
...
+        loadConfigs();
+
+        if (AUTHENTICATION_MODE.equals("UserPrompt")) {
+            waitForUserToProceed("You will be prompted *twice* for credentials during this script. Please return to the console after authenticating.");
+        }
+
+        ConnectionStringBuilder kustoCsb = generateConnectionString(kustoUrl, AUTHENTICATION_MODE);
+        ConnectionStringBuilder ingestCsb = generateConnectionString(ingestUrl, AUTHENTICATION_MODE);
+        Client kustoClient = null;
+        try (IngestClient ingestClient = IngestClientFactory.createClient(ingestCsb)) {
```
yihezkel, line 82: I can also move the ingest client to right before the while loop for ingesting.

The benefit to moving it is to make it clear where it's needed/used.
The benefit to keeping it here is to keep it simple, creating both clients in one place.

I prefer the former, but suspect Yochai prefers the latter.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,399 @@
...
+        }
+    }
+
+    private static ConnectionStringBuilder generateConnectionString(String endpointUrl, String authenticationMode) {
+        ConnectionStringBuilder csb = null;
+        switch (authenticationMode) {
+            case "UserPrompt":
+                csb = ConnectionStringBuilder.createWithUserPrompt(endpointUrl);
+                break;
+            case "ManagedIdentity":
```
yihezkel, line 251: This feature is new, and is not yet merged (see https://github.com/Azure/azure-kusto-java/pull/194)

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,399 @@
...
+    private static ClientRequestProperties createClientRequestProperties(String scope) {
+        return createClientRequestProperties(scope, null);
+    }
+
+    private static ClientRequestProperties createClientRequestProperties(String scope, String timeout) {
+        ClientRequestProperties clientRequestProperties = new ClientRequestProperties();
+        clientRequestProperties.setClientRequestId(String.format("%s;%s", scope, UUID.randomUUID()));
+
+        // Tip: While not common, you can alter the request default command timeout using the below command, e.g. to set the timeout to 10 minutes, use "10m"
+        if (StringUtils.isNotBlank(timeout)) {
```
yihezkel, line 333: Why is `timeout` is a 4th distinct type of config, in addition to config file (`kusto-sample-config.json`), environment variables, and static constants meant to be configured at the top of this class?

I suspect the answer is that we just want to show **how** `timeout` would be used, but it wouldn't likely be used, and this approach to configs is what @yogilad considers best aligned with that. But I have to ask in case it's a mistake.

README.md
```diff
@@ -9,9 +9,10 @@ dev: [![Build status](https://msazure.visualstudio.com/One/_apis/build/status/Cu
 
 This is the Microsoft Azure Kusto client library which allows communication with Kusto to bring data in (ingest) and query information already stored in the database.
 This library contains 3 different modules:
-- data: the main client that allows interaction with Kusto. It's able to create a connection, issue (control) commands and query data.
-- ingest: this provides an easy way to bring data into Kusto
-- samples 
+- data: The main client that allows interaction with Kusto, including creating a connection, issuing (control) commands and querying data.
+- ingest: Provides an easy way to bring data into Kusto
+- quickstart: Self-contained, configurable, runnable sample app, for easily getting started wth the SDK.
```
yogilad, line 14:
```suggestion
- quickstart: Self-contained, configurable and runnable sample app for easily getting started with the SDK.
```

README.md
```diff
@@ -9,9 +9,10 @@ dev: [![Build status](https://msazure.visualstudio.com/One/_apis/build/status/Cu
 
 This is the Microsoft Azure Kusto client library which allows communication with Kusto to bring data in (ingest) and query information already stored in the database.
 This library contains 3 different modules:
-- data: the main client that allows interaction with Kusto. It's able to create a connection, issue (control) commands and query data.
-- ingest: this provides an easy way to bring data into Kusto
-- samples 
+- data: The main client that allows interaction with Kusto, including creating a connection, issuing (control) commands and querying data.
```
yogilad, line 12: nit: Be consistent about use of period at the end of a bullet

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -291,22 +291,38 @@ void validate() {
...
-        json,
-        singlejson,
-        multijson,
-        avro,
-        apacheavro,
-        parquet,
-        orc,
-        raw,
-        w3clogfile
+        csv(IngestionMapping.IngestionMappingKind.Csv, false),
```
yogilad, line 294: Isn't this a breaking change?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/SecurityUtils.java
```diff
@@ -14,7 +14,7 @@
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 
-class SecurityUtils {
+public class SecurityUtils {
     private SecurityUtils() {
         // Hide constructor as this is a Utils class
```
yogilad, line 19: Comment is no longer relevant

quickstart/ONECLICK-INSTRUCTION-BOX.md
```diff
@@ -0,0 +1,11 @@
+### Instructions
```
yogilad, line 1: This file's name should be lowercase only, as OneClick is case sensitive on the matter

quickstart/ONECLICK-INSTRUCTION-BOX.md
```diff
@@ -0,0 +1,11 @@
+### Instructions
+1. Download the **DOWNLOAD_LINK** as a zip file
+2. Unpack the script to your folder of choice
```
yogilad: this is not a script, but an app I assume

quickstart/README.md
```diff
@@ -0,0 +1,36 @@
...
+2. Modify `kusto-sample-config.json`, changing `KustoUri`, `IngestUri` and `DatabaseName` appropriately for your ADX cluster
+
+### Retrieving the app from OneClick
+1. Open a browser and type your cluster's URL (e.g. https://myadxcluster.westeurope.kusto.windows.net/), and you will be redirected to the _Azure Data Explorer_ website
+2. Open the left-side pane via the hamburger menu, if it isn't already open
+3. On the left-side pane, choose _Data_
+4. Click on _Generate Sample App Code_ button
+5. Follow the wizard
+6. Download the app as a zip file
+7. Unpack the script to your folder of choice
```
yogilad:
```suggestion
7. Unpack the source code to your folder of choice

```

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,395 @@
...
+public class KustoSampleApp {
+    // TODO - Config:
+    //  If this quickstart app was downloaded from OneClick, kusto-sample-config.json should be pre-populated with your cluster's details
+    //  If this quickstart app was downloaded from GitHub, edit kusto-sample-config.json and modify the cluster URL and database fields appropriately
+    private static final String CONFIG_FILE_NAME = "kusto-sample-config.json";
+
+    // TODO - Config (Optional): Change the authentication method from User Prompt to any of the other options
+    //  Some of the auth modes require additional environment variables to be set in order to work (see usage in generateConnectionString below)
+    //  Managed Identity Authentication only works when running as an Azure service (webapp, function, etc.)
+    private static final String AUTHENTICATION_MODE = "UserPrompt"; // Options: (UserPrompt|ManagedIdentity|AppKey|AppCertificate)
```
yogilad, line 39: DeviceCode?

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yogilad: I don't mind actually, though shorten the exception catch to 
```
try{ create client}
catch{exit} 
everything else
```

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,395 @@
...
+        if (AUTHENTICATION_MODE.equals("UserPrompt")) {
+            waitForUserToProceed("You will be prompted *twice* for credentials during this script. Please return to the console after authenticating.");
+        }
+
+        ConnectionStringBuilder kustoCsb = generateConnectionString(kustoUrl, AUTHENTICATION_MODE);
+        ConnectionStringBuilder ingestCsb = generateConnectionString(ingestUrl, AUTHENTICATION_MODE);
+        Client kustoClient = null;
+        try (IngestClient ingestClient = IngestClientFactory.createClient(ingestCsb)) {
+            // Tip: Avoid creating a new Kusto Client for each use. Instead, create the clients once and reuse them.
+            kustoClient = new ClientImpl(kustoCsb);
```
yogilad, line 84: How come ingestClient is tried and caught and queryClient is not?

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,395 @@
...
+                 2) More than 1GB of data was queued for ingestion for the same table by the same user
+                 3) More than 5 minutes have passed since the first file was queued for ingestion for the same table by the same user
+                For more information about customizing the ingestion batching policy, see: https://docs.microsoft.com/azure/data-explorer/kusto/management/batchingpolicy
+
+                Tip 1: Though most users should be fine with the defaults, to speed up ingestion, such as during development and
+                 in this sample app, we opt to modify the default ingestion policy to ingest data after 10 seconds have passed.
+                Tip 2: This is generally a one-time configuration.
+                Tip 3: You can also skip the batching for some files using the Flush-Immediately property, though this option should be used with care as it is inefficient.
+             */
+            waitForUserToProceed(String.format("Alter the batching policy for '%s.%s'", databaseName, tableName));
```
yogilad, line 110: In retrospect I think this should also fall under the !useExsitingTable condition

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,395 @@
...
+            waitForUserToProceed(String.format("Get existing row count in '%s.%s'", databaseName, tableName));
+            execute(kustoClient, databaseName, String.format("%s | count", tableName), ExecutionType.QUERY);
+
+            for (Map<String, String> file : dataToIngest) {
+                IngestionProperties.DataFormat dataFormat = IngestionProperties.DataFormat.valueOf(file.get("format").toLowerCase());
+                boolean useExistingMapping = Boolean.parseBoolean(file.get("useExistingMapping"));
+                String mappingName = file.get("mappingName");
+
+                // Learn More: For more information about providing inline mappings and mapping references, see: https://docs.microsoft.com/azure/data-explorer/kusto/management/mappings
+                if (!useExistingMapping) {
```
yogilad, line 126: This code section makes me think we need OneClick to explicitly provide the mapping mode (None, Use exiting Mapping, Create Mapping, Use Mapping Value)

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,395 @@
...
+            PrivateKey privateKey = SecurityUtils.getPrivateKey(publicCertFilePath);
+            X509Certificate x509Certificate = SecurityUtils.getPublicCertificate(privateKeyFilePath);
+            return ConnectionStringBuilder.createWithAadApplicationCertificate(endpointUrl, appId, x509Certificate, privateKey, tenantId);
+        } catch (IOException | GeneralSecurityException e) {
+            die("Couldn't create ConnectionStringBuilder for app certificate authentication", e);
+            return null;
+        }
+    }
+
+    private static boolean execute(Client kustoClient, String databaseName, String query, ExecutionType executionType) {
```
yogilad, line 291: I don't like this merge of execute and control command.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yogilad: Strictly for demonstrating how to set it

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    private static void ingestDataFromFile(IngestClient ingestClient, String databaseName, String tableName, String uri, IngestionProperties.DataFormat dataFormat, String mappingName) {
+        IngestionProperties ingestionProperties = createIngestionProperties(databaseName, tableName, dataFormat, mappingName);
+
+        // Tip 1: For optimal ingestion batching and performance, specify the uncompressed data size in the file descriptor.
+        //  Otherwise, the service will determine the file size, but it will require an additional s2s call, and may not be accurate for compressed files.
+        // Tip 2: To correlate between ingestion operations in your applications and Kusto, set the source id and log it somewhere.
+
+        File fileToIngest = new File(".\\" + uri);
```
yogilad, line 344: While this does serve to get the file size, I would not one anyone who copied this example code to open a file for the sake of getting the size from the path.
Perhaps the API should change to accept the size explicitly and this can be done externally

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,395 @@
...
+        //  See the SDK's kusto-samples module and the E2E tests in kusto-ingest for additional references.
+    }
+
+    private static void ingestDataFromBlob(IngestClient ingestClient, String databaseName, String tableName, String blobUrl, IngestionProperties.DataFormat dataFormat, String mappingName) {
+        IngestionProperties ingestionProperties = createIngestionProperties(databaseName, tableName, dataFormat, mappingName);
+
+        // Tip 1: For optimal ingestion batching and performance, specify the uncompressed data size in the file descriptor instead of the default below of 0.
+        //  Otherwise, the service will determine the file size, but it will require an additional s2s call, and may not be accurate for compressed files.
+        // Tip 2: To correlate between ingestion operations in your applications and Kusto, set the source id and log it somewhere
+        BlobSourceInfo blobSourceInfo = new BlobSourceInfo(blobUrl, 0, UUID.randomUUID());
```
yogilad, line 366: Follow the above comment, perhaps the size should come from the outside (as 0 or default = 0)

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,395 @@
...
+                Thread.sleep(TimeUnit.SECONDS.toMillis(1));
+            } catch (InterruptedException e) {
+                System.out.println("Exception while sleeping waiting for ingestion, with stack trace:");
+                e.printStackTrace();
+                throw e;
+            }
+        }
+        System.out.println();
+
+        System.out.printf("Step %s: Get post-ingestion row count for '%s.%s'%n", step++, databaseName, tableName);
```
yogilad, line 188: there were like a million steps that occurred before this ++ :)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: Any existing uses of this enum should work identically. This is only an addition, enabling `getIngestionMappingKind` and `isMappingRequired`.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/SecurityUtils.java
yihezkel: Why?

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: That's on my plate, but I didn't want it to hold up getting this PR out.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: The problem with shortening it is that the **use** of the ingest client needs to be inside the `try`, which is why the catch is so far down.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: Only the `IngestClient` implements `Closable`

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: I personally don't think this logic is complex, especially once we separate it to its own method (recall that `main` will be much simpler). Your call.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: Each of them is incremented via `waitForUserToProceed()`

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: Why not?
Not only is all the code in this method identical, but also in the Java SDK there is only 1 `execute` command, used identically for both queries and control commands.
Also, I thought you liked this idea last time we spoke about it.

I think duplicating code makes it harder for new users to understand the code.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: My intent was to demonstrate both approaches.
I'll consider ways to address your request in the previous comment.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yogilad: IngestClient = null
Try { 
ingestClient = CreateClient())
}
Catch 
{
error 
}

Do something with the client 

Or do I have java figured out backwards?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yogilad: This feels like a breaking change.
Would I be able to switch case on it???
Would I be able to enumerate it?

Why not add a couple of X2Y APIs which take the enum and return the value instead.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/SecurityUtils.java
yogilad: Because you no longer hide it - it's public

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yogilad: It kind of feels odd.
Commands need to return success (printing is decorative)
Query needs to return results (printing is substitute for handling it)

yogilad: This makes it less reusable - user wise. 
He'd need to split the function to do something

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: That's how I did it originally, but the problem with that is it doesn't take advantage of Java's "try-with-resources". It's similar to a `finally`, which ensures that `Closable` resources are properly closed no matter what, and is therefore generally best practices.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: Yes, all existing uses will work identically.
If by enumerate you're referring to using a hardcoded sequential number representing the order of the item in the enum, in Java that's never assumed unless explicitly instantiated as a property of each enum value. So nobody should do that, like assuming order in a data structure whose order isn't guaranteed.

This approach is OO - the object that supplies the name of the object also supplies the other things it should know about (its `IngestionMappingKind` and `isMappingRequired`).

ingest/src/main/java/com/microsoft/azure/kusto/ingest/SecurityUtils.java
yihezkel: The class is public, but it's still a Utils class, so we don't want them to create an instance of the class. So the constructor should be private.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: I don't see why they'd have to split it to use it. In the Java SDK, it's combined. The user sends all commands, whether control commands or queries, to this one method. To me, it seems simpler.

The underlying method for both returns the result table (`KustoOperationResult`) if successful, where the user can see details (details if control command which probably isn't interesting in most cases, and query result if query). To me, this consistency is intuitive. Splitting into distinct cases to me sounds unintuitive.

But it seems you see it the opposite way, so I'll split it.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: I want to be sure I understand the approach you prefer:
- I can add a fileSize parameter to this `ingestDataFromFile()` method, and set its value when calling the method via `fileToIngest.length()`
- Same, but simply use a value of 0 instead of calling `fileToIngest.length()`
- Same as the 1st idea, but get its value from the json so that they neither see us passing 0 (which is never the right thing to do), nor calling `fileToIngest.length()` which you fear would cause them to make bad performance decisions.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: Without also requiring that `alterTable` is true?

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,461 @@
...
+
+    private static void alterBatchingPolicy(Client kustoClient, String databaseName, String tableName) {
+        /*
+            Tip 1: Though most users should be fine with the defaults, to speed up ingestion, such as during development and
+             in this sample app, we opt to modify the default ingestion policy to ingest data after 10 seconds have passed.
+            Tip 2: This is generally a one-time configuration.
+            Tip 3: You can also skip the batching for some files using the Flush-Immediately property, though this option should be used with care as it is inefficient.
+         */
+        waitForUserToProceed(String.format("Alter the batching policy for '%s.%s'", databaseName, tableName));
+        String command = String.format(".alter table %s policy ingestionbatching @'%s'", tableName, BATCHING_POLICY);
```
yihezkel, line 330: Wholesale overwrite using .alter, or don't drop if not specified using .alter-merge?

quickstart/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Using the App for the first time
+
+### Prerequisites
+1. Set up Java on your machine. For instructions on how to install Java, consult a Java environment setup tutorial, like [this one](https://www.tutorialspoint.com/java/java_environment_setup.htm).
+2. Set up [Apache Maven](https://maven.apache.org/install.html), which is the most popular Java dependency management tool.
+
+### Retrieving the app from GitHub
+1. Download the app files from this GitHub repo
+2. Modify `kusto-sample-config.json`, changing `KustoUri`, `IngestUri` and `DatabaseName` appropriately for your ADX cluster
```
yihezkel, line 17: I had named the instruction `ONECLICK_INSTRUCTION_BOX.md` in line with the md file convention, but you asked me to rename for consistency with Python.

So the question here is similar - the filename `kusto-sample-config.json` uses hyphens instead of underscores, per Java's convention. However, the file in Python is slightly different in line with Python's conventions - `kusto_sample_config.json`.

Change this to be identical to the Python file's name?

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,461 @@
...
+
+                /*
+                    Learn More:
+                    Kusto batches data for ingestion efficiency. The default batching policy ingests data when one of the following conditions are met:
+                     1) More than 1,000 files were queued for ingestion for the same table by the same user
+                     2) More than 1GB of data was queued for ingestion for the same table by the same user
+                     3) More than 5 minutes have passed since the first file was queued for ingestion for the same table by the same user
+                    For more information about customizing the ingestion batching policy, see: https://docs.microsoft.com/azure/data-explorer/kusto/management/batchingpolicy
+                 */
+                alterBatchingPolicy(kustoClient, databaseName, tableName);
```
yihezkel, line 95: I can also require that `alterTable` is true.

Pro is that we don't want to make any unexpected changes, so an explicit flag would help prevent that.
Con is that this flag is for the table, not batching policy. To resolve this, we can also change the flag to be `alterExistingSchemas` (or something else), which I think would imply permission for both.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,461 @@
...
+            if (useExistingTable) {
+                if (alterTable) {
+                    // Tip: Usually table was originally created with a schema appropriate for the data being ingested, so this wouldn't be needed.
+                    // Learn More: For more information about altering table schemas, see: https://docs.microsoft.com/en-us/azure/data-explorer/kusto/management/alter-table-command
+                    alterExistingTableToProvidedSchema(kustoClient, databaseName, tableName, tableSchema);
+                }
+
+                if (queryData) {
+                    // Learn More: For more information about Kusto Query Language (KQL), see: https://docs.microsoft.com/azure/data-explorer/write-queries
+                    queryExistingNumberOfRows(kustoClient, databaseName, tableName);
```
yihezkel, line 80: I don't think it makes sense to query existing rows if it's a new table.

On the other hand, a user may want "Query-only" mode. In that case, though, I suggest that between this and the 2nd set of queries, better to run the 2nd (since it paints a more complete picture with 2 queries instead of 1).

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: Wrap in `if (false)`
Add comment why commented out - didn't want to mess with your batching policy

Also move this method to within `createNewTable`

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,461 @@
...
+        waitForUserToProceed(String.format("Create table '%s.%s' if it doesn't exist", databaseName, tableName));
+        String command = String.format(".create table %s %s", tableName, tableSchema);
+        if (!executeControlCommand(kustoClient, databaseName, command)) {
+            die("Failed to create table or validate it exists");
+        }
+    }
+
+    private static void alterExistingTableToProvidedSchema(Client kustoClient, String databaseName, String tableName, String tableSchema) {
+        waitForUserToProceed(String.format("Alter existing table '%s.%s' to align with the provided schema", databaseName, tableName));
+        String command = String.format(".alter table %s %s", tableName, tableSchema);
```
yihezkel, line 234: Ask Tzvia

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,461 @@
...
+            die("Failed to alter table");
+        }
+    }
+
+    private static boolean executeControlCommand(Client kustoClient, String databaseName, String controlCommand) {
+        ClientRequestProperties clientRequestProperties = createClientRequestProperties("SampleApp_ControlCommand");
+        KustoOperationResult result;
+        try {
+            result = kustoClient.execute(databaseName, controlCommand, clientRequestProperties);
+            System.out.printf("Response from executed control command '%s':%n", controlCommand);
```
yihezkel, line 245: Add comment that printing is only to help demonstrate behind the scenes, and you likely wouldn't

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: Just use 0

Just pass the URL, without a `File` object.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: See other comment - comment it out, add comment, and move within `createNewTable()`

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: Tzvia requested that this be `.alter-merge`

quickstart/README.md
yihezkel: Tal requested that the filenames be identical.

quickstart/oneclick_instruction_box.md
```diff
@@ -0,0 +1,14 @@
+### Prerequisites
+1. Set up Java on your machine. For instructions, consult a Java environment setup tutorial, like [this one](https://www.tutorialspoint.com/java/java_environment_setup.htm).
```
AsafMah, line 2: Maybe we should recommend openjdk over oracles?

quickstart/oneclick_instruction_box.md
```diff
@@ -0,0 +1,14 @@
+### Prerequisites
+1. Set up Java on your machine. For instructions, consult a Java environment setup tutorial, like [this one](https://www.tutorialspoint.com/java/java_environment_setup.htm).
+2. Set up [Apache Maven](https://maven.apache.org/install.html), which is the most popular Java dependency management tool.
```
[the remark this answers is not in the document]
yihezkel: It never came with maven

quickstart/oneclick_instruction_box.md
```diff
@@ -0,0 +1,14 @@
+### Prerequisites
+1. Set up Java on your machine. For instructions, consult a Java environment setup tutorial, like [this one](https://www.tutorialspoint.com/java/java_environment_setup.htm).
+2. Set up [Apache Maven](https://maven.apache.org/install.html), which is the most popular Java dependency management tool.
+
+### Instructions
+1. Download the **DOWNLOAD_LINK** as a zip file
```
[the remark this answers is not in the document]
yihezkel: Yes, OneClick fills this in

quickstart/oneclick_instruction_box.md
```diff
@@ -0,0 +1,14 @@
+### Prerequisites
+1. Set up Java on your machine. For instructions, consult a Java environment setup tutorial, like [this one](https://www.tutorialspoint.com/java/java_environment_setup.htm).
+2. Set up [Apache Maven](https://maven.apache.org/install.html), which is the most popular Java dependency management tool.
+
+### Instructions
+1. Download the **DOWNLOAD_LINK** as a zip file
+2. Extract the app source code to your folder of choice
+3. Open a command line window to the folder extracted to above
+4. Run `mvn clean install` to compile the source code into a binary
```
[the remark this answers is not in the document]
yihezkel: I think it's easier to install it so it'll go to the .m2 directory and can be sued on the computer from other contexts.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,465 @@
...
+        }
+    }
+
+    private static boolean executeControlCommand(Client kustoClient, String databaseName, String controlCommand) {
+        ClientRequestProperties clientRequestProperties = createClientRequestProperties("SampleApp_ControlCommand");
+        KustoOperationResult result;
+        try {
+            result = kustoClient.execute(databaseName, controlCommand, clientRequestProperties);
+            // Tip: Generally wouldn't print the response from a control command. We print here to demonstrate what the response looks like.
+            System.out.printf("Response from executed control command '%s':%n", controlCommand);
```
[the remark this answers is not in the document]
yihezkel: That's how I wrote it, but Yochai preferred they be separate so users can see more easily their different purposes.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,465 @@
...
+            if (useExistingTable) {
+                if (alterTable) {
+                    // Tip: Usually table was originally created with a schema appropriate for the data being ingested, so this wouldn't be needed.
+                    // Learn More: For more information about altering table schemas, see: https://docs.microsoft.com/en-us/azure/data-explorer/kusto/management/alter-table-command
+                    alterMergeExistingTableToProvidedSchema(kustoClient, databaseName, tableName, tableSchema);
+                }
+
+                if (queryData) {
+                    // Learn More: For more information about Kusto Query Language (KQL), see: https://docs.microsoft.com/azure/data-explorer/write-queries
+                    queryExistingNumberOfRows(kustoClient, databaseName, tableName);
```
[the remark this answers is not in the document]
yihezkel: Both to teach, and to show the change in the number of rows in the tables post-ingestion.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,465 @@
...
+            }
+
+            if (ingestData) {
+                for (Map<String, String> file : dataToIngest) {
+                    IngestionProperties.DataFormat dataFormat = IngestionProperties.DataFormat.valueOf(file.get("format").toLowerCase());
+                    String mappingName = file.get("mappingName");
+
+                    // Tip: This is generally a one-time configuration.
+                    // Learn More: For more information about providing inline mappings and mapping references, see: https://docs.microsoft.com/azure/data-explorer/kusto/management/mappings
+                    if (!createIngestionMappings(Boolean.parseBoolean(file.get("useExistingMapping")), kustoClient, databaseName, mappingName, file.get("mappingValue"), dataFormat)) {
```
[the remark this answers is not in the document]
yihezkel: I don't think we want to explain too many details here, as that would clutter the Quickstart. I think the intent is for them to use the linked docs for details like that.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,465 @@
...
+    }
+
+    private static void ingestData(Map<String, String> file, IngestionProperties.DataFormat dataFormat, IngestClient ingestClient, String databaseName, String tableName, String mappingName) {
+        String sourceType = file.get("sourceType");
+        String uri = file.get("dataSourceUri");
+        waitForUserToProceed(String.format("Ingest '%s' from %s", uri, sourceType));
+        // Tip: When ingesting json files, if each line represents a single-line json, use MULTIJSON format even if the file only contains one line.
+        //  If the json contains whitespace formatting, use SINGLEJSON. In this case, only one data row json object is allowed per file.
+        if (dataFormat == IngestionProperties.DataFormat.json) {
+            dataFormat = IngestionProperties.DataFormat.multijson;
```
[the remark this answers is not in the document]
yihezkel: That's a Quickstart-specific design decision, so that they're less likely to fail on this common issue.
The comment is there to clarify why it's being done.

quickstart/oneclick_instruction_box.md
yihezkel: Interesting idea. I do often prefer open source, especially considering that we're **in** an open source project. But I don't think there's any **practical** difference between them, and my impression is that officially Oracle is considered the reference version of Java.

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -0,0 +1,465 @@
...
+                } else {
+                    csb = ConnectionStringBuilder.createWithAadManagedIdentity(endpointUrl, clientId);
+                }
+                break;
+            case "AppKey":
+                // TODO - Config (Optional): App Id & tenant, and App Key to authenticate with
+                // For information about how to procure an AAD Application see: https://docs.microsoft.com/azure/data-explorer/provision-azure-ad-app
+                csb = ConnectionStringBuilder.createWithAadApplicationCredentials(endpointUrl, System.getenv("APP_ID"), System.getenv("APP_KEY"), System.getenv("APP_TENANT"));
+                break;
+            case "AppCertificate":
```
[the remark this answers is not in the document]
yihezkel: I'll add the comment about configuring the parameters, but I don't think we want to repeat the comment about creating an AAD app.

quickstart/oneclick_instruction_box.md
AsafMah: The site is very outdated (it recommends notepad as an editor but not intellij).
And Oracle requires you to create an account to download JDK 8, and I think it installs bloatware by default.

Check out https://adoptium.net/?variant=openjdk8

quickstart/oneclick_instruction_box.md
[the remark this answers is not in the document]
AsafMah: Mmmm, I guess I'm just used to it from Intellij

quickstart/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+## Using the App for the first time
+
+### Prerequisites
+1. Set up Java on your machine. For instructions, consult a Java environment setup tutorial, like [this one](https://www.tutorialspoint.com/java/java_environment_setup.htm).
+2. Set up [Apache Maven](https://maven.apache.org/install.html), which is the most popular Java dependency management tool.
+
+### Retrieving the app from GitHub
+1. Download the app files from this GitHub repo
```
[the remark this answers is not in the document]
AsafMah: If they don't know how to use git they probably won't know how to use github to download the zip.

You can either explain it or add a link (but it's for the master branch and not the latest release) https://github.com/Azure/azure-kusto-java/archive/refs/heads/master.zip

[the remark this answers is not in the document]
yihezkel: We need to draw the line somewhere, and I think downloading a zip is well before that line:
image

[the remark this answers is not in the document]
AsafMah: If someone isn't already familiar with github they will not know how to find this button

quickstart/oneclick_instruction_box.md
yihezkel: What bloatware does the Oracle JDK install? That's the version I use, and I don't remember it including any such additional software.

Do you think it's relevant that Oracle's JDK is the official reference version?
I think Oracle also supports its JDKs for much longer than OpenJDK, so my impression is that it's **preferred in the enterprise**.
Oracle's version also includes improvements that can't be included in OpenJDK because they aren't open source.

I don't think we're in the business of trying to find them the most up-to-date tutorial for setting up their local Java env. It's a bit of a bonus that we even provide an example of one.

yihezkel: I got lost in this discussion, so forgot to check whether we're even **recommending** which JDK to use. We're not, as far as I remember and see in the README.

I think this is a valuable discussion to decide which JDK to recommend if we were recommending one. But since we don't, and conceptually which JDK they use is very much out of scope of this Quickstart, I don't think this discussion should hold up releasing this Qucikstart which the PMs are under a tight schedule to release.

AsafMah: Fair enough, I'll resolve

AsafMah (approved):

Issue #186, opened 2021-09-07
Share HTTP client across requests


margelatu:
Hi, folks, we're using a few Azure Data Explorer clusters in a multi-region deployment of a service. The service exposes a REST API, allowing its users to query data which is found in the ADX cluster -- so we're talking about a long-lived application here.

The service uses the `kusto-data` library to execute the queries against the ADX cluster, specifically one of the flavours of `com.microsoft.azure.kusto.data.Client.execute()` see https://github.com/Azure/azure-kusto-java/blob/master/data/src/main/java/com/microsoft/azure/kusto/data/Client.java#L12 

What we noticed is that the implementation of the `Client`:
  * creates a new HTTP client object for each request, along with everything it entails such as an associated connection pool, a pool manager, etc.
  * does _not_ reuse HTTP connections or clients;
  * does _not_ close the HTTP client or the connection it uses in a request.

This can be observed here: https://github.com/Azure/azure-kusto-java/blob/cde236165432d48786c86812a0a0739e902b4f8a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java#L47

This is problematic from multiple aspects:
  * from a performance perspective, each query the service makes to ADX incurs the price of creating the new client and its associated resources, creating a new connection, SSL handshake, etc.
  * from a resource perspective, this is a waste of server resources.

What the library should probably do is create and reuse an HTTP client across multiple queries / requests. The client could be created during some sort of initialisation phase and should be closed during a "close" phase.

Any thoughts or feedback about this would be greatly appreciated.

yihezkel:
Thanks Ionut! We agree, and called this out as well when implementing a change related to the HttpClient in April:
https://github.com/Azure/azure-kusto-java/blob/master/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java#L199

We'll address this as soon as priorities permit.

margelatu:
@yihezkel Thanks for the response. Do you have a rough idea on when this could be tackled ?

yihezkel:
Hi @margelatu ,

Would you like to make a contribution to this? If not, we'll take care of it.

larisaad:
Hi @yihezkel, yes, we're preparing a PR for this.

margelatu:
@yihezkel the PR is now open thanks to @larisaad , see https://github.com/Azure/azure-kusto-java/pull/196

PR #196, opened 2021-11-16
Share HTTP client across requests

larisaad:
#### Pull Request Description

* Reuse the HTTP client an its associated connection pool across requests avoiding the overhead of creating new connections.
* Allow for a customized HTTP client: configurable connection pool size, max idle time, keep alive strategy through the new public class `HttpClientProperties`.

---

#### Future Release Comment
Kusto Data Client factory methods to allow for customizations of the underlying HTTP client (such as configurable connection pool size, max idle time, keep alive strategy) through a `HttpClientProperties` object.

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- https://github.com/Azure/azure-kusto-java/issues/186


larisaad:
Running some performance tests against `kusto-data` without and with this fix (~20req/s) we got:
* Improved p99.9 response time,

| kusto-data 2.8.2 | with HTTP fix |
| --- | --- |
| kusto-old2 | kusto-20conn2 |


* Significant improvements for garbage collection times & frequency, heap size, TCP events.

| kusto-data 2.8.2 | with HTTP fix |
| --- | --- |
| kusto-gc-counts-old | kusto-gc-counts-20conn |
| kusto-heap-old | kusto-heap-20conn |
| kusto-tcp-old | kusto-tcp-20conn |

margelatu:
@yihezkel Could you have a look at this, please ? Thank you !

data/src/main/java/com/microsoft/azure/kusto/data/HttpClientFactory.java
```diff
@@ -0,0 +1,116 @@
...
+        final HttpClientProperties properties = Optional.ofNullable(providedProperties)
+                .orElse(HttpClientProperties.builder().build());
+        final HttpClientBuilder httpClientBuilder = HttpClientBuilder.create()
+                .useSystemProperties()
+                .setMaxConnTotal(properties.maxConnectionTotal())
+                .setMaxConnPerRoute(properties.maxConnectionRoute())
+                .evictExpiredConnections()
+                .evictIdleConnections(properties.maxIdleTime(), TimeUnit.SECONDS);
+
+        if (properties.isKeepAlive()) {
```
AsafMah, line 59: This throws if properties == null. Try
```suggestion
        if (properties.isKeepAlive() != null && properties.isKeepAlive()) {
```

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -118,18 +116,18 @@ static InputStream postToStreamingOutput(String url, String payload, long timeou
                 }
                 // Though the server responds with a gzip/deflate Content-Encoding header, we reach here because httpclient uses LazyDecompressingStream which handles the above logic
                 returnInputStream = true;
+                httpResponse.close();
```
AsafMah, line 119: You shouldn't close this always.
In streaming query the connection stays open for the user.
Run the test testPerformanceKustoOperationResultVsJsonVsStreamingQuery to see.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -44,15 +42,17 @@ private Utils() {
         // Hide constructor, as this is a static utility class
     }
 
-    static String post(String url, String payload, InputStream stream, long timeoutMs, Map<String, String> headers, boolean leaveOpen) throws DataServiceException, DataClientException {
+    static String post(String url, String payload, InputStream stream, long timeoutMs, Map<String, String> headers,
```
AsafMah, line 45: It's weird that the httpclient is the value before the last.
I personally think it should be first.

AsafMah (commented):
Some small comments for now

larisaad:
@AsafMah @yihezkel Could you have a look on this one? Also, could I get access to the CI logs?

data/src/main/java/com/microsoft/azure/kusto/data/HttpClientFactory.java
```diff
@@ -0,0 +1,116 @@
...
+import java.io.IOException;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A singleton factory of HTTP clients.
+ */
+class HttpClientFactory {
+
+    private static final Logger logger = LoggerFactory.getLogger(HttpClientFactory.class);
```
yihezkel, line 24: nit: consts should be all upper case

data/src/main/java/com/microsoft/azure/kusto/data/HttpClientProperties.java
```diff
@@ -0,0 +1,173 @@
...
+     * server instructions set in the {@code Keep-Alive} response header; if the response doesn't contain a
+     * {@code Keep-Alive} header, the client will use a default keep-alive period indicated by
+     * {@linkplain #maxKeepAliveTime()}.
+     *
+     * @return whether or not a custom connection keep-alive strategy should be used
+     *
+     * @see #maxKeepAliveTime()
+     * @see <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive">Keep-Alive</a>
+     */
+    public Boolean isKeepAlive() {
```
yihezkel, line 55: Wouldn't this be simpler as a `boolean`?

data/src/main/java/com/microsoft/azure/kusto/data/HttpClientProperties.java
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * The maximum number of connections the client may keep open at the same time per route.
+     *
+     * @return the maximum number of connections per route
+     */
+    public Integer maxConnectionRoute() {
+        return maxConnectionRoute;
+    }
+
+    public static class Builder {
```
yihezkel, line 88: `HttpClientPropertiesBuilder`

samples/src/main/java/Query.java
```diff
@@ -19,7 +20,15 @@ public static void main(String[] args) {
...
-            ClientImpl client = new ClientImpl(csb);
+
+            HttpClientProperties properties = HttpClientProperties.builder()
+                    .keepAlive(true)
+                    .maxKeepAliveTime(120)
+                    .maxConnectionsPerRoute(40)
+                    .maxConnectionsTotal(40)
+                    .build();
+
+            ClientImpl client = new ClientImpl(csb, properties);
```
yihezkel, line 31: This is great to include a sample!

I suggest also adding 1 E2E test, and of course UTs.

data/src/main/java/com/microsoft/azure/kusto/data/HttpClientFactory.java
larisaad: We used Google code style, but i renamed the logger with upper case to stay consistent with the rest of the code.

samples/src/main/java/Query.java
larisaad: Looking at the code we added, the factory and the properties class, I don't see what UT we can add for that. Regarding the E2E, there isn't new functionality added in this PR and I think testing should be covered by existing E2Es.

data/src/main/java/com/microsoft/azure/kusto/data/ClientFactory.java
```diff
@@ -12,11 +12,53 @@ private ClientFactory() {
         // Hide the default constructor, as this is a factory with static methods
```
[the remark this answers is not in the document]
larisaad: It looks like 500 is returned when refreshing an auth token. I don't know how can it be related to this PR. Could you guys help me out debug this?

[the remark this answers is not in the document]
yihezkel: Sure, I'll take a look!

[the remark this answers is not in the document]
yihezkel: When I run all tests via your branch locally, everything passes. So there are no blocking issues on that front.

There must be an issue with the CI settings, which we had recently changed. I'll take a look.

samples/src/main/java/Query.java
yihezkel: I think UTs would be unnecessary for the properties class, but I think it would be important to unit test `create()` in `HttpClientFactory`.

I would recommend an E2E test that injects the `CloseableHttpClient` (you can make that property `protected` in `ClientImpl`, executes 2 commands (e.g. queries), and spy to verify that the same object was used for both calls.

data/src/main/java/com/microsoft/azure/kusto/data/ClientFactory.java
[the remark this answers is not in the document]
yihezkel: I confirmed that our CI passes with the current code in this branch. For some reason it isn't updating the snapshot below, so we'll just have to force push it once it has the UTs + E2E test.

samples/src/main/java/Query.java
larisaad: @yihezkel added the UT and E2E you mentioned.

data/src/main/java/com/microsoft/azure/kusto/data/ClientFactory.java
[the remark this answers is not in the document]
larisaad: Thank you for looking into this.

yihezkel (approved):

PR #197, opened 2021-11-28
Updated surefire version to make tests run

AsafMah:
#### Pull Request Description

Various fixes to make our tests run.
---

#### Future Release Comment
**Breaking Changes:**
- Reverted mandatory DataFormat

**Features:**
- Added github actions for running builds

**Fixes:**
- Fixed pipeline to run and test
- Fixed failing tests and other minor fixes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,50 @@
+# This workflow will install the dependencies, run tests and lint every push
+
+name: Build Java
+
+on: push
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: ['8', '11', '17']
+    name: Java ${{ matrix.java }}
+    steps:
+#      - name: Download Maven
+#        run: |
+#          curl -sL https://www-eu.apache.org/dist/maven/maven-3/3.8.4/binaries/apache-maven-3.8.4-bin.zip -o maven.zip
+#          apt-get update
+#          apt-get -y install unzip
+#          unzip -d /usr/share maven.zip
+#          rm maven.zip
+#          ln -s /usr/share/apache-maven-3.8.4/bin/mvn /usr/bin/mvn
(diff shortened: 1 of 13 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,46 @@
...
+name: Build Java
+
+on: push
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: ['8', '11', '17']
```
yihezkel, line 12: Why 11 and 17?

AsafMah: LTS versions

yihezkel: But we only target 8

AsafMah: Really? that's not great, but we can remove them if so

yihezkel: It would require an effort to make it build with future versions. Not a huge effort, but it'd have to be prioritized.

.github/workflows/build.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Run the Maven verify phase
+        run: mvn --batch-mode --update-snapshots verify
+        env:
+          DM_CONNECTION_STRING: https://ingest-sdkse2etest.eastus.kusto.windows.net
+          ENGINE_CONNECTION_STRING: https://sdkse2etest.eastus.kusto.windows.net
+          TEST_DATABASE: e2e
+          APP_ID: ${{ secrets.APP_ID }}
+          APP_KEY: ${{ secrets.APP_KEY }}
+          TENANT_ID: ${{ secrets.TENANT_ID }}
+          AUTOMATIC_TESTING: 1
```
yihezkel, line 40: I suggest calling this `CI_EXECUTION` or `UNATTENDED_EXECUTION`

yihezkel (approved):

PR #198, opened 2021-12-12
Added ManagedStreaming improvements

AsafMah:
#### Pull Request Description

Adds new features to ManagedStreaming client:
- Exponential sleep with backoff
- Custom request id
-Added factory methods for easy construction
-Supports non-resettable streams
-Supports LeaveOpen
-Fallback on big files

---

#### Future Release Comment
_[Add description of your change, to include in the next release]_
_[Delete any or all irrelevant sections, e.g. if your change does not warrant a release comment at all]_

**Breaking Changes:**
- None

**Features:**
Adds new features to ManagedStreaming client:
- Exponential sleep with backoff
- Custom request id
-Added factory methods for easy construction
-Supports non-resettable streams
-Supports LeaveOpen
-Fallback on big files

**Fixes:**
- None

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
@@ -54,6 +54,27 @@ public class ConnectionStringBuilder {
         this.applicationNameForTracing = null;
     }
 
+    public ConnectionStringBuilder(ConnectionStringBuilder other){
+        this.clusterUrl = other.clusterUrl;
+        this.usernameHint = other.usernameHint;
+        this.applicationClientId = other.applicationClientId;
+        this.applicationKey = other.applicationKey;
+        this.x509Certificate = other.x509Certificate;
+        this.x509CertificateChain = other.x509CertificateChain;
+        this.privateKey = other.privateKey;
+        this.aadAuthorityId = other.aadAuthorityId;
+        this.accessToken = other.accessToken;
+        this.tokenProvider = other.tokenProvider;
+        this.managedIdentityClientId = other.managedIdentityClientId;
+        this.useDeviceCodeAuth = other.useDeviceCodeAuth;
+        this.useManagedIdentityAuth = other.useManagedIdentityAuth;
+        this.useUserPromptAuth = other.useUserPromptAuth;
+        this.userNameForTracing = other.userNameForTracing;
+        this.clientVersionForTracing = other.clientVersionForTracing;
+        this.applicationNameForTracing = other.applicationNameForTracing;
+    }
+
+
     public String getClusterUrl() {
         return clusterUrl;
     }
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
@@ -37,15 +37,7 @@ public class ExponentialRetry {
         return currentAttempt;
     }
 
-    public boolean shouldTry() {
-        return currentAttempt < maxAttempts;
-    }
-
-    public void doBackoff(){
-        if (!shouldTry()) {
-            throw new IllegalStateException("Max attempts exceeded");
-        }
-
+    public boolean doBackoff(){
         double sleepTime = getCurrentSleepSecs();
         double jitter = (float)Math.random() * maxJitterSecs;
         double sleepMs = (sleepTime + jitter) * 1000;
@@ -55,6 +47,8 @@ public class ExponentialRetry {
             throw new RuntimeException("Interrupted while sleeping");
         }
         currentAttempt++;
+
+        return currentAttempt < maxAttempts;
     }
 
 
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
@@ -2,34 +2,24 @@ package com.microsoft.azure.kusto.ingest;
 
 public class ExponentialRetry {
     private final int maxAttempts;
-    private final double sleepBaseSecs;
-    private final double maxJitterSecs;
+    double sleepBaseSecs;
+    double maxJitterSecs;
     private int currentAttempt;
 
     public ExponentialRetry(int maxAttempts) {
-        this(maxAttempts, 1.0, 1.0);
-    }
-
-    public ExponentialRetry(int maxAttempts, double sleepBaseSecs, double maxJitterSecs) {
         this.maxAttempts = maxAttempts;
-        this.sleepBaseSecs = sleepBaseSecs;
-        this.maxJitterSecs = maxJitterSecs;
+        this.sleepBaseSecs = 1.0;
+        this.maxJitterSecs = 1.0;
         this.currentAttempt = 0;
     }
 
-    public int getMaxAttempts() {
-        return maxAttempts;
+    public ExponentialRetry(ExponentialRetry other) {
+        this.maxAttempts = other.maxAttempts;
+        this.sleepBaseSecs = other.sleepBaseSecs;
+        this.maxJitterSecs = other.maxJitterSecs;
     }
 
-    public double getSleepBaseSecs() {
-        return sleepBaseSecs;
-    }
-
-    public double getMaxJitterSecs() {
-        return maxJitterSecs;
-    }
-
-    public double getCurrentSleepSecs() {
+    private double getCurrentSleepSecs() {
         return sleepBaseSecs * (float) Math.pow(2, currentAttempt);
     }
 
@@ -37,12 +27,15 @@ public class ExponentialRetry {
         return currentAttempt;
     }
 
-    public boolean doBackoff(){
-        double sleepTime = getCurrentSleepSecs();
-        double jitter = (float)Math.random() * maxJitterSecs;
-        double sleepMs = (sleepTime + jitter) * 1000;
+    public String nextTimeToSleep() {
+        return String.format("%s +~ %s seconds", getCurrentSleepSecs(), maxJitterSecs);
+    }
+
+    public boolean doBackoff() {
+        double jitterSeconds = (float) Math.random() * maxJitterSecs;
+        double sleepMs = (getCurrentSleepSecs() + jitterSeconds) * 1000;
         try {
-            Thread.sleep((long)sleepMs);
+            Thread.sleep((long) sleepMs);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             throw new RuntimeException("Interrupted while sleeping", e);
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
@@ -2,18 +2,18 @@ package com.microsoft.azure.kusto.ingest;
 
 public class ExponentialRetry {
     private final int maxAttempts;
-    private final double sleepBase;
-    private final double maxJitter;
+    private final double sleepBaseSecs;
+    private final double maxJitterSecs;
     private int currentAttempt;
 
     public ExponentialRetry(int maxAttempts) {
         this(maxAttempts, 1.0, 0.0);
     }
 
-    public ExponentialRetry(int maxAttempts, double sleepBaseSec, double maxJitterSec) {
+    public ExponentialRetry(int maxAttempts, double sleepBaseSecs, double maxJitterSecs) {
         this.maxAttempts = maxAttempts;
-        this.sleepBase = sleepBaseSec;
-        this.maxJitter = maxJitterSec;
+        this.sleepBaseSecs = sleepBaseSecs;
+        this.maxJitterSecs = maxJitterSecs;
         this.currentAttempt = 0;
     }
 
@@ -21,16 +21,16 @@ public class ExponentialRetry {
         return maxAttempts;
     }
 
-    public double getSleepBase() {
-        return sleepBase;
+    public double getSleepBaseSecs() {
+        return sleepBaseSecs;
     }
 
-    public double getMaxJitter() {
-        return maxJitter;
+    public double getMaxJitterSecs() {
+        return maxJitterSecs;
     }
 
-    public double getCurrentSleepMs() {
-        return sleepBase * (float) Math.pow(2, currentAttempt);
+    public double getCurrentSleepSecs() {
+        return sleepBaseSecs * (float) Math.pow(2, currentAttempt);
     }
 
     public int getCurrentAttempt() {
@@ -46,8 +46,8 @@ public class ExponentialRetry {
             throw new IllegalStateException("Max attempts exceeded");
         }
 
-        double sleepTime = getCurrentSleepMs();
-        double jitter = (float)Math.random() * maxJitter;
+        double sleepTime = getCurrentSleepSecs();
+        double jitter = (float)Math.random() * maxJitterSecs;
         double sleepMs = (sleepTime + jitter) * 1000;
         try {
             Thread.sleep((long)sleepMs);
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
@@ -1,16 +1,24 @@
 package com.microsoft.azure.kusto.ingest;
 
+import com.microsoft.azure.kusto.ingest.exceptions.IngestionClientException;
+import com.microsoft.azure.kusto.ingest.exceptions.IngestionServiceException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.invoke.MethodHandles;
+
 public class ExponentialRetry {
+    private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
     private final int maxAttempts;
     double sleepBaseSecs;
     double maxJitterSecs;
-    private int currentAttempt;
 
     public ExponentialRetry(int maxAttempts) {
         this.maxAttempts = maxAttempts;
         this.sleepBaseSecs = 1.0;
         this.maxJitterSecs = 1.0;
-        this.currentAttempt = 0;
     }
 
     public ExponentialRetry(ExponentialRetry other) {
@@ -19,30 +27,36 @@ public class ExponentialRetry {
         this.maxJitterSecs = other.maxJitterSecs;
     }
 
-    private double getCurrentSleepSecs() {
-        return sleepBaseSecs * (float) Math.pow(2, currentAttempt);
-    }
+    public <T> T execute(KustoCheckedFunction<Integer, T> function) throws IngestionClientException, IngestionServiceException {
+        for (int currentAttempt = 0; currentAttempt < maxAttempts; currentAttempt++) {
+            log.info("Attempt {}", currentAttempt);
 
-    public int getCurrentAttempt() {
-        return currentAttempt;
-    }
+            try {
+                T result = function.apply(currentAttempt);
+                if (result != null) {
+                    return result;
+                }
+            }
+            catch (Exception e) {
+                log.error("Error is permanent, stopping", e);
+                throw e;
+            }
 
-    public String nextTimeToSleep() {
-        return String.format("%s +~ %s seconds", getCurrentSleepSecs(), maxJitterSecs);
-    }
+            double currentSleepSecs = sleepBaseSecs * (float) Math.pow(2, currentAttempt);
 
-    public boolean doBackoff() {
-        double jitterSecs = (float) Math.random() * maxJitterSecs;
-        double sleepMs = (getCurrentSleepSecs() + jitterSecs) * 1000;
-        try {
-            Thread.sleep((long) sleepMs);
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
-            throw new RuntimeException("Interrupted while sleeping", e);
+            log.info("Operation failed, trying again after sleep of %{} +~ %{} seconds", currentSleepSecs, maxJitterSecs);
+
+            double jitterSecs = (float) Math.random() * maxJitterSecs;
+            double sleepMs = (currentSleepSecs + jitterSecs) * 1000;
+            try {
+                Thread.sleep((long) sleepMs);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException("Interrupted while sleeping", e);
+            }
         }
-        currentAttempt++;
 
-        return currentAttempt < maxAttempts;
+        return null;
     }
 
 
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
@@ -44,7 +44,8 @@ public class ExponentialRetry {
         try {
             Thread.sleep((long)sleepMs);
         } catch (InterruptedException e) {
-            throw new RuntimeException("Interrupted while sleeping");
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Interrupted while sleeping", e);
         }
         currentAttempt++;
 
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
@@ -4,7 +4,7 @@ public class ExponentialRetry {
     private final int maxAttempts;
     private final double sleepBase;
     private final double maxJitter;
-    private int retries;
+    private int currentAttempt;
 
     public ExponentialRetry(int maxAttempts) {
         this(maxAttempts, 1.0, 0.0);
@@ -14,7 +14,7 @@ public class ExponentialRetry {
         this.maxAttempts = maxAttempts;
         this.sleepBase = sleepBaseSec;
         this.maxJitter = maxJitterSec;
-        this.retries = 0;
+        this.currentAttempt = 0;
     }
 
     public int getMaxAttempts() {
@@ -30,18 +30,22 @@ public class ExponentialRetry {
     }
 
     public double getCurrentSleepMs() {
-        return sleepBase * (float) Math.pow(2, retries);
+        return sleepBase * (float) Math.pow(2, currentAttempt);
     }
 
     public int getCurrentAttempt() {
-        return retries;
+        return currentAttempt;
     }
 
-    public boolean shouldRetry() {
-        return retries < maxAttempts;
+    public boolean shouldTry() {
+        return currentAttempt < maxAttempts;
     }
 
     public void doBackoff(){
+        if (!shouldTry()) {
+            throw new IllegalStateException("Max attempts exceeded");
+        }
+
         double sleepTime = getCurrentSleepMs();
         double jitter = (float)Math.random() * maxJitter;
         double sleepMs = (sleepTime + jitter) * 1000;
@@ -50,7 +54,7 @@ public class ExponentialRetry {
         } catch (InterruptedException e) {
             throw new RuntimeException("Interrupted while sleeping");
         }
-        retries++;
+        currentAttempt++;
     }
 
 
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
@@ -32,8 +32,8 @@ public class ExponentialRetry {
     }
 
     public boolean doBackoff() {
-        double jitterSeconds = (float) Math.random() * maxJitterSecs;
-        double sleepMs = (getCurrentSleepSecs() + jitterSeconds) * 1000;
+        double jitterSecs = (float) Math.random() * maxJitterSecs;
+        double sleepMs = (getCurrentSleepSecs() + jitterSecs) * 1000;
         try {
             Thread.sleep((long) sleepMs);
         } catch (InterruptedException e) {
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
@@ -7,7 +7,7 @@ public class ExponentialRetry {
     private int currentAttempt;
 
     public ExponentialRetry(int maxAttempts) {
-        this(maxAttempts, 1.0, 0.0);
+        this(maxAttempts, 1.0, 1.0);
     }
 
     public ExponentialRetry(int maxAttempts, double sleepBaseSecs, double maxJitterSecs) {
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
@@ -29,7 +29,7 @@ public class ExponentialRetry {
 
     public <T> T execute(KustoCheckedFunction<Integer, T> function) throws IngestionClientException, IngestionServiceException {
         for (int currentAttempt = 0; currentAttempt < maxAttempts; currentAttempt++) {
-            log.info("Attempt {}", currentAttempt);
+            log.info("execute: Attempt {}", currentAttempt);
 
             try {
                 T result = function.apply(currentAttempt);
@@ -38,21 +38,21 @@ public class ExponentialRetry {
                 }
             }
             catch (Exception e) {
-                log.error("Error is permanent, stopping", e);
+                log.error("execute: Error is permanent, stopping", e);
                 throw e;
             }
 
             double currentSleepSecs = sleepBaseSecs * (float) Math.pow(2, currentAttempt);
-
-            log.info("Operation failed, trying again after sleep of %{} +~ %{} seconds", currentSleepSecs, maxJitterSecs);
-
             double jitterSecs = (float) Math.random() * maxJitterSecs;
             double sleepMs = (currentSleepSecs + jitterSecs) * 1000;
+
+            log.info("execute: Attempt {} failed, trying again after sleep of {} seconds", currentAttempt, sleepMs / 1000);
+
             try {
                 Thread.sleep((long) sleepMs);
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
-                throw new RuntimeException("Interrupted while sleeping", e);
+                throw new RuntimeException("execute: Interrupted while sleeping", e);
             }
         }
 
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
@@ -1,0 +1,57 @@
+package com.microsoft.azure.kusto.ingest;
+
+public class ExponentialRetry {
+    private final int maxAttempts;
+    private final double sleepBase;
+    private final double maxJitter;
+    private int retries;
+
+    public ExponentialRetry(int maxAttempts) {
+        this(maxAttempts, 1.0, 0.0);
+    }
+
+    public ExponentialRetry(int maxAttempts, double sleepBaseSec, double maxJitterSec) {
+        this.maxAttempts = maxAttempts;
+        this.sleepBase = sleepBaseSec;
+        this.maxJitter = maxJitterSec;
+        this.retries = 0;
+    }
+
+    public int getMaxAttempts() {
+        return maxAttempts;
+    }
+
+    public double getSleepBase() {
+        return sleepBase;
+    }
+
+    public double getMaxJitter() {
+        return maxJitter;
+    }
+
+    public double getCurrentSleepMs() {
+        return sleepBase * (float) Math.pow(2, retries);
+    }
+
+    public int getCurrentAttempt() {
+        return retries;
+    }
+
+    public boolean shouldRetry() {
+        return retries < maxAttempts;
+    }
+
+    public void doBackoff(){
+        double sleepTime = getCurrentSleepMs();
+        double jitter = (float)Math.random() * maxJitter;
+        double sleepMs = (sleepTime + jitter) * 1000;
+        try {
+            Thread.sleep((long)sleepMs);
+        } catch (InterruptedException e) {
+            throw new RuntimeException("Interrupted while sleeping");
+        }
+        retries++;
+    }
+
+
+}
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -46,12 +46,11 @@ public abstract class IngestClientBase {
             URIBuilder existingEndpoint;
             try {
                 existingEndpoint = new URIBuilder(dataSource);
-                endpointUriToSuggestStr = amendEndpointUri(existingEndpoint);
+                endpointUriToSuggestStr = emendEndpointUri(existingEndpoint);
             } catch (URISyntaxException e) {
                 log.error("Couldn't parse dataSource '{}', so no suggestion can be made.", dataSource, e);
-            }
-            catch (IllegalArgumentException e) {
-                log.error("URI is already in the correct format '{}', so no suggestion can be made.", dataSource, e);
+            } catch (IllegalArgumentException e) {
+                log.error("URL is already in the correct format '{}', so no suggestion can be made.", dataSource, e);
             }
         }
 
@@ -60,5 +59,5 @@ public abstract class IngestClientBase {
 
     protected abstract String retrieveServiceType() throws IngestionServiceException, IngestionClientException;
 
-    protected abstract String amendEndpointUri(URIBuilder existingEndpoint);
+    protected abstract String emendEndpointUri(URIBuilder existingEndpoint);
 }
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -2,6 +2,7 @@ package com.microsoft.azure.kusto.ingest;
 
 import com.microsoft.azure.kusto.ingest.exceptions.IngestionClientException;
 import com.microsoft.azure.kusto.ingest.exceptions.IngestionServiceException;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.client.utils.URIBuilder;
 import org.slf4j.Logger;
@@ -45,7 +46,7 @@ public abstract class IngestClientBase {
             URIBuilder existingEndpoint;
             try {
                 existingEndpoint = new URIBuilder(dataSource);
-                endpointUriToSuggestStr = emendEndpointUri(existingEndpoint);
+                endpointUriToSuggestStr = amendEndpointUri(existingEndpoint);
             } catch (URISyntaxException e) {
                 log.error("Couldn't parse dataSource '{}', so no suggestion can be made.", dataSource, e);
             }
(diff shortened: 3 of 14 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -45,16 +46,19 @@ protected String generateEndpointSuggestion(String existingSuggestedEndpointUri,
             URIBuilder existingEndpoint;
             try {
                 existingEndpoint = new URIBuilder(dataSource);
-                endpointUriToSuggestStr = emendEndpointUri(existingEndpoint);
+                endpointUriToSuggestStr = amendEndpointUri(existingEndpoint);
```
yihezkel, line 49: It's not a common word, but the more accurate term here is "emend", not "amend"

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
```diff
@@ -59,6 +59,15 @@
         this.azureStorageClient = azureStorageClient;
     }
 
+    public static String generateDmUriSuggestion(URIBuilder existingEndpoint) {
+        if (existingEndpoint.getHost().startsWith(INGEST_PREFIX))
```
yihezkel, line 63: nit: Case insensitive

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -45,16 +46,19 @@ protected String generateEndpointSuggestion(String existingSuggestedEndpointUri,
             URIBuilder existingEndpoint;
             try {
                 existingEndpoint = new URIBuilder(dataSource);
-                endpointUriToSuggestStr = emendEndpointUri(existingEndpoint);
+                endpointUriToSuggestStr = amendEndpointUri(existingEndpoint);
             } catch (URISyntaxException e) {
                 log.error("Couldn't parse dataSource '{}', so no suggestion can be made.", dataSource, e);
             }
+            catch (IllegalArgumentException e) {
```
yihezkel, line 53: nit: `catch` should be on the previous line

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -45,16 +46,19 @@ protected String generateEndpointSuggestion(String existingSuggestedEndpointUri,
             URIBuilder existingEndpoint;
             try {
                 existingEndpoint = new URIBuilder(dataSource);
-                endpointUriToSuggestStr = emendEndpointUri(existingEndpoint);
+                endpointUriToSuggestStr = amendEndpointUri(existingEndpoint);
             } catch (URISyntaxException e) {
                 log.error("Couldn't parse dataSource '{}', so no suggestion can be made.", dataSource, e);
             }
+            catch (IllegalArgumentException e) {
+                log.error("URI is already in the correct format '{}', so no suggestion can be made.", dataSource, e);
```
yihezkel, line 54: It's specifically a URL. A URI is much broader.

I'm debating whether this is worth erroring, or warning. WDYT?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
```diff
@@ -59,6 +59,15 @@
...
+        {
+            throw new IllegalArgumentException("The URI is already a DM URI");
```
yihezkel, line 65: 1) This is specifically a URL, not just a URI
2) It'd be more accurate to say, "The URL is already formatted as the suggested DM endpoint, so no suggestion can be made"

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
```diff
@@ -269,9 +278,8 @@ public IngestionResult ingestFromResultSet(ResultSetSourceInfo resultSetSourceIn
     }
 
     @Override
-    protected String emendEndpointUri(URIBuilder existingEndpoint) {
-        existingEndpoint.setHost(INGEST_PREFIX + existingEndpoint.getHost());
-        return existingEndpoint.toString();
+    protected String amendEndpointUri(URIBuilder existingEndpoint) {
+        return generateDmUriSuggestion(existingEndpoint);
```
yihezkel, line 282: Why create a new method? I usually think 1-line methods are harder to read/maintain.

In this case, the way to emend the URL is by suggesting the DM endpoint, whereas in the `StreamingIngestClient` we do that by suggesting the Engine endpoint.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -49,6 +50,15 @@
         this.streamingClient = streamingClient;
     }
 
+    public static String generateEngineUriSuggestion(URIBuilder existingEndpoint) {
+        if (!existingEndpoint.getHost().startsWith(IngestClientBase.INGEST_PREFIX)) {
+            throw new IllegalArgumentException("The URI is already an engine URI");
```
yihezkel, line 55: 1. This is specifically a URL, not just a URI
2. It'd be more accurate to say, "The URL is already formatted as the suggested Engine endpoint, so no suggestion can be made"

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
yihezkel: Actually I now see that you're reusing this method, so in that context this makes sense.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -13,37 +13,74 @@
...
+    final QueuedIngestClient queuedIngestClient;
+    final StreamingIngestClient streamingIngestClient;
+    private final ExponentialRetry exponentialRetryTemplate;
+
+    /**
+     * Creates a new ManagedStreamingIngestClient from a DM connection string.
+     * This method infers the engine connection string from the DM connection string.
+     * For advanced usage, use {@link ManagedStreamingIngestClient#ManagedStreamingIngestClient(ConnectionStringBuilder, ConnectionStringBuilder)}
+     * @param dmConnectionString dm connection string
+     * @return a new ManagedStreamingIngestClient
```
yihezkel, line 57: Add `throws` to the javadoc

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -13,37 +13,74 @@
...
+     */
+    public static ManagedStreamingIngestClient fromDmConnectionString(ConnectionStringBuilder dmConnectionString) throws URISyntaxException {
```
yihezkel, line 59: I'm sure I'm just missing something, but I don't understand the reason why we'd want a method to infer the other endpoint. In general, we've been moving away from inferring the DM/Engine endpoints from each other, since there are scenarios when it doesn't work. We've been moving towards explicit.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -13,37 +13,74 @@
...
+        engineConnectionString.setClusterUrl(StreamingIngestClient.generateEngineUriSuggestion(new URIBuilder(dmConnectionString.getClusterUrl())));
+        return new ManagedStreamingIngestClient(dmConnectionString, engineConnectionString);
+    }
+
+    /**
+     * Creates a new ManagedStreamingIngestClient from an engine connection string.
+     * This method infers the DM connection string from the engine connection string.
+     * For advanced usage, use {@link ManagedStreamingIngestClient#ManagedStreamingIngestClient(ConnectionStringBuilder, ConnectionStringBuilder)}
+     * @param engineConnectionString engine connection string
+     * @return a new ManagedStreamingIngestClient
```
yihezkel, line 70: Add `throws` to javadoc

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -52,6 +89,17 @@ public ManagedStreamingIngestClient(ResourceManager resourceManager,
...
+    }
+
+    ManagedStreamingIngestClient(ResourceManager resourceManager,
+                                        AzureStorageClient storageClient,
+                                        StreamingClient streamingClient,
+                                        ExponentialRetry retryTemplate) {
+        log.info("Creating a new ManagedStreamingIngestClient from raw parts");
+        queuedIngestClient = new QueuedIngestClient(resourceManager, storageClient);
+        streamingIngestClient = new StreamingIngestClient(streamingClient);
+        exponentialRetryTemplate = retryTemplate;
```
yihezkel, line 102: Just wondering out loud - do they need this kind of configurability that they have full control over the ExponentialRetry class to pass in? Don't we know the service better than they do to to set at least the `sleepBaseSecs`? I'd expect them to pass in 1 or 2 config params, or keep `sleepBaseSecs` as non-configurable.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -112,15 +160,50 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
+        byte[] streamingBytes;
+
+        try {
+            streamingBytes = IOUtils.readNBytes(streamSourceInfo.getStream(), MAX_STREAMING_SIZE_BYTES + 1);
+        } catch (IOException e) {
+            throw new IngestionClientException("Failed to read from stream.", e);
         }
-        streamSourceInfo.setLeaveOpen(true);
 
+        ByteArrayInputStream byteArrayStream = new ByteArrayInputStream(streamingBytes);
```
yihezkel, line 176: I don't see how closing is being managed, but if not fully managed, we probably want this to be within a `try-with-resources` block.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -112,15 +160,50 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
+
+        ExponentialRetry retry = new ExponentialRetry(exponentialRetryTemplate.getMaxAttempts(), exponentialRetryTemplate.getSleepBaseSecs(),
+                exponentialRetryTemplate.getMaxJitterSecs());
         try {
-            for (int i = 0; i < MAX_RETRY_CALLS; i++) {
+            while (retry.shouldTry()) {
                 try {
-                    return streamingIngestClient.ingestFromStream(streamSourceInfo, ingestionProperties);
+                    log.info("Streaming ingest attempt {}", retry.getCurrentAttempt());
+                    String clientRequestId = String.format("KJC.executeManagedStreamingIngest;%s;%d", sourceId,
```
yihezkel, line 204: I'm sure you already are, but just calling out that we should be deliberate in our `clientRequestId` including when we change approaches mid-ingest (at first glance I'd expect that the CRID shouldn't change).

ingest/src/test/java/com/microsoft/azure/kusto/ingest/CloseableByteArrayInputStream.java
```diff
@@ -0,0 +1,28 @@
...
+    public int read(byte[] b) throws IOException {
+        if (closed) {
+            throw new IOException("Stream is closed");
+        }
+        return super.read(b);
+    }
+
+    @Override
+    public void close() {
+        closed = true;
```
yihezkel, line 26: Since we don't call `super()`, it looks like it's not actually closed. I need to reread this PR to better understand the semantics and intent here, but either way I suspect we can name the class to be clearer that we don't allow the stream to be closed the traditional way.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -54,6 +54,27 @@ private ConnectionStringBuilder(String clusterUrl) {
         this.applicationNameForTracing = null;
     }
 
+    public ConnectionStringBuilder(ConnectionStringBuilder other){
+        this.clusterUrl = other.clusterUrl;
+        this.usernameHint = other.usernameHint;
+        this.applicationClientId = other.applicationClientId;
+        this.applicationKey = other.applicationKey;
+        this.x509Certificate = other.x509Certificate;
```
yihezkel, line 62: I'm not sure it's necessary here because CSBs have such a short useful lifetime (just for creating the client), but nevertheless I'd deep copy mutable objects here to prevent hard-to-find and unintended bugs where changing 1 CSB could affect the other.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
```diff
@@ -0,0 +1,61 @@
...
+        return currentAttempt;
+    }
+
+    public boolean shouldTry() {
+        return currentAttempt < maxAttempts;
+    }
+
+    public void doBackoff(){
+        if (!shouldTry()) {
+            throw new IllegalStateException("Max attempts exceeded");
```
yihezkel, line 46: I would not expect throwing an exception when finding that we've already reached the maximum attempts.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: We talked about it in the python implementation.
I am very much for this, because for the most common case the user is using the same CSB for both with the only difference is the "ingest-". 
I think it's unnecessary repetition and is prone to bugs (changing in one and missing in the other)

In advanced cases where it's not the user can still explicitly provide both of them, but I think for the common case that's better.

If you're strongly against it we can a have a team discussion (before we release the python version)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
AsafMah: I think it should be an error. 99% the user has made a mistake, I want them to be able to find it easily

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
AsafMah: So I can close this?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: It helps for the tests.
Either way that's a package-private c-tor, so the users can't access any of it

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: ByteArrayInputStream's close is a no-op:
https://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayInputStream.html

Maybe I'll add a comment?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/CloseableByteArrayInputStream.java
AsafMah: ByteArrayInputStream's close is a no-op:
https://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayInputStream.html

Which is why I made this class.
I could call super for clarity, but it doesn't do anything.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
AsafMah: What is the correct way here?
The method is called while the object is in the wrong state

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
AsafMah: The potential mutable objects are:
```
private List<X509Certificate> x509CertificateChain;
private Callable<String> tokenProvider;
```

Other than list, I don't know how to clone the rest

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: If you mean when we move to queued - we don't use ClientRequestId in queued, not sure how that would work

yihezkel: Right, great point :-)

But doesn't this approach call methods in both ManagedStreaming and Streaming classes? I'm sure it is, but just making sure our CRID is consistent.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
yihezkel: I think `this.x509Certificate = SerializationUtils.clone(other.x509Certificate);` would work.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
yihezkel: I hear your point, but I wouldn't personally call it in the "wrong state". I'd say it's perfectly normal for `ExponentialRetry` to reach the max tries, but just its terminal state.

If we did want an exception, I'd expect a checked exception.

I'd have the method return a `boolean`.

yihezkel: Wait, I'm looking now at how it's used (probably should've done that before), and see that you check `shouldRetry()` before calling `doBackoff()`, in which case if we've exhausted our tries we're indeed in an unexpected state.

How about simplifying it by removing `shouldRetry()` or at least making it `protected`, and `doBackoff()` would return false if it exhausted its retries. So the interface relevant to invoking this class would simply be `while(doBackoff())`.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
yihezkel: Nah, I hear this approach. Since we also provide a way to be explicit, I think this approach is good too.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
yihezkel: To help with the tests, I'd make the members of the `ExponentialRetry` class `protected`.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
yihezkel: Ah, thanks for explaining. Yeah, I think a comment would help make it clear.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
yihezkel: Closed!

ingest/src/test/java/com/microsoft/azure/kusto/ingest/CloseableByteArrayInputStream.java
yihezkel: Nah, no need. I think just a comment would help.
Thanks for explaining.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        double sleepTime = getCurrentSleepSecs();
+        double jitter = (float)Math.random() * maxJitterSecs;
+        double sleepMs = (sleepTime + jitter) * 1000;
+        try {
+            Thread.sleep((long)sleepMs);
+        } catch (InterruptedException e) {
+            throw new RuntimeException("Interrupted while sleeping");
```
yihezkel, line 55: I thought in Java we shouldn't swallow `InterruptedException`s because it leave us in a bad state.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: As far as I tested it does work, if you think there's a flow I missed I'd be happy to know

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: Why?

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
AsafMah: Not sure I'm comfortable with it
What if there are fields that can't be serialized?
I think it would lead to subtle errors

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
AsafMah: I don't want to add it to the whole hierarchy, so I changed it to handle like SO recommends:
```
        try {
            Thread.sleep((long)sleepMs);
        } catch (InterruptedException e) {
            Thread.currentThread().interrupt();
            throw new RuntimeException("Interrupted while sleeping", e);
        }
```

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
AsafMah: Like so?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/CloseableByteArrayInputStream.java
AsafMah: Is the comment ok?

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
yihezkel: The method signature is:
`public static <T extends Serializable> T clone(final T object)`

It indeeds requires that passed parameters are serializable, but the good news is that it throws `SerializationException` if it can't, so any errors should be easy to identify.

Of course any subtype can be passed, but `X509Certificate` itself seems `Serializable`, because its parent `Certificate` is, and it doesn't contain any additional members that can be set. (It has 2 transient members, but those are simply from the parent object)
And `PrivateKey` implements `Key`, which is serializable.

We can't really deep-copy `Callable`.

yihezkel: Actually, I see that `X509Certificate` and `PrivateKey` seem to effectively be immutable, as neither has methods to change their state (except `destroy()`, but arguably that should be allowed to destroy).

So nevermind, there isn't a reasonable need to deep-copy these 2.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
yihezkel: So that tests can access (and change) `ExponentialRetry`'s members.

In Java, the test class is generally placed within the same package as the main class. In fact, it's usually in an identical folder, except within `test` instead of `main`, for the very reason of getting this "trick" to work.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -112,15 +162,51 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
+                streamSourceInfo.getStream().close();
+            } catch (IOException e) {
+                log.warn("Failed to close stream", e);
+            }
+        }
+
+        StreamSourceInfo managedSourceInfo = new StreamSourceInfo(byteArrayStream, true, sourceId);
+        managedSourceInfo.setCompressionType(streamSourceInfo.getCompressionType());
+
+        ExponentialRetry retry = new ExponentialRetry(exponentialRetryTemplate.getMaxAttempts(), exponentialRetryTemplate.getSleepBaseSecs(),
```
yihezkel, line 201: nit: I'd personally create a copy constructor:
`public ExponentialRetry(ExponentialRetry other)`

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -112,15 +162,51 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
+        }
+
+        // NOTE: ByteArrayInputStream's close method is a no-op, so we don't need to close it.
+        ByteArrayInputStream byteArrayStream = new ByteArrayInputStream(streamingBytes);
+
+        if (streamingBytes.length > MAX_STREAMING_SIZE_BYTES) {
+            log.info("Stream size is greater than max streaming size (%d bytes). Falling back to queued.");
+            StreamSourceInfo managedSourceInfo = new StreamSourceInfo(new SequenceInputStream(byteArrayStream, streamSourceInfo.getStream()),
+                    streamSourceInfo.isLeaveOpen(), sourceId);
+            managedSourceInfo.setCompressionType(streamSourceInfo.getCompressionType());
```
yihezkel, line 185: It seems to be a mistake that the `StreamSourceInfo` constructor doesn't have an overload that accepts a `compressionType` parameter (same comment for line 198 below).

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -139,21 +225,23 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
                         }
                     }
 
-                    log.info("Streaming ingestion failed, trying again", e);
+                    log.info(String.format("Streaming ingestion failed, trying again after sleep of %s +~ %s seconds", retry.getCurrentSleepSecs(),
```
yihezkel, line 228: I think this would be best encapsulated within `ExponentialRetry`. Here we have an external class tasked with printing the state of that class.

I'm not sure why anyone else even needs access to `currentSleepSecs` and `maxJitterSecs`. I don't think I would provide a getter for them.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -139,21 +225,23 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
                     try {
-                        streamSourceInfo.getStream().reset();
+                        managedSourceInfo.getStream().reset();
                     } catch (IOException ioException) {
-                        throw new IngestionClientException("Ingestion failed transiently but the stream isn't resettable therefore ingestion wasn't retried", ioException);
+                        throw new IngestionClientException("Failed to reset stream", ioException);
                     }
                 }
-            }
+            } while (retry.doBackoff());
```
yihezkel, line 237: I suspect a new user would guess that `retry.doBackoff()` returns true if it ran, and false if it didn't. I think they'd be surprised to learn that it returns true if it should run a further time, and otherwise false.

I'd implement this as a `while(retry.doBackoff())`, where `retry.doBackoff()` returns true if it ran, and false if it didn't.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
```diff
@@ -0,0 +1,56 @@
...
+    public double getCurrentSleepSecs() {
+        return sleepBaseSecs * (float) Math.pow(2, currentAttempt);
+    }
+
+    public int getCurrentAttempt() {
+        return currentAttempt;
+    }
+
+    public boolean doBackoff(){
+        double sleepTime = getCurrentSleepSecs();
```
yihezkel, line 41: nit: I don't think this assignment helps

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -112,15 +162,51 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
+        } catch (IOException e) {
+            throw new IngestionClientException("Failed to read from stream.", e);
+        }
+
+        // NOTE: ByteArrayInputStream's close method is a no-op, so we don't need to close it.
+        ByteArrayInputStream byteArrayStream = new ByteArrayInputStream(streamingBytes);
+
+        if (streamingBytes.length > MAX_STREAMING_SIZE_BYTES) {
+            log.info("Stream size is greater than max streaming size (%d bytes). Falling back to queued.");
+            StreamSourceInfo managedSourceInfo = new StreamSourceInfo(new SequenceInputStream(byteArrayStream, streamSourceInfo.getStream()),
```
yihezkel, line 183: I get that `streamSourceInfo` is already used in this scope, but this `StreamSourceInfo` is only used for streaming, right? When I think of `ManagedStreaming`, I think it can be used for either queued or streaming.

yihezkel: Sorry, it can indeed be used for either. I missed that!

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
```diff
@@ -0,0 +1,56 @@
+package com.microsoft.azure.kusto.ingest;
+
+public class ExponentialRetry {
```
yihezkel, line 3: nit: I think they're both valid design decisions so I'm not requesting a change here, but I personally would've passed to this class a `Function` that makes the call + error handling logic, and `doBackoff()` actually kicks off making the call repeatedly with backoff.
I just see that approach as better encapsulated.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -120,13 +134,19 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
             throw new IngestionClientException(String.format("Mapping reference must be specified for DataFormat '%s' in streaming ingestion.", dataFormat.name()));
         }
 
+        ClientRequestProperties clientRequestProperties = null;
+        if (clientRequestId != null) {
```
yihezkel, line 138: nit: Currently this either has a real value or `null`, but I think it's future-safer to instead test `StringUtils.isBlank()`.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/CloseableByteArrayInputStream.java
```diff
@@ -0,0 +1,29 @@
+package com.microsoft.azure.kusto.ingest;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.io.IOException;
+
+// NOTE: ByteArrayInputStream's close method is a no-op, this class makes reading after close into an error, for test purposes
```
yihezkel, line 7: nit: I'd remove "NOTE: "

ingest/src/test/java/com/microsoft/azure/kusto/ingest/CloseableByteArrayInputStream.java
yihezkel: Great, thanks! See nit comment, though

ingest/src/test/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClientTest.java
```diff
@@ -14,29 +16,56 @@
...
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.atLeast;
+import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
```
yihezkel, line 68: Down with wildcards!!

ingest/src/test/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClientTest.java
```diff
@@ -263,24 +306,35 @@ void IngestFromFile_CompressedJson() throws Exception {
...
          * When the given stream content is already compressed, the user must specify that in the stream source info.
          * This method verifies if the stream was compressed correctly.
          */
         InputStream stream = argumentCaptor.getValue();
         verifyCompressedStreamContent(stream, data);
+        verifyClientRequestId(0, sourceId);
+        if (leaveOpen) {
+            assertDoesNotThrow(() -> inputStream.read(new byte[1]));
+        } else {
+            //noinspection ResultOfMethodCallIgnored - we are expecting an exception, so we don't care about the result
```
yihezkel, line 335: I'm not a big fan of adding these, especially if the only issue we have without them is a SonarLint warning. Personally, I'd rather make a dummy assignment if we're bothered by the warning.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
yihezkel: Yeah that looks perfect!

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: I think the more surprising behavior would be that `retry.doBackoff()` does nothing the first time you call it, but after that it sleeps.

But I see your point, and I'm not sure what's "more correct"

yihezkel: I do think `doBackoff()` should return true/execute the first time you call it. I'd only return false if it didn't run, which would be the last time it's called.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
AsafMah: So can I close it?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: Question - intellij says that the copy c-tor  should copy all fields, but I think we shouldn't copy "currentAttempt". WDYT?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClientTest.java
AsafMah: Replaced with assignment, looks good?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: The name doBackoff implies that it's doing something.
I don't see any sense in putting it at the start of the loop, just so it could do nothing the first time.

I agree that maybe the return value of "returns whether or not you can call it again" is unclear, but I don't think changing the semantics is the solution.

Maybe a comment? or something else?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
AsafMah: It would be kind of annoying here because in the body of the loop we do a `return`, so we'll need to make it work with generics and everything.
I think the current approach is good enough for now.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
```diff
@@ -0,0 +1,49 @@
...
+    public boolean doBackoff() {
+        double jitterSecs = (float) Math.random() * maxJitterSecs;
+        double sleepMs = (getCurrentSleepSecs() + jitterSecs) * 1000;
+        try {
+            Thread.sleep((long) sleepMs);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Interrupted while sleeping", e);
+        }
+        currentAttempt++;
```
yihezkel, line 43: Rename as `prevAttempt`

yihezkel: And make 1-based index

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
```diff
@@ -0,0 +1,49 @@
...
+
+        return currentAttempt < maxAttempts;
```
yihezkel, line 45: Move to beginning:
`
if (prevAttempt >= maxAttempt) {
    return false;
}
`

And then at the end just `return true;`

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
yihezkel: Nevermind my suggestion about `do`/`while` --> `while` (was based on the `function` approach we decided not to pursue).

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
yihezkel: Yeah, without `currentAttempt`

yihezkel: Maybe mark `currentAttempt` as `transient`

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public ExponentialRetry(ExponentialRetry other) {
+        this.maxAttempts = other.maxAttempts;
+        this.sleepBaseSecs = other.sleepBaseSecs;
+        this.maxJitterSecs = other.maxJitterSecs;
+    }
+
+    private double getCurrentSleepSecs() {
+        return sleepBaseSecs * (float) Math.pow(2, currentAttempt);
```
yihezkel, line 23: `currentAttempt` --> `currentAttempt - 1` (because of 1-based index)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
```diff
@@ -19,30 +27,36 @@ public ExponentialRetry(ExponentialRetry other) {
...
-        double jitterSecs = (float) Math.random() * maxJitterSecs;
-        double sleepMs = (getCurrentSleepSecs() + jitterSecs) * 1000;
-        try {
-            Thread.sleep((long) sleepMs);
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
-            throw new RuntimeException("Interrupted while sleeping", e);
+    public <T> T execute(KustoCheckedFunction<Integer, T> function) throws IngestionClientException, IngestionServiceException {
+        for (int currentAttempt = 0; currentAttempt < maxAttempts; currentAttempt++) {
+            log.info("Attempt {}", currentAttempt);
```
yihezkel, line 32: "ExponentialRetry.execute: Attempt {}"

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
```diff
@@ -19,30 +27,36 @@ public ExponentialRetry(ExponentialRetry other) {
...
+                }
+            }
+            catch (Exception e) {
+                log.error("Error is permanent, stopping", e);
+                throw e;
+            }
+
+            double currentSleepSecs = sleepBaseSecs * (float) Math.pow(2, currentAttempt);
+
+            log.info("Operation failed, trying again after sleep of %{} +~ %{} seconds", currentSleepSecs, maxJitterSecs);
```
yihezkel, line 47: Why not move this to after where `sleepMs` is defined, and change to:
`log.info("ExponentialRetry.execute: Attempt {} failed, trying again after sleep of {} milliseconds", currentAttempt, sleepMs);`

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -214,23 +214,27 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
-                                log.error("Error is permanent, stopping.");
                                 throw e;
                             }
                         } catch (JSONException je) {
                             log.info("Failed to parse json in exception, continuing.", je);
                         }
                     }
 
-                    log.info(String.format("Streaming ingestion failed, trying again after sleep of %s", retry.nextTimeToSleep()), e);
+                    log.info(String.format("Streaming ingestion failed attempt %d", currentAttempt), e);
```
yihezkel, line 224: Reminder to change this to `currentAttempt + 1` if we stick with 0-based indexing

yihezkel (approved):
Is this not beautiful? I personally love it :-)

Everything related to exponentially retrying is encapsulated within `ExponentialRetry`. It's also reusable now for anything else. A model for our C# code and beyond!

(Added a few minor suggestions surrounding logging and whether `currentAttempt` should be logged with 0- or 1-based indexing)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: The question is, if I display it as "attempt 1" should the ClientRequestId also be 1 indexed? Because that feels wrong (the CRI is machine readable), but not having the same value between them also feels wrong..

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ExponentialRetry.java
AsafMah: Great idea (I used seconds)

Also, we don't need to write the class name, that's part of the log:
[main] INFO com.microsoft.azure.kusto.ingest.ExponentialRetry - execute: Attempt 0

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
yihezkel: As discussed, no issue

PR #199, opened 2021-12-12
Minor pom and code improvements

yihezkel:
#### Pull Request Description

Minor pom and code improvements

---

#### Future Release Comment
**Fixes:**
- Specify runtime scope for slf4j's impl (simple)
- Do not exclude the slf4j's impl from maven-dependency-plugin
- Relatedly, turn on ignoreNonCompile in maven-dependency-plugin
- Minor readability improvement in ClientImpl constructor

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -41,18 +41,15 @@ public class ClientImpl implements Client, StreamingClient {
     private final String userNameForTracing;
 
     public ClientImpl(ConnectionStringBuilder csb) throws URISyntaxException {
-        String url = csb.getClusterUrl();
-        URI clusterUri = new URI(url);
-        String host = clusterUri.getHost();
-        Objects.requireNonNull(clusterUri.getAuthority(), "clusterUri.authority");
-
-        String auth = clusterUri.getAuthority().toLowerCase();
+        URI clusterUrlForParsing = new URI(csb.getClusterUrl());
+        String host = clusterUrlForParsing.getHost();
+        Objects.requireNonNull(clusterUrlForParsing.getAuthority(), "clusterUri.authority");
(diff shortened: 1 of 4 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -41,18 +41,15 @@
...
-
-        String auth = clusterUri.getAuthority().toLowerCase();
+        URI clusterUrlForParsing = new URI(csb.getClusterUrl());
+        String host = clusterUrlForParsing.getHost();
+        Objects.requireNonNull(clusterUrlForParsing.getAuthority(), "clusterUri.authority");
+        String auth = clusterUrlForParsing.getAuthority().toLowerCase();
         if (host == null && auth.endsWith(FEDERATED_SECURITY_SUFFIX)) {
-            url = new URIBuilder().setScheme(clusterUri.getScheme()).setHost(auth.substring(0, clusterUri.getAuthority().indexOf(FEDERATED_SECURITY_SUFFIX))).toString();
-            csb.setClusterUrl(url);
+            csb.setClusterUrl(new URIBuilder().setScheme(clusterUrlForParsing.getScheme()).setHost(auth.substring(0, clusterUrlForParsing.getAuthority().indexOf(FEDERATED_SECURITY_SUFFIX))).toString());
```
AsafMah, line 49: Editing the user's connection string is pretty unexpected
Is there a good reason to do it? do we do it anywhere else?

yihezkel: I don't know the answer to either, unfortunately. My intent was just to make this code more readable.

yihezkel: It seems changing the user's connection string was intentional, per https://github.com/Azure/azure-kusto-java/commit/c5079020c972903c61c8051b318ace921523bfec

AsafMah (approved):

PR #202, opened 2021-12-16
Fix DataFormat bugs and add enhancements

yihezkel:
#### Pull Request Description

Fix DataFormat bugs and add enhancements

---

#### Future Release Comment

**Features:**
- Enhanced existing DataFormat enum to include IsMappingRequired and isCompressible for better OO design

**Fixes:**
- Inconsistent use of null IngestionProperty.DataFormat, allowing for missing extension from the uploaded blob, when it should default to .csv
- getMappingReferencce() would fail unnecessarily because it was comparing the DataFormat to the IngestionMappingKind, because the former doesn't have a one-to-one with the latter, but instead a one-to-many relationship with the latter

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -59,7 +59,6 @@ public abstract class IngestClientBase {
 
     protected abstract String emendEndpointUri(URIBuilder existingEndpoint);
 
-    // We don't support compression of Parquet and Orc files
     static boolean shouldCompress(CompressionType sourceCompressionType, IngestionProperties.DataFormat dataFormat) {
         return (sourceCompressionType == null) && (dataFormat == null || dataFormat.isCompressible());
     }
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -2,6 +2,7 @@ package com.microsoft.azure.kusto.ingest;
 
 import com.microsoft.azure.kusto.ingest.exceptions.IngestionClientException;
 import com.microsoft.azure.kusto.ingest.exceptions.IngestionServiceException;
+import com.microsoft.azure.kusto.ingest.source.CompressionType;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.client.utils.URIBuilder;
 import org.slf4j.Logger;
@@ -57,4 +58,9 @@ public abstract class IngestClientBase {
     protected abstract String retrieveServiceType() throws IngestionServiceException, IngestionClientException;
 
     protected abstract String emendEndpointUri(URIBuilder existingEndpoint);
+
+    // We don't support compression of Parquet and Orc files
+    static boolean shouldCompress(CompressionType sourceCompressionType, IngestionProperties.DataFormat dataFormat) {
+        return (sourceCompressionType == null) && (dataFormat == null || dataFormat.isCompressible());
+    }
 }
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
@@ -9,6 +9,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.microsoft.azure.kusto.data.Ensure;
 import com.microsoft.azure.kusto.ingest.exceptions.IngestionClientException;
 import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -250,6 +251,7 @@ public class IngestionProperties {
      *
      * @return The DataFormat if it exists, and otherwise defaults to CSV
      */
+    @NotNull
     public DataFormat getDataFormat() {
         return dataFormat != null ? dataFormat : DataFormat.csv;
     }
@@ -319,7 +321,7 @@ public class IngestionProperties {
             }
         }
 
-        if (dataFormat != null && ingestionMappingKind != null && ingestionMappingKind != IngestionMapping.IngestionMappingKind.unknown && !dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
+        if (dataFormat != null &&  dataFormat.isMappingRequired() && !dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
                 message += String.format("Wrong ingestion mapping for format '%s'; found '%s' mapping kind.", dataFormat.name(), ingestionMappingKind.name());
         }
 
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
@@ -8,8 +8,11 @@ import com.fasterxml.jackson.annotation.PropertyAccessor;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.microsoft.azure.kusto.data.Ensure;
 import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -27,6 +30,8 @@ public class IngestionProperties {
     private List<String> ingestIfNotExists;
     private IngestionMapping ingestionMapping;
     private Map<String, String> additionalProperties;
+    private DataFormat dataFormat;
+    private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
     /**
      * Creates an initialized {@code IngestionProperties} instance with a given {@code databaseName} and {@code tableName}.
@@ -174,7 +179,7 @@ public class IngestionProperties {
     Map<String, String> getIngestionProperties() throws IOException {
         Map<String, String> fullAdditionalProperties = new HashMap<>();
         if (!dropByTags.isEmpty() || !ingestByTags.isEmpty() || !additionalTags.isEmpty()) {
-            ArrayList<String> tags = new ArrayList<>();
+            List<String> tags = new ArrayList<>();
             if (!additionalTags.isEmpty()) {
                 tags.addAll(additionalTags);
             }
@@ -201,6 +206,10 @@ public class IngestionProperties {
         }
         fullAdditionalProperties.putAll(additionalProperties);
 
+        if (dataFormat != null) {
+            fullAdditionalProperties.put("format", dataFormat.name());
+        }
+
         String mappingReference = ingestionMapping.getIngestionMappingReference();
         if (StringUtils.isNotBlank(mappingReference)) {
             fullAdditionalProperties.put("ingestionMappingReference", mappingReference);
@@ -219,7 +228,7 @@ public class IngestionProperties {
     }
 
     public void setDataFormat(DataFormat dataFormat) {
-        additionalProperties.put("format", dataFormat.name());
+        this.dataFormat = dataFormat;
     }
 
     /**
@@ -228,21 +237,27 @@ public class IngestionProperties {
      * @param dataFormatName One of the string values in: {@link DataFormat DataFormat}
      */
     public void setDataFormat(String dataFormatName) {
-        String dataFormat = DataFormat.valueOf(dataFormatName.toLowerCase()).name();
-        if (!dataFormat.isEmpty()) {
-            additionalProperties.put("format", dataFormat);
+        try {
+            this.dataFormat = DataFormat.valueOf(dataFormatName.toLowerCase());
+        } catch (IllegalArgumentException ex) {
+            log.warn("IngestionProperties.setDataFormat(): Invalid dataFormatName of {}. Per the API's specification, DataFormat property value wasn't set.", dataFormatName);
         }
     }
 
-    public String getDataFormatStr() {
-        return additionalProperties.get("format");
+    /**
+     * Returns the DataFormat if it exists, and otherwise defaults to CSV
+     *
+     * @return The DataFormat if it exists, and otherwise defaults to CSV
+     */
+    public DataFormat getDataFormat() {
+        return dataFormat != null ? dataFormat : DataFormat.csv;
     }
 
     /**
      * Sets the predefined ingestion mapping name:
      *
      * @param mappingReference     The name of the mapping declared in the destination Kusto database, that
-     *                             describes the mapping between fields of a object and columns of a Kusto table.
+     *                             describes the mapping between fields of an object and columns of a Kusto table.
      * @param ingestionMappingKind The data format of the object to map.
      */
     public void setIngestionMapping(String mappingReference, IngestionMapping.IngestionMappingKind ingestionMappingKind) {
@@ -279,8 +294,6 @@ public class IngestionProperties {
         Ensure.stringIsNotBlank(databaseName, "databaseName");
         Ensure.stringIsNotBlank(tableName, "tableName");
         Ensure.argIsNotNull(reportMethod, "reportMethod");
-        // TODO: Investigate whether we want dataFormat to be required (including checking other SDKs), and if so, add it to the next major release (as it's a breaking change)
-        //Ensure.argIsNotNull(getDataFormatStr(), "dataFormat");
 
         if (ingestionMapping.getColumnMappings() != null) {
             Ensure.isFalse(StringUtils.isNotBlank(ingestionMapping.getIngestionMappingReference()), "Both mapping reference and column mappings were defined");
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
@@ -7,6 +7,7 @@ import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
 import com.fasterxml.jackson.annotation.PropertyAccessor;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.microsoft.azure.kusto.data.Ensure;
+import com.microsoft.azure.kusto.ingest.exceptions.IngestionClientException;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -290,17 +291,41 @@ public class IngestionProperties {
     /**
      * Validate the minimum non-empty values needed for data ingestion and mappings.
      */
-    void validate() {
+    void validate() throws IngestionClientException {
         Ensure.stringIsNotBlank(databaseName, "databaseName");
         Ensure.stringIsNotBlank(tableName, "tableName");
         Ensure.argIsNotNull(reportMethod, "reportMethod");
 
+        String mappingReference = ingestionMapping.getIngestionMappingReference();
+        IngestionMapping.IngestionMappingKind ingestionMappingKind = ingestionMapping.getIngestionMappingKind();
+        String message = null;
+
         if (ingestionMapping.getColumnMappings() != null) {
-            Ensure.isFalse(StringUtils.isNotBlank(ingestionMapping.getIngestionMappingReference()), "Both mapping reference and column mappings were defined");
-            IngestionMapping.IngestionMappingKind ingestionMappingKind = ingestionMapping.getIngestionMappingKind();
-            for (ColumnMapping column : ingestionMapping.getColumnMappings()) {
-                Ensure.isTrue(column.isValid(ingestionMappingKind), String.format("Column mapping '%s' is invalid", column.columnName));
+            if (StringUtils.isNotBlank(mappingReference)) {
+                message += String.format("Both mapping reference '%s' and column mappings were defined.", mappingReference);
             }
+            for (ColumnMapping column : ingestionMapping.getColumnMappings()) {
+                if (!column.isValid(ingestionMappingKind)) {
+                    message += String.format("Column mapping '%s' is invalid", column.columnName);
+                }
+            }
+        } else if (StringUtils.isBlank(mappingReference)) {
+            if (dataFormat != null && dataFormat.isMappingRequired()) {
+                message += String.format("Mapping must be specified for '%s' format.", dataFormat.name());
+            }
+
+            if (ingestionMappingKind != null && ingestionMappingKind != IngestionMapping.IngestionMappingKind.unknown) {
+                message += String.format("If IngestionMappingKind was defined ('%s'), mapping must be specified.", ingestionMappingKind);
+            }
+        }
+
+        if (dataFormat != null && ingestionMappingKind != null && ingestionMappingKind != IngestionMapping.IngestionMappingKind.unknown && !dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
+                message += String.format("Wrong ingestion mapping for format '%s'; found '%s' mapping kind.", dataFormat.name(), ingestionMappingKind.name());
+        }
+
+        if (message != null) {
+            log.error(message);
+            throw new IngestionClientException(message);
         }
     }
 
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
@@ -234,7 +234,7 @@ public class IngestionProperties {
         }
     }
 
-    public String getDataFormat() {
+    public String getDataFormatStr() {
         return additionalProperties.get("format");
     }
 
@@ -280,7 +280,7 @@ public class IngestionProperties {
         Ensure.stringIsNotBlank(tableName, "tableName");
         Ensure.argIsNotNull(reportMethod, "reportMethod");
         // TODO: Investigate whether we want dataFormat to be required (including checking other SDKs), and if so, add it to the next major release (as it's a breaking change)
-        //Ensure.argIsNotNull(getDataFormat(), "dataFormat");
+        //Ensure.argIsNotNull(getDataFormatStr(), "dataFormat");
 
         if (ingestionMapping.getColumnMappings() != null) {
             Ensure.isFalse(StringUtils.isNotBlank(ingestionMapping.getIngestionMappingReference()), "Both mapping reference and column mappings were defined");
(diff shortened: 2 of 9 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -292,30 +292,32 @@ void validate() {
...
+        csv(IngestionMapping.IngestionMappingKind.Csv, false, true),
+        tsv(IngestionMapping.IngestionMappingKind.Csv, false, true),
+        scsv(IngestionMapping.IngestionMappingKind.Csv, false, true),
+        sohsv(IngestionMapping.IngestionMappingKind.Csv, false, true),
+        psv(IngestionMapping.IngestionMappingKind.Csv, false, true),
+        txt(IngestionMapping.IngestionMappingKind.unknown, false, true),
+        tsve(IngestionMapping.IngestionMappingKind.Csv, false, true),
+        json(IngestionMapping.IngestionMappingKind.Json, true, true),
+        singlejson(IngestionMapping.IngestionMappingKind.Json, true, true),
+        multijson(IngestionMapping.IngestionMappingKind.Json, true, true),
```
yihezkel, line 304: This had been treated as **not** being a mappingRequiredFormat. But I believe that's a bug.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -292,30 +292,32 @@ void validate() {
...
+        avro(IngestionMapping.IngestionMappingKind.Avro, true, false),
```
yihezkel, line 305: This had been treated as being compressible. But I believe that's a bug.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -147,25 +147,17 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
+    private String getMappingReference(IngestionProperties ingestionProperties, IngestionProperties.DataFormat dataFormat) throws IngestionClientException {
         IngestionMapping ingestionMapping = ingestionProperties.getIngestionMapping();
         String mappingReference = ingestionMapping.getIngestionMappingReference();
         IngestionMapping.IngestionMappingKind ingestionMappingKind = ingestionMapping.getIngestionMappingKind();
-        if (IngestionMapping.mappingRequiredFormats.contains(format)) {
+        if (dataFormat.isMappingRequired()) {
             String message = null;
-            if (!format.equalsIgnoreCase(ingestionMappingKind.name())) {
-                message = String.format("Wrong ingestion mapping for format %s, found %s mapping kind.", format, ingestionMappingKind.name());
+            if (!dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
```
yihezkel, line 156: This was a bug, expecting identical DF/IMK (in reality they have a many-to-one relationship, via a mapping)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -234,7 +234,7 @@ public void setDataFormat(String dataFormatName) {
         }
     }
 
-    public String getDataFormat() {
+    public String getDataFormatStr() {
```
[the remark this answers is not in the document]
yihezkel: For sure, but this is a unique situation, where otherwise we're "lying" about the return type. In this case, since we have a data type called `DataFormat`, one would expect `getDataFormat()` to return an object of that type. My approach is that user of even a class API should never be surprised, which is why I added the "Str" to the end.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
```diff
@@ -146,21 +146,22 @@ public IngestionResult ingestFromFile(FileSourceInfo fileSourceInfo, IngestionPr
             String filePath = fileSourceInfo.getFilePath();
             Ensure.fileExists(filePath);
             CompressionType sourceCompressionType = AzureStorageClient.getCompression(filePath);
-            boolean shouldCompress = AzureStorageClient.shouldCompress(sourceCompressionType, ingestionProperties.getDataFormat());
+            IngestionProperties.DataFormat dataFormat = IngestionProperties.DataFormat.getDataFormatFromString(ingestionProperties.getDataFormatStr());
```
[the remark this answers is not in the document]
yihezkel: In the API, we get the DataFormat **as** a string. The only transformation we make is to then parse it.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
```diff
@@ -192,13 +193,14 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
             } else if (streamSourceInfo.getStream().available() <= 0) {
                 throw new IngestionClientException("The provided stream is empty.");
             }
-            boolean shouldCompress = AzureStorageClient.shouldCompress(streamSourceInfo.getCompressionType(), ingestionProperties.getDataFormat());
+            IngestionProperties.DataFormat dataFormat = IngestionProperties.DataFormat.getDataFormatFromString(ingestionProperties.getDataFormatStr());
```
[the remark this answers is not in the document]
yihezkel: GitHub comparison makes it look like these 2 times that we get the `DataFormat` as a `String` and convert to a `DataFormat` object are in the same method, but the 1st is in `ingestFromFile()`, and the 2nd is in `ingestFromStream()`.

Are you suggesting adding a method to combine those 2 steps:
`protected IngestionProperties.DataFormat extractDataFormatFromIngestionProperties(IngestionProperties ingestionProperties) {
    IngestionProperties.DataFormat.getDataFormatFromString(ingestionProperties.getDataFormatStr())
}

Or are you suggesting finding all commonality between these 2 methods (unrelated to my change) and unifying those parts?
There are only 2 lines that are currently shared between the 2. With a little effort we can also unify the calls to `shouldCompress()` and `genBlobName()`. I don't see much benefit to that effort.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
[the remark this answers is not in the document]
AsafMah: Ok, but we should note this is a breaking change (renaming a public method).
If we're ok with this, I think we should not store DataFormat as an str, but as a DataFormat like we do with the tags.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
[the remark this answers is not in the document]
AsafMah: I guess they're different enough, you can leave it as-is

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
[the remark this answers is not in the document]
yihezkel: Yeah I'm trying to get this in davka before the release I want to do this week to 3.0.0.

I'm not sure what you're referring to regarding tags. This field is part of `AdditionalProperties`, which is a Map<String, String>. `Tags` is a `List<String>`. I think it's intentional keeping it a String because we don't need the entire object to be serialized and sent over the wire.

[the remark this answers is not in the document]
AsafMah: I don't understand what do you mean.

Other properties, such as tags and ingestionmapping, have their interface as their actual type (list<string> and IngestionMapping), and when they're serialized we turn them into the proper strings.

DataFormat should follow the same logic - the interface will be with the type DataFormat, and when we serialize it we'll turn it into string.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -115,16 +115,16 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
         streamSourceInfo.validate();
         ingestionProperties.validate();
 
-        String format = getFormat(ingestionProperties);
-        String mappingReference = getMappingReference(ingestionProperties, format);
+        IngestionProperties.DataFormat dataFormat = IngestionProperties.DataFormat.getDataFormatFromString(ingestionProperties.getDataFormatStr());
+        String mappingReference = getMappingReference(ingestionProperties, dataFormat);
         try {
             InputStream stream = (streamSourceInfo.getCompressionType() != null) ? streamSourceInfo.getStream() : compressStream(streamSourceInfo.getStream(), streamSourceInfo.isLeaveOpen());
```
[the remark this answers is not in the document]
yihezkel: Not sure.
It does seem inconsistent between the clients to only look at the file extension here, and not the DataFormat.

Maybe @ArielYehezkely knows.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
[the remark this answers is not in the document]
yihezkel: The `IngestionMapping` object is all relevant because it requires sending all that data on the wire. There's nothing in that object the service doesn't need.

But everywhere else, we minimize what we send to only what's needed. For example, I see `private List<String> dropByTags`, `private List<String> ingestByTags` and `private List<String> additionalTags`. I don't see any unnecessary object info being sent over the wire.

The `DataFormat` is stored as a record in `private Map<String, String> additionalProperties`. That's just a String to String mapping. We **could** move the DataFormat to be in its own field, but I'm not seeing any additional info from its enum object we'd **want** to include.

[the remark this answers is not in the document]
AsafMah: Yes, I meant moving DataFormat into its own field rather than additionalProperties. 
The advantage we get is type safety.
We don't have to parse it whenever we use it, and it is less error prone.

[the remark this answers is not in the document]
yihezkel: That would be much more than a breaking change - we'd have to change it in every SDK, and on the service side!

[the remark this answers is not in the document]
AsafMah: No we won't? 
Instead of having it in additionalFields, just have a DataFormat typed field, and do 
fullAdditionalProperties.put("format", d.name());
Why would we need to change anything else?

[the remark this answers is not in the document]
yihezkel: So you're suggesting maintaining in `IngestionProperties` both the full enum (I still don't understand what from the enum would be useful to send other than the name) and the name, and sending both over the wire?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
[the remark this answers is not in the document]
yihezkel: I spoke to @ArielYehezkely , and he doesn't see any reason to not do this. So I added it.

Good catch, @AsafMah !

ingest/src/test/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientTest.java
```diff
@@ -312,7 +311,7 @@ void generateName() {
         CompressionType compressionTypeRes2 = AzureStorageClient.getCompression(avroLocalCompressFileName);
         holder.name = avroLocalFileName;
         String avroName = genName.apply(DataFormat.avro, compressionTypeRes);
-        assert(avroName.endsWith("avi.avro.avro.gz"));
+        assert(avroName.endsWith("avi.avro.avro"));
```
yihezkel, line 314: Avro was previously treated as compressible

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -57,4 +58,9 @@ protected String generateEndpointSuggestion(String existingSuggestedEndpointUri,
     protected abstract String retrieveServiceType() throws IngestionServiceException, IngestionClientException;
 
     protected abstract String emendEndpointUri(URIBuilder existingEndpoint);
+
+    // We don't support compression of Parquet and Orc files
```
AsafMah, line 62: Remove this comment (it's not accurate and doesn't add anything)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -325,6 +340,17 @@ void validate() {
         public boolean isMappingRequired() {
             return mappingRequired;
         }
+
+        public boolean isCompressible() {
+            return compressible;
+        }
+
+        public static IngestionProperties.DataFormat getDataFormatFromString(String ingestionPropertiesDataFormat) {
```
AsafMah, line 348: unused

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
```diff
@@ -226,19 +228,17 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
         }
     }
 
-    private long estimateFileRawSize(String filePath, IngestionProperties.DataFormat format) {
+    private long estimateFileRawSize(String filePath, boolean isCompressible) {
```
AsafMah, line 231: I don't think we have this logic in the other SDKs. We probably want it, right? I'll add it to the list

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
```diff
@@ -146,21 +146,22 @@ public IngestionResult ingestFromFile(FileSourceInfo fileSourceInfo, IngestionPr
...
+            IngestionProperties.DataFormat dataFormat = ingestionProperties.getDataFormat();
+            boolean shouldCompress = IngestClientBase.shouldCompress(sourceCompressionType, dataFormat);
 
             File file = new File(filePath);
             String blobName = genBlobName(
                     file.getName(),
                     ingestionProperties.getDatabaseName(),
                     ingestionProperties.getTableName(),
-                    ingestionProperties.getDataFormat(),
+                    dataFormat.name(), // Used to use an empty string if the DataFormat was empty. Now it can't be empty, with a default of CSV.
```
AsafMah, line 157: If we want to keep the old behaviour , there's nothing stopping us from returning null in getDataFormat() and sending null here. But I think it's better to default to csv instead of having an "empty" case (The comment reads to me as "this is a bad change, but we have to do it" , but I think it's a good change).

As a bonus, you can annotate `getDataFormat` with `@NotNull` and then remove the null checks in `shouldCompress` and `genBlobName`

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -147,25 +147,17 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
-        return format;
-    }
-
-    private String getMappingReference(IngestionProperties ingestionProperties, String format) throws IngestionClientException {
+    private String getMappingReference(IngestionProperties ingestionProperties, IngestionProperties.DataFormat dataFormat) throws IngestionClientException {
         IngestionMapping ingestionMapping = ingestionProperties.getIngestionMapping();
         String mappingReference = ingestionMapping.getIngestionMappingReference();
         IngestionMapping.IngestionMappingKind ingestionMappingKind = ingestionMapping.getIngestionMappingKind();
-        if (IngestionMapping.mappingRequiredFormats.contains(format)) {
+        if (dataFormat.isMappingRequired()) {
```
AsafMah, line 154: Not introduced in this PR - but are we only checking this in streaming ingestion? shouldn't this check also be in queued?

AsafMah (changes requested):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: I intentionally left it in as it's a public method, and even though we're no longer using it, I see no reason to remove it.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
yihezkel: I assume if we did a full client comparison we'd find several gaps just because someone thought of 1 at some point after others were implemented.

As for this specific point, I agree that it would only make sense to include it in each client.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
yihezkel: 'The comment reads to me as "this is a bad change, but we have to do it" , but I think it's a good change'
That wasn't my intention - I was just pointing out context for troubleshooting in case it causes unexpected bugs at some point in the future.

I see `shouldCompress()` as more resilient if it can handle a null `DataFormat`, even if it's not currently called that way in our code.

`@NotNull` surprisingly doesn't actually enforce that a parameter isn't null.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -147,34 +146,6 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
-        return format;
-    }
-
-    private String getMappingReference(IngestionProperties ingestionProperties, String format) throws IngestionClientException {
-        IngestionMapping ingestionMapping = ingestionProperties.getIngestionMapping();
-        String mappingReference = ingestionMapping.getIngestionMappingReference();
-        IngestionMapping.IngestionMappingKind ingestionMappingKind = ingestionMapping.getIngestionMappingKind();
-        if (IngestionMapping.mappingRequiredFormats.contains(format)) {
-            String message = null;
-            if (!format.equalsIgnoreCase(ingestionMappingKind.name())) {
```
yihezkel, line 164: This is not identical to the checks in `IngestionProperties.validate()`. There, `IngestionMappingKind` can be `null` or `unknown` and not fail this. The question is what is correct - both in all cases, or whether `StreamingClient` is a special case?

@ArielYehezkely ?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -147,34 +146,6 @@ public IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, Inges
...
-                message = String.format("Wrong ingestion mapping for format %s, found %s mapping kind.", format, ingestionMappingKind.name());
-            }
-            if (StringUtils.isBlank(mappingReference)) {
-                message = String.format("Mapping reference must be specified for %s format.", format);
```
yihezkel, line 168: This is not identical to the checks in `IngestionProperties.validate()`. There, `columnMappings` can be provided instead of a `mappingReference` and not fail this. The question is what is correct - both in all cases, and whether `StreamingClient` is a special case?

@ArielYehezkely ?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
AsafMah: I guess it's hard to argue against more defensive code.
I think you should add `@NotNull` anyway, just to make sure we don't accidentally change the contract

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
AsafMah: In python we check it in the ctor of ingestionproperties so it's all of the clients:
```python
        if ingestion_mapping is not None and ingestion_mapping_reference is not None:
            raise KustoMappingAndMappingReferenceError()

        if data_format.mapping_required and ingestion_mapping_reference is None and ingestion_mapping is None:
            raise KustoMissingMappingReferenceError(data_format.kusto_value)

```

At first I thought we should do it here, but it's actually a subtle bug I noticed now, because if you have streaming ingestion and you provide it with column mapping, the check will pass but streaming ingestion only supports reference.
Maybe something to add to the other PR?

Make sure that if you refactor it here you don't have the same pitfall

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
AsafMah: I wrote on the other comment before I saw this. StreamingIngestClient can only have mappingReference, so this check is correct.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
AsafMah: How would unknown fail this?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
yihezkel: I just don't understand what that would achieve.

The javadoc for the method already specifies that it doesn't return null:
image

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
yihezkel: I vaguely remember that StreamingIngest requires a mapping reference only, so if that's true, I'll add that restriction to `StreamingIngestClient.ingestFromStream()` now:
`Ensure.stringIsNotBlank(ingestionProperties.getIngestionMapping().getIngestionMappingReference(), "ingestionMappingReference");`

Might as well get it all working as expected before a major bump.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
yihezkel: See in the `IngestionProperties.validate()` method:
`if (dataFormat != null && ingestionMappingKind != null && ingestionMappingKind != IngestionMapping.IngestionMappingKind.unknown && !dataFormat.getIngestionMappingKind().equals(ingestionMappingKind))`

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
AsafMah: It requires a mapping reference only for types that require a mapping.
You don't have to pass it for csv for example.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
AsafMah: You'd get better warnings, it solidifies the contract.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
AsafMah: You added this line in this pr :)

But still my point stands - unknown wouldn't fail either

yihezkel: Right, my point is that they're not identical, so I'm trying to identify which is the correct approach.

AsafMah: I think unknown shouldn't exist, or should always be an error when the user passes it.

I remember hearing that txt and raw can have csv mapping, and if not they shouldn't have mapping at all.

I think mapping kind should always be provided, so null should fail it too.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
yihezkel: I think they might only see a warning if they're using an IDE with (certain types of?) lint checking.

But why do we even want to add a (not useful, IMHO) `@NotNull` annotation to a method that explicitly handles `NULL`?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
yihezkel: Yeah I just figured that out and modified the code accordingly.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
AsafMah: @NotNull is on the return type, we guarantee that the method doesn't return null

yihezkel: Then I don't think that would impact any warnings in code using this method - I don't recall adding `@NotNull` to the return type doing that in my experience. I think it could only produce a warning if someone had this method return null (assuming they're using an IDE with (certain types of?) lint checking), but that's already clear in the javadoc that it doesn't. I think the annotation on its own (without explicit code to enforce) is only to document.

I think the question here boils down to how we want to manage API contract documentation -
1) Javadoc
2) Annotations, like `@NotNull`
3) Both
4) Something else

I'm flexible on this, but my thoughts are:
- We should be using javadoc no matter what, as that's the standard for documenting APIs.
- I prefer one source of truth, meaning I prefer not using annotations for documentation.
- I see these annotations as less beneficial for return types as for input parameters, since it helps the **user** of the method.
- I do indeed see a benefit to these annotations, and have incorporated them into several (most?) projects I've built. But that's been when I both (a) use a framework that enforces them as well (like IIRC Spring provides) and not just a coder-side IDE lint check, and (b) do it consistently throughout the codebase.

Actually, I'm looking at the code now to see whether we use it, and in fact we do about 40 times - and it looks like I wrote most of them! Mostly on input parameters, but a few on the return type. So I'm a bit embarrassed :-) but it shows that I'm very much into it, though inconsistently - meaning it's good to have this discussion of how we want to manage API contract documentation.

AsafMah: I feel the extra security from annotations is worth it, even if it's just a lint check.
I feel like the annotations are part of the function signature, and they are the "stronger" source of truths, like types.

But to be fair, @NotNull is much less valuable the @CanBeNull, since the ladder actually stops NREs.
So I trust you here, you can decide if you want to annotate or not.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
yihezkel: Nevermind, I aligned this check to how it originally was. I was missing a `&&  dataFormat.isMappingRequired()`, though they are effectively identical as we already have a check a few lines up whether a mapping is defined if mappings are required.
The code just makes a bit more sense now :-)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
yihezkel: Already addressed the fact that Streaming is a unique case, - if mapping is required, it must be a reference.

AsafMah (approved):

PR #204, opened 2021-12-23
Offer typed parameterized query helpers

yihezkel:
#### Pull Request Description

Offer typed parameterized query helpers

---

#### Future Release Comment
**Features:**
- Offer typed parameterized query helpers

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
@@ -27,7 +27,7 @@ import java.util.regex.Pattern;
 public class ClientRequestProperties {
     public static final String OPTION_SERVER_TIMEOUT = "servertimeout";
     public static final String OPTION_CLIENT_REQUEST_ID = "ClientRequestId";
-    public static final Pattern PATTERN =
+    public static final Pattern KUSTO_TIME_REGEX =
             Pattern.compile("(-?)(?:(\\d+)(\\.))?(?:([0-2]?\\d)(:))?([0-5]?\\d)(:)([0-5]?\\d)(?:(\\.)(\\d+))?",
                     Pattern.CASE_INSENSITIVE);
     private static final String OPTIONS_KEY = "Options";
@@ -147,7 +147,7 @@ public class ClientRequestProperties {
     }
 
     private long parseTimeoutFromTimespanString(String str) throws ParseException {
-        Matcher matcher = PATTERN.matcher(str);
+        Matcher matcher = KUSTO_TIME_REGEX.matcher(str);
         if (!matcher.matches()) {
             throw new ParseException(String.format("Failed to parse timeout string as a timespan. Value: '%s'", str));
         }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
@@ -3,16 +3,16 @@
 
 package com.microsoft.azure.kusto.data;
 
+import com.microsoft.azure.kusto.data.format.*;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.ParseException;
 import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.time.Duration;
+import java.time.LocalDateTime;
 import java.time.LocalTime;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -25,15 +25,15 @@ import java.util.regex.Pattern;
  * check out https://docs.microsoft.com/en-us/azure/kusto/api/netfx/request-properties#list-of-clientrequestproperties
  */
 public class ClientRequestProperties {
-    private static final String OPTIONS_KEY = "Options";
-    private static final String PARAMETERS_KEY = "Parameters";
     public static final String OPTION_SERVER_TIMEOUT = "servertimeout";
     public static final String OPTION_CLIENT_REQUEST_ID = "ClientRequestId";
+    public static final Pattern PATTERN =
+            Pattern.compile("(?:(\\d+)(\\.))?(?:([0-2]?\\d)(:))?(?:([0-5]?\\d))(:)(?:([0-5]?\\d))(?:(\\.)(\\d+))?",
+                    Pattern.CASE_INSENSITIVE);
+    private static final String OPTIONS_KEY = "Options";
+    private static final String PARAMETERS_KEY = "Parameters";
     private final Map<String, Object> parameters;
     private final Map<String, Object> options;
-    private static final Pattern PATTERN =
-            Pattern.compile("(?:(\\d+)\\.)?((?:[0-2]?\\d:)?(?:[0-5]?\\d):(?:[0-5]?\\d)(?:\\.\\d+)?)",
-                    Pattern.CASE_INSENSITIVE);
     static final long MAX_TIMEOUT_MS = TimeUnit.HOURS.toSeconds(1) * 1000;
 
     public ClientRequestProperties() {
@@ -59,6 +59,65 @@ public class ClientRequestProperties {
 
     public void setParameter(String name, Object value) {
         parameters.put(name, value);
+    }
+
+    public void setParameter(String name, String value) {
+        Ensure.stringIsNotBlank(name, "name");
+        Ensure.argIsNotNull(value, "value");
+
+        parameters.put(name, value);
+    }
+
+    public void setParameter(String name, Date value) {
+        Ensure.stringIsNotBlank(name, "name");
+        Ensure.argIsNotNull(value, "value");
+
+        parameters.put(name, new CslDateTimeFormat(value).toString());
+    }
+
+    public void setParameter(String name, LocalDateTime value) {
+        Ensure.stringIsNotBlank(name, "name");
+        Ensure.argIsNotNull(value, "value");
+
+        parameters.put(name, new CslDateTimeFormat(value).toString());
+    }
+
+    public void setParameter(String name, Duration value) {
+        Ensure.stringIsNotBlank(name, "name");
+        Ensure.argIsNotNull(value, "value");
+
+        parameters.put(name, new CslTimeFormat(value).toString());
+    }
+
+    public void setParameter(String name, boolean value) {
+        Ensure.stringIsNotBlank(name, "name");
+
+        parameters.put(name, new CslBoolFormat(value).toString());
+    }
+
+    public void setParameter(String name, int value) {
+        Ensure.stringIsNotBlank(name, "name");
+
+        parameters.put(name, new CslIntFormat(value).toString());
+    }
+
+    public void setParameter(String name, long value) {
+        Ensure.stringIsNotBlank(name, "name");
+
+        parameters.put(name, new CslLongFormat(value).toString());
+    }
+
+    public void setParameter(String name, double value) {
+        Ensure.stringIsNotBlank(name, "name");
+
+        parameters.put(name, new CslRealFormat(value).toString());
+    }
+
+    public void setParameter(String name, UUID value) {
+        Ensure.stringIsNotBlank(name, "name");
+        Ensure.argIsNotNull(value, "value");
+
+        parameters.put(name, new CslUuidFormat(value).toString());
     }
 
     public Object getParameter(String name) {
@@ -99,7 +158,13 @@ public class ClientRequestProperties {
             return MAX_TIMEOUT_MS;
         }
 
-        millis += TimeUnit.NANOSECONDS.toMillis(LocalTime.parse(matcher.group(2)).toNanoOfDay());
+        String timespanWithoutDays = "";
+        for (int i = 3; i <= 9; i++) {
+            if (matcher.group(i) != null) {
+                timespanWithoutDays += matcher.group(i);
+            }
+        }
+        millis += TimeUnit.NANOSECONDS.toMillis(LocalTime.parse(timespanWithoutDays).toNanoOfDay());
         return millis;
     }
 
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
@@ -3,7 +3,13 @@
 
 package com.microsoft.azure.kusto.data;
 
-import com.microsoft.azure.kusto.data.format.*;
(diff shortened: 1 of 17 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/format/CslFormat.java
```diff
@@ -0,0 +1,22 @@
+package com.microsoft.azure.kusto.data.format;
+
+public abstract class CslFormat {
```
yihezkel, line 3: I considered using generics here so we can have 1 `T value` property, but didn't since we need specialized behavior for the subclasses.

yihezkel:
I spoke with @ArielYehezkely and Amit Ofir, and we agreed that this use case is precisely something an SDK should offer - provide the user helper methods to make complex service formatting requirements transparent.

But I want to discuss this change with @AsafMah in terms of SDK parity.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -25,15 +25,15 @@
  * check out https://docs.microsoft.com/en-us/azure/kusto/api/netfx/request-properties#list-of-clientrequestproperties
  */
 public class ClientRequestProperties {
-    private static final String OPTIONS_KEY = "Options";
-    private static final String PARAMETERS_KEY = "Parameters";
     public static final String OPTION_SERVER_TIMEOUT = "servertimeout";
     public static final String OPTION_CLIENT_REQUEST_ID = "ClientRequestId";
+    public static final Pattern KUSTO_TIME_REGEX =
+            Pattern.compile("(-?)(?:(\\d+)(\\.))?(?:([0-2]?\\d)(:))?([0-5]?\\d)(:)([0-5]?\\d)(?:(\\.)(\\d+))?",
```
yogilad, line 31: TIMESPAN

data/src/main/java/com/microsoft/azure/kusto/data/format/CslBoolFormat.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public Boolean getValue() {
+        return value;
+    }
+
+    @Override
+    String getValueAsString() {
+        Ensure.argIsNotNull(value, "value");
+
+        return Boolean.TRUE.equals(value) ? "true" : "false";
```
yogilad, line 26: Why not simply `return value ? "true" : "false";`

data/src/main/java/com/microsoft/azure/kusto/data/format/CslDateTimeFormat.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.util.Date;
+
+public class CslDateTimeFormat extends CslFormat {
+    public static final String KUSTO_DATETIME_PATTERN = "yyyy-MM-dd'T'HH:mm:ss.SSSSSSS'Z'";
```
yogilad, line 13: IIRC kusto can process partial datetime strings, e.g. this is valid query
`KustoLogs
| where Timestamp  between (datetime("2021") .. 20m)`

data/src/main/java/com/microsoft/azure/kusto/data/format/CslUuidFormat.java
```diff
@@ -0,0 +1,30 @@
+package com.microsoft.azure.kusto.data.format;
+
+import com.microsoft.azure.kusto.data.Ensure;
+
+import java.util.UUID;
+
+public class CslUuidFormat extends CslFormat {
```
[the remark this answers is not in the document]
yihezkel: On the one hand, it seems Java has no defined convention for this scenario.
https://stackoverflow.com/questions/2236807/java-naming-convention-with-acronyms brings up a few ideas, including:
- "For 4+ letter acronyms, use mixed case. The standard library does this, and it just makes good sense."
- Several "canonical" Java classnames, showing there's clearly no consistent approach.

On the other hand we import (literally 2 lines above) a class with the acronym-in-all-caps convention :-)
I would argue it makes sense there because it's **only** the acronym.

WDYT?

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -61,6 +61,65 @@ public void setParameter(String name, Object value) {
...
+        Ensure.argIsNotNull(value, "value");
+
+        parameters.put(name, new CslDateTimeFormat(value).toString());
+    }
+
+    public void setParameter(String name, LocalDateTime value) {
+        Ensure.stringIsNotBlank(name, "name");
+        Ensure.argIsNotNull(value, "value");
+
+        parameters.put(name, new CslDateTimeFormat(value).toString());
```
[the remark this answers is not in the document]
yihezkel: I'm leaning against it - I see this like setting a value in a `HashMap`, where the user is trusted to manage writing each key.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: I'm naming these for the Kusto datatypes they're relevant to, and this is for the datatype called, "time".

data/src/main/java/com/microsoft/azure/kusto/data/format/CslBoolFormat.java
yihezkel: Because `Boolean` can be null, in which case testing it will throw a NPE.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: Oops, I just checked out https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/scalar-data-types/ and see that the datatype's name is indeed "timespan". Weird, because I've only ever seen it as "time(...)".

I'll update the terminology accordingly.

data/src/main/java/com/microsoft/azure/kusto/data/format/CslDateTimeFormat.java
yihezkel: You're right, the Csl parser in C# is much more mature and can handle things like that. I'd suggest that achieving full feature parity with C# is out of scope in this PR, but I tried to reasonably mimic the C# DateTime behavior.

data/src/main/java/com/microsoft/azure/kusto/data/format/CslFormat.java
```diff
@@ -0,0 +1,38 @@
...
+
+    // For example, parses "int(7)" as "7"
+    // If type wasn't found in valueWithType, returns it unchanged
+    public static String parseValueFromValueWithType(String valueWithType, String type) {
+        String valueWithTypeRegex = String.format("%s\\s*\\(\\s*(.*\\S+)\\s*\\)\\s*", type);
+        Pattern valueWithTypePattern = Pattern.compile(valueWithTypeRegex, Pattern.CASE_INSENSITIVE);
+        Matcher matcher = valueWithTypePattern.matcher(valueWithType);
+        if (matcher.matches()) {
+            return matcher.group(1);
+        } else {
```
yihezkel, line 34: I know this is unpopular in the "elegant" circles, but I personally consider this to be more readable.

AsafMah: Do you mean the "no new line before else" part? 
Because that's fine in java

data/src/main/java/com/microsoft/azure/kusto/data/format/CslFormat.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public String toString() {
+        return getType() + "(" + getValueOrNullAsString() + ")";
+    }
+
+    // For example, parses "int(7)" as "7"
+    // If type wasn't found in valueWithType, returns it unchanged
+    public static String parseValueFromValueWithType(String valueWithType, String type) {
+        String valueWithTypeRegex = String.format("%s\\s*\\(\\s*(.*\\S+)\\s*\\)\\s*", type);
```
AsafMah, line 29: I'm pretty sure the ".*" is not needed

yihezkel: We need it in case there's a whitespace character in the matching group.

data/src/main/java/com/microsoft/azure/kusto/data/format/CslFormat.java
yihezkel: I think the popular eleagent approach is:
`if (matcher.matches()) {
return matcher.group(1);
}
return valueWithType;`

AsafMah: oh, then I agree with the elegant approach. (with new lines)

data/src/main/java/com/microsoft/azure/kusto/data/format/CslFormat.java
AsafMah: I tried it in regex101, it's the same results with or without it.
Do you want to capture the whitespace inside the parens or not?

data/src/main/java/com/microsoft/azure/kusto/data/format/CslFormat.java
yihezkel: I'm not a fan of its readability.

data/src/main/java/com/microsoft/azure/kusto/data/format/CslFormat.java
yihezkel: I haven't tried a tester, but I indeed want to capture any whitespace that's part of the value inside the parens.

AsafMah: That's part of the value = includes leading and trailing?
Those are your options (unescaped regexes):

```regex
int\s*\((.*)\)\s* 

int(aaaa) => "aaaa"
int( aaaa  )  => " aaaa  "
int(a a) => "a a"
int( a a ) => " a a "

int\s*\(\s*(.*?)\s*\)\s*

int(aaaa) => "aaaa"
int( aaaa  )  => "aaaa"
int(a a) => "a a"
int( a a ) => "a a"

```

yihezkel: It doesn't make much sense for `int`, but consider others, like `datetime(2022-01-05 9:59:32)`

AsafMah: Ok I see what you mean, but I still prefer `\s*\(\s*(.*?)\s*\)\s*` (it's equivalent)

yihezkel: Are you sure they're equivalent? I think mine enforces that there's some non-whitespace inside the capture group, and I don't think yours does.

AsafMah: Ok yes if that's something we want we can keep it as is

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -3,16 +3,16 @@
...
 import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.time.Duration;
+import java.time.LocalDateTime;
 import java.time.LocalTime;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
+import java.util.*;
```
AsafMah, line 15: unstar

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -25,15 +25,15 @@
  * check out https://docs.microsoft.com/en-us/azure/kusto/api/netfx/request-properties#list-of-clientrequestproperties
  */
 public class ClientRequestProperties {
-    private static final String OPTIONS_KEY = "Options";
-    private static final String PARAMETERS_KEY = "Parameters";
     public static final String OPTION_SERVER_TIMEOUT = "servertimeout";
     public static final String OPTION_CLIENT_REQUEST_ID = "ClientRequestId";
+    public static final Pattern KUSTO_TIMESPAN_REGEX =
+            Pattern.compile("(-?)(?:(\\d+)(\\.))?(?:([0-2]?\\d)(:))?([0-5]?\\d)(:)([0-5]?\\d)(?:(\\.)(\\d+))?",
```
AsafMah, line 31: I feel like a regex this big deserves a comment with an explanation and example (or a link if you took it from somewhere)

AsafMah: Also I would prefer "PATTERN" over "REGEX"

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -297,7 +298,7 @@ public byte getByte(String columnName) {
     }
 
     public short getShort(String columnName) {
-        return (short) getShort(findColumn(columnName));
+        return getShort(findColumn(columnName));
```
AsafMah, line 301: Why aren't the others also implemented this way?
```java
public int getInt(String columnName) {
        return getInt(findColumn(columnName));
        }
```

pom.xml
```diff
@@ -40,6 +40,7 @@
         <!-- Compile dependencies -->
         <slf4j.version>1.8.0-beta4</slf4j.version>
         <commons-lang3.version>3.11</commons-lang3.version>
+        <commons-text.version>1.9</commons-text.version>
```
AsafMah, line 43: For what are we using this?

yihezkel: As discussed, for unescaping strings

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
yihezkel: I looked briefly, and it looks like their call stacks are the same in practice, as `private Object get(String colName)` calls `public int findColumn(String columnName)`. It's just inconsistent how sometimes the `get[DataType]` method calls `findColumn` directly, and sometimes calls it via `get`.

AsafMah (approved):
Approved after offline discussion

data/src/main/java/com/microsoft/azure/kusto/data/format/CslFormat.java
yihezkel: Used ternary, as discussed

PR #205, opened 2021-12-26
Validate IngestionProperties properly

yihezkel:
#### Pull Request Description

Removed enum DataFormat.Unknown.
Ingestion will now fail validation if any of the following ingestion property conditions are met:
- Mapping isn't specified for a data format that requires one.
- IngestionMappingKind was defined, but a mapping wasn't defined.
- Mapping was provided for a data format that doesn't support mappings.
- Provided IngestionMappingKind doesn't match the data format's expected ingestion mapping format.
- Both mapping reference and column mappings were defined.
- Column mappings were defined, but IngestionMappingKind wasn't.
- Mapping reference was defined, but IngestionMappingKind wasn't.
- A provided column mapping is invalid for the provided IngestionMappingKind.

---

#### Future Release Comment

**Breaking Changes:**
- Removed enum DataFormat.Unknown
- Ingestion will now fail validation if any of the following ingestion property conditions are met:
-- Mapping isn't specified for a data format that requires one.
-- IngestionMappingKind was defined, but a mapping wasn't defined.
-- Mapping was provided for a data format that doesn't support mappings.
-- Provided IngestionMappingKind doesn't match the data format's expected ingestion mapping format.
-- Both mapping reference and column mappings were defined.
-- Column mappings were defined, but IngestionMappingKind wasn't.
-- Mapping reference was defined, but IngestionMappingKind wasn't.
-- A provided column mapping is invalid for the provided IngestionMappingKind.

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
@@ -312,34 +312,30 @@ public class IngestionProperties {
 
         if ((ingestionMapping.getColumnMappings() == null) && StringUtils.isBlank(mappingReference)) {
             if (dataFormat.isMappingRequired()) {
-                message += String.format("Mapping must be specified for '%s' format.", dataFormat.name());
+                message += String.format("Mapping must be specified for '%s' format.%n", dataFormat.name());
             }
 
             if (ingestionMappingKind != null) {
-                message += String.format("IngestionMappingKind was defined ('%s'), so a mapping must be defined as well.", ingestionMappingKind);
+                message += String.format("IngestionMappingKind was defined ('%s'), so a mapping must be defined as well.%n", ingestionMappingKind);
             }
         } else { // a mapping was provided
             if (dataFormat.getIngestionMappingKind() != null && !dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
-                message += String.format("Wrong ingestion mapping for format '%s'; mapping kind should be '%s', but was '%s'.",
+                message += String.format("Wrong ingestion mapping for format '%s'; mapping kind should be '%s', but was '%s'.%n",
                         dataFormat.name(), dataFormat.getIngestionMappingKind(), ingestionMappingKind != null ? ingestionMappingKind.name() : "null");
             }
 
             if (ingestionMapping.getColumnMappings() != null) {
                 if (StringUtils.isNotBlank(mappingReference)) {
-                    message += String.format("Both mapping reference '%s' and column mappings were defined.", mappingReference);
+                    message += String.format("Both mapping reference '%s' and column mappings were defined.%n", mappingReference);
                 }
 
-                if (ingestionMappingKind == null) {
-                    message += "Column mappings were provided, so the IngestionMappingKind must be provided as well.";
-                } else {
+                if (ingestionMappingKind != null) {
                     for (ColumnMapping column : ingestionMapping.getColumnMappings()) {
                         if (!column.isValid(ingestionMappingKind)) {
-                            message += String.format("Column mapping '%s' is invalid", column.columnName);
+                            message += String.format("Column mapping '%s' is invalid.%n", column.columnName);
                         }
                     }
                 }
-            } else if (ingestionMappingKind == null) {
-                message += "A mapping reference was provided, so the IngestionMappingKind must be provided as well.";
             }
         }
 
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
@@ -231,14 +231,17 @@ public class IngestionProperties {
     }
 
     /**
-     * Sets the data format. If null is passed, then it does not set it.
+     * Sets the data format.
      *
      * @param dataFormat One of the values in: {@link DataFormat DataFormat}
+     * @throws IllegalArgumentException  if null argument is passed
      */
     public void setDataFormat(@NotNull DataFormat dataFormat) {
-        if (dataFormat != null) {
-            this.dataFormat = dataFormat;
+        if (dataFormat == null) {
+            throw new IllegalArgumentException("DataFormat cannot be set to null");
         }
+
+        this.dataFormat = dataFormat;
     }
 
     /**
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
@@ -9,6 +9,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.microsoft.azure.kusto.data.Ensure;
 import com.microsoft.azure.kusto.ingest.exceptions.IngestionClientException;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.TextStringBuilder;
 import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -311,40 +312,41 @@ public class IngestionProperties {
 
         String mappingReference = ingestionMapping.getIngestionMappingReference();
         IngestionMapping.IngestionMappingKind ingestionMappingKind = ingestionMapping.getIngestionMappingKind();
-        String message = "";
+        TextStringBuilder message = new TextStringBuilder();
 
         if ((ingestionMapping.getColumnMappings() == null) && StringUtils.isBlank(mappingReference)) {
             if (dataFormat.isMappingRequired()) {
-                message += String.format("Mapping must be specified for '%s' format.%n", dataFormat.name());
+                message.appendln("Mapping must be specified for '%s' format.", dataFormat.name());
             }
 
             if (ingestionMappingKind != null) {
-                message += String.format("IngestionMappingKind was defined ('%s'), so a mapping must be defined as well.%n", ingestionMappingKind);
+                message.appendln("IngestionMappingKind was defined ('%s'), so a mapping must be defined as well.", ingestionMappingKind);
             }
         } else { // a mapping was provided
             if (dataFormat.getIngestionMappingKind() != null && !dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
-                message += String.format("Wrong ingestion mapping for format '%s'; mapping kind should be '%s', but was '%s'.%n",
+                message.appendln("Wrong ingestion mapping for format '%s'; mapping kind should be '%s', but was '%s'.",
                         dataFormat.name(), dataFormat.getIngestionMappingKind(), ingestionMappingKind != null ? ingestionMappingKind.name() : "null");
             }
 
             if (ingestionMapping.getColumnMappings() != null) {
                 if (StringUtils.isNotBlank(mappingReference)) {
-                    message += String.format("Both mapping reference '%s' and column mappings were defined.%n", mappingReference);
+                    message.appendln("Both mapping reference '%s' and column mappings were defined.", mappingReference);
                 }
 
                 if (ingestionMappingKind != null) {
                     for (ColumnMapping column : ingestionMapping.getColumnMappings()) {
                         if (!column.isValid(ingestionMappingKind)) {
-                            message += String.format("Column mapping '%s' is invalid.%n", column.columnName);
+                            message.appendln("Column mapping '%s' is invalid.", column.columnName);
                         }
                     }
                 }
             }
         }
 
-        if (!message.equals("")) {
-            log.error(message);
-            throw new IngestionClientException(message);
+        if (!message.isEmpty()) {
+            String messageStr = message.build();
+            log.error(messageStr);
+            throw new IngestionClientException(messageStr);
         }
     }
 
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
@@ -43,6 +43,7 @@ public class IngestionProperties {
      * <p>{@code reportMethod} : {@code IngestionReportMethod.Queue;}</p>
      * <p>{@code flushImmediately} : {@code false;}</p>
      * <p>{@code additionalProperties} : {@code new HashMap();}</p>
+     * <p>{@code dataFormat} : {@code DataFormat.csv;}</p>
      * </blockquote>
      *
      * @param databaseName the name of the database in the destination Kusto cluster.
@@ -60,10 +61,12 @@ public class IngestionProperties {
         this.ingestIfNotExists = new ArrayList<>();
         this.additionalTags = new ArrayList<>();
         this.ingestionMapping = new IngestionMapping();
+        this.dataFormat = DataFormat.csv;
     }
 
     /**
      * Copy constructor for {@code IngestionProperties}.
+     *
      * @param other the instance to copy from.
      */
     public IngestionProperties(IngestionProperties other) {
@@ -72,6 +75,7 @@ public class IngestionProperties {
         this.reportLevel = other.reportLevel;
         this.reportMethod = other.reportMethod;
         this.flushImmediately = other.flushImmediately;
+        this.dataFormat = other.getDataFormat();
         this.additionalProperties = new HashMap<>(other.additionalProperties);
         this.dropByTags = new ArrayList<>(other.dropByTags);
         this.ingestByTags = new ArrayList<>(other.ingestByTags);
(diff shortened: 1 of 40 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -303,27 +310,39 @@ void validate() throws IngestionClientException {
...
-        } else if (StringUtils.isBlank(mappingReference)) {
+        if ((ingestionMapping.getColumnMappings() == null) && StringUtils.isBlank(mappingReference)) {
             if (dataFormat.isMappingRequired()) {
                 message += String.format("Mapping must be specified for '%s' format.", dataFormat.name());
             }
 
-            if (ingestionMappingKind != null && ingestionMappingKind != IngestionMapping.IngestionMappingKind.unknown) {
-                message += String.format("If IngestionMappingKind was defined ('%s'), mapping must be specified.", ingestionMappingKind);
+            if (ingestionMappingKind != null) {
+                message += String.format("IngestionMappingKind was defined ('%s'), so a mapping must be defined as well.", ingestionMappingKind);
```
yihezkel, line 319: This rule is in the .NET SDK

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -303,27 +310,39 @@ void validate() throws IngestionClientException {
...
             }
-        }
+        } else { // a mapping was provided
+            if (dataFormat.getIngestionMappingKind() == null) {
+                message += String.format("Data format '%s' does not support mappings", dataFormat.name());
```
yihezkel, line 323: This effectively applies to `RAW` and `TXT`. There's an ongoing e-mail thread about whether this restriction is valid, but as far as I can see it is. Will wait to merge until that thread is concluded.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -303,27 +310,39 @@ void validate() throws IngestionClientException {
...
+            } else if (!dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
+                message += String.format("Wrong ingestion mapping for format '%s'; mapping kind should be '%s', but was '%s'.",
+                        dataFormat.name(), dataFormat.getIngestionMappingKind(), ingestionMappingKind != null ? ingestionMappingKind.name() : "null");
+            }
+
+            if (ingestionMapping.getColumnMappings() != null) {
+                if (StringUtils.isNotBlank(mappingReference)) {
+                    message += String.format("Both mapping reference '%s' and column mappings were defined.", mappingReference);
```
yihezkel, line 331: This rule is in the .NET SDK

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -303,27 +310,39 @@ void validate() throws IngestionClientException {
...
+                }
 
-        if (dataFormat.isMappingRequired() && !dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
-            message += String.format("Wrong ingestion mapping for format '%s'; found '%s' mapping kind.", dataFormat.name(), ingestionMappingKind != null ? ingestionMappingKind.name() : "null");
+                if (ingestionMappingKind == null) {
+                    message += "Column mappings were provided, so the IngestionMappingKind must be provided as well.";
```
yihezkel, line 335: In the Java SDK, setting either type of mapping requires also passing an `IngestionMappingKind`

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -303,27 +310,39 @@ void validate() throws IngestionClientException {
...
+                } else {
+                    for (ColumnMapping column : ingestionMapping.getColumnMappings()) {
+                        if (!column.isValid(ingestionMappingKind)) {
+                            message += String.format("Column mapping '%s' is invalid", column.columnName);
+                        }
+                    }
+                }
+            } else if (ingestionMappingKind == null) {
+                message += "A mapping reference was provided, so the IngestionMappingKind must be provided as well.";
```
yihezkel, line 344: In the Java SDK, setting either type of mapping requires also passing an `IngestionMappingKind`

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -301,40 +308,52 @@ void validate() throws IngestionClientException {
...
-            }
-            for (ColumnMapping column : ingestionMapping.getColumnMappings()) {
-                if (!column.isValid(ingestionMappingKind)) {
-                    message += String.format("Column mapping '%s' is invalid", column.columnName);
-                }
-            }
-        } else if (StringUtils.isBlank(mappingReference)) {
+        if ((ingestionMapping.getColumnMappings() == null) && StringUtils.isBlank(mappingReference)) {
             if (dataFormat.isMappingRequired()) {
                 message += String.format("Mapping must be specified for '%s' format.", dataFormat.name());
```
yihezkel, line 315: I don't think we should remove this until 3.0.1, just to be sure everything else is stable first. This would be a restriction relaxation, so that can be part of a patch version.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: I'm going to remove this restriction now.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: I'm debating whether to remove this now, or in 3.0.1. I think it's best to wait.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: I'm debating whether to remove this now, or in 3.0.1. I think it's best to wait.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -301,40 +308,50 @@ void validate() throws IngestionClientException {
...
             }
 
-            if (ingestionMappingKind != null && ingestionMappingKind != IngestionMapping.IngestionMappingKind.unknown) {
-                message += String.format("If IngestionMappingKind was defined ('%s'), mapping must be specified.", ingestionMappingKind);
+            if (ingestionMappingKind != null) {
+                message += String.format("IngestionMappingKind was defined ('%s'), so a mapping must be defined as well.", ingestionMappingKind);
             }
-        }
+        } else { // a mapping was provided
+            if (dataFormat.getIngestionMappingKind() != null && !dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
```
yihezkel, line 322: Can RAW/TXT be accompanied by **any** type of mapping? I assume not.

yihezkel: CSV

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -230,16 +230,23 @@ public void setIngestIfNotExists(List<String> ingestIfNotExists) {
...
 
-    public void setDataFormat(DataFormat dataFormat) {
-        this.dataFormat = dataFormat;
+    /**
+     * Sets the data format. If null is passed, then it does not set it.
+     *
+     * @param dataFormat One of the values in: {@link DataFormat DataFormat}
+     */
+    public void setDataFormat(@NotNull DataFormat dataFormat) {
+        if (dataFormat != null) {
```
AsafMah, line 239: @NotNull means that we expect to never get null as a parameter (usually throws if we got it). If we handle it then the correct annotation is @CanBeNull

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -301,40 +308,50 @@ void validate() throws IngestionClientException {
...
+                }
 
-        if (dataFormat.isMappingRequired() && !dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
-            message += String.format("Wrong ingestion mapping for format '%s'; found '%s' mapping kind.", dataFormat.name(), ingestionMappingKind != null ? ingestionMappingKind.name() : "null");
+                if (ingestionMappingKind == null) {
+                    message += "Column mappings were provided, so the IngestionMappingKind must be provided as well.";
+                } else {
+                    for (ColumnMapping column : ingestionMapping.getColumnMappings()) {
+                        if (!column.isValid(ingestionMappingKind)) {
+                            message += String.format("Column mapping '%s' is invalid", column.columnName);
```
AsafMah, line 337: Make message a StringBuilder (if you go to the += here idea will do it for you)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
AsafMah: This should be a comment in the code rather than on the PR

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -301,40 +308,50 @@ void validate() throws IngestionClientException {
...
-            }
-            for (ColumnMapping column : ingestionMapping.getColumnMappings()) {
-                if (!column.isValid(ingestionMappingKind)) {
-                    message += String.format("Column mapping '%s' is invalid", column.columnName);
-                }
-            }
-        } else if (StringUtils.isBlank(mappingReference)) {
+        if ((ingestionMapping.getColumnMappings() == null) && StringUtils.isBlank(mappingReference)) {
             if (dataFormat.isMappingRequired()) {
                 message += String.format("Mapping must be specified for '%s' format.", dataFormat.name());
```
AsafMah, line 315: Maybe there should be a newline between concatenated messages?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -301,40 +308,50 @@ void validate() throws IngestionClientException {
 
         String mappingReference = ingestionMapping.getIngestionMappingReference();
```
AsafMah, line 309: Not exactly related to this PR specfically, but have you seen my PR in node -https://github.com/Azure/azure-kusto-node/pull/154/files?
The column mappings didn't work correctly, so I took the effort to also make them safer by introducing a type for each type of mapping.
Do you think something like that will work here?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/StreamingIngestClientTest.java
```diff
@@ -456,7 +457,7 @@ void IngestFromFile_JsonWrongMappingKind_IngestionClientException() {
         IngestionClientException ingestionClientException = assertThrows(IngestionClientException.class,
                 () -> streamingIngestClient.ingestFromFile(fileSourceInfo, ingestionProperties),
                 "Expected IngestionClientException to be thrown, but it didn't");
-        assertTrue(ingestionClientException.getMessage().contains("Wrong ingestion mapping for format 'json'; found 'Csv' mapping kind."));
```
AsafMah, line 459: Add some more negative tests for the new rules

AsafMah (changes requested):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: We don't allow them to pass null here.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: I have a different approach to some lint smells, and this is one of them.

My raitonale is:
1) Here, it's the generation of an error message, which is not the happy flow and rare.
2) Further, it's limited to just the number of columns. The performance benefit of StringBuilder is negligible in this case, in which case I very much prefer the readability of `+=`.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: Good idea

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: I hadn't seen that PR. I assume you're referring to `ColumnMapping.isValid()` - what's the potential pitfall there?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: I generally try to avoid TODOs in the code. Instead (I think we discussed this over Teams as well), I already made this change locally, waiting to create a PR for it until after this is merged, which we'll do concurrently with a discussion with Tzvia.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: Nevermind, this check is new in this PR, so I'll just remove it now.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: Nevermind, this check is new in this PR, so I'll just remove it now.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/StreamingIngestClientTest.java
yihezkel: I just checked, and see that the 2 new rules are ones that we just discovered we don't need. So I removed them, and now there are 0 new validation rules.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel: More accurately, they can technically send null, but it's pointless.

AsafMah: I disagree, we allow them to pass null, since the method explicitly supports it. 
I think you should throw if it's null, having it do nothing is a surprising behaviour.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
AsafMah: I don't think += is more readable than append.
Especially if you're going to add new lines, you can use TextStringBuilder and `appendLn`

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
AsafMah: It's about rather than having one `ColumnMapping` type, having a type for each kind of mapping (E.g `JsonColumnMapping`) that each of them supports only the feature of that column mapping (stuff like transforms and paths), and have their kind initialized correctly.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/StreamingIngestClientTest.java
AsafMah: Did you push it?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -230,16 +231,26 @@ public void setIngestIfNotExists(List<String> ingestIfNotExists) {
...
 
-    public void setDataFormat(DataFormat dataFormat) {
+    /**
+     * Sets the data format.
+     *
+     * @param dataFormat One of the values in: {@link DataFormat DataFormat}
+     * @throws IllegalArgumentException  if null argument is passed
+     */
+    public void setDataFormat(@NotNull DataFormat dataFormat) {
+        if (dataFormat == null) {
```
AsafMah, line 241: Use Ensure.argisnotnull

AsafMah (approved):
Fix the ensure and you're ready to go

PR #207, opened 2022-01-06
Remove dependency on Java 9/a specific version of a specific project of Java 8

yihezkel:
#### Pull Request Description

Remove dependency on Java 9/a specific version of a specific project of Java 8. sun.misc.IOUtils is a part of OpenJDK, but isn't standard in Java. Further, the method used here was only added about 1 year ago (see https://hg.openjdk.java.net/jdk8u/monojdk8u/rev/8994a5a78287). Strangely, it was added to the Java 8 repo and not the Java 9 repo, but is annotated with `@since 1.9`!

---

#### Future Release Comment
Remove dependency on Java 9/a specific version of a specific project of Java 8. sun.misc.IOUtils is a part of OpenJDK, but isn't standard in Java. Further, the method used here was only added about 1 year ago (see https://hg.openjdk.java.net/jdk8u/monojdk8u/rev/8994a5a78287). Strangely, it was added to the Java 8 repo and not the Java 9 repo, but is annotated with `@since 1.9`!

**Fixes:**
- Remove dependency on Java 9/a specific version of a specific project of Java 8 (sun.misc.IOUtils)

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionUtils.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionUtils.java
@@ -5,7 +5,6 @@ import com.microsoft.azure.kusto.ingest.source.FileSourceInfo;
 import com.microsoft.azure.kusto.ingest.source.ResultSetSourceInfo;
 import com.microsoft.azure.kusto.ingest.source.StreamSourceInfo;
 import com.univocity.parsers.csv.CsvRoutines;
-
 import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -20,7 +19,11 @@ import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
 
 public class IngestionUtils {
-    private final static Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+    private IngestionUtils() {
+        // Hide the default constructor, since this is a utils class
+    }
+
+    private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
     @NotNull
     public static StreamSourceInfo fileToStream(FileSourceInfo fileSourceInfo, boolean resettable) throws IngestionClientException, FileNotFoundException {
@@ -51,5 +54,17 @@ public class IngestionUtils {
         }
         ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(byteArrayOutputStream.toByteArray());
         return new StreamSourceInfo(byteArrayInputStream, false, resultSetSourceInfo.getSourceId());
+    }
+
+    public static byte[] readBytesFromInputStream(InputStream value, int testByteArraySize) throws IOException {
+        ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+        int numBytesRead;
+        byte[] data = new byte[testByteArraySize];
+
+        while ((numBytesRead = value.read(data, 0, data.length)) != -1) {
+            buffer.write(data, 0, numBytesRead);
+        }
+
+        return buffer.toByteArray();
     }
 }
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionUtils.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionUtils.java
@@ -56,13 +56,16 @@ public class IngestionUtils {
         return new StreamSourceInfo(byteArrayInputStream, false, resultSetSourceInfo.getSourceId());
     }
 
-    public static byte[] readBytesFromInputStream(InputStream value, int testByteArraySize) throws IOException {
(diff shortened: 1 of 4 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionUtils.java
```diff
@@ -52,4 +55,16 @@ public static StreamSourceInfo resultSetToStream(ResultSetSourceInfo resultSetSo
         ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(byteArrayOutputStream.toByteArray());
         return new StreamSourceInfo(byteArrayInputStream, false, resultSetSourceInfo.getSourceId());
     }
+
+    public static byte[] readBytesFromInputStream(InputStream value, int testByteArraySize) throws IOException {
```
AsafMah, line 59: The semantics for this method are wrong.
This will always read the entire buffer to memory. Unless I'm missing something, `testByteArraySize` doesn't do anything except setting the size of the buffer for each read.

The semantics we want is to read a maximum of `testByteArraySize` (or a better name) , or less if the stream ends before.

Look at the python implementation:
```python
def read_until_size_or_end(stream: IO[AnyStr], size: int) -> io.BytesIO:
    pos = 0
    result = io.BytesIO()
    while True:
        try:
            returned = stream.read(size - pos)
            pos += len(returned)
            result.write(returned)

            if len(returned) == 0 or pos == size:
                result.seek(0, io.SEEK_SET)
                return result

        except BlockingIOError:
            continue
```





Now the reason it didn't fail the test, is because the SequenceStream we put it in is opaque, and we can't know which bytes come from the byte array and which come from the original stream.
So I can't think of a good way to test it

AsafMah (changes requested):
See code comment

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionUtils.java
AsafMah: That's the openjdk implementation BTW
```java
    public static byte[] readNBytes(InputStream is, int len) throws IOException {
        if (len < 0) {
            throw new IllegalArgumentException("len < 0");
        }

        List<byte[]> bufs = null;
        byte[] result = null;
        int total = 0;
        int remaining = len;
        int n;
        do {
            byte[] buf = new byte[Math.min(remaining, DEFAULT_BUFFER_SIZE)];
            int nread = 0;

            // read to EOF which may read more or less than buffer size
            while ((n = is.read(buf, nread,
                    Math.min(buf.length - nread, remaining))) > 0) {
                nread += n;
                remaining -= n;
            }

            if (nread > 0) {
                if (MAX_BUFFER_SIZE - total < nread) {
                    throw new OutOfMemoryError("Required array size too large");
                }
                total += nread;
                if (result == null) {
                    result = buf;
                } else {
                    if (bufs == null) {
                        bufs = new ArrayList<>();
                        bufs.add(result);
                    }
                    bufs.add(buf);
                }
            }
            // if the last call to read returned -1 or the number of bytes
            // requested have been read then break
        } while (n >= 0 && remaining > 0);

        if (bufs == null) {
            if (result == null) {
                return new byte[0];
            }
            return result.length == total ?
                result : Arrays.copyOf(result, total);
        }

        result = new byte[total];
        int offset = 0;
        remaining = total;
        for (byte[] b : bufs) {
            int count = Math.min(b.length, remaining);
            System.arraycopy(b, 0, result, offset, count);
            offset += count;
            remaining -= count;
        }

        return result;
    }
```

yihezkel: If your intent was to read exactly the specified number of bytes, why not use `java.io.InputStream.read(byte[] b, int off, int len)`?

yihezkel: Nevermind, I see `readNBytes` calls that, but does more, presumably per the requirements you were describing.

AsafMah: Specifically, read waits until any data is available and then returns whatever it got.
We need to keep reading until we're at the end of the stream or we passed the size.

yihezkel: OK I updated it to the semantics you describe, instead of reading everything.

AsafMah:
Works great.

But style choice - I don't like assignments inside of a while loop's condition. Let's discuss

yihezkel:
@AsafMah Let me know when you want to discuss. I think this is the standard Java way to read from an `InputStream` in a loop, because it's so convenient as a stopping condition (among others).

Here are the first 2 stackoverflow search results: https://stackoverflow.com/questions/309424/how-do-i-read-convert-an-inputstream-into-a-string-in-java and https://stackoverflow.com/questions/6160432/java-inputstream-reading-problem

AsafMah:
I know it's a common pattern, I'm not a big fan of it usually, but here unlike in the examples it's not the only thing in the while, which makes it less readable in my opinion

AsafMah (approved):

PR #208, opened 2022-01-11
Only make clients Closeable if they have resources that need to be closed

yihezkel:
#### Pull Request Description

Only make clients Closeable if they have resources that need to be closed

---

#### Future Release Comment

**Breaking Changes:**
- Only make clients Closeable if they have resources that need to be closed


AsafMah:
Now because the base IngestClient is not closeable, wherever you use
`IngestClient client = IngestClientFactory.createClient(csb);`

You no longer close the client, even though you should.
I rather have a no-op close for StreamingClient instead of having queued and managed be non-closeable unless you cast them.

yihezkel:
@AsafMah This was a stupid mistake on my part. I hastily made this PR based on a hunch without thinking it through - of course you're right that since we have a parent interface we hope they'll use for all clients (as appropriate), it should implement `Closeable`.

Abandoning (with my tail between my legs)

PR #210, opened 2022-01-11
Fix access modifiers

yihezkel:
#### Pull Request Description

Fix access modifiers in ColumnMapping, IngestionBlobInfo and IngestionStatusInTableDescription.

---

#### Future Release Comment
Fix access modifiers in ColumnMapping, IngestionBlobInfo and IngestionStatusInTableDescription.

**Breaking Changes:**
- Fix access modifiers in ColumnMapping, IngestionBlobInfo and IngestionStatusInTableDescription.

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
@@ -87,38 +87,39 @@ public class QueuedIngestClient extends IngestClientBase implements IngestClient
                     ingestionProperties.getDatabaseName(), ingestionProperties.getTableName());
             String urlWithoutSecrets = SecurityUtils.removeSecretsFromUrl(blobSourceInfo.getBlobPath());
             if (blobSourceInfo.getRawSizeInBytes() > 0L) {
-                ingestionBlobInfo.rawDataSize = blobSourceInfo.getRawSizeInBytes();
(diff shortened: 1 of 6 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClient.java
```diff
@@ -87,38 +87,39 @@ public IngestionResult ingestFromBlob(BlobSourceInfo blobSourceInfo, IngestionPr
                     ingestionProperties.getDatabaseName(), ingestionProperties.getTableName());
             String urlWithoutSecrets = SecurityUtils.removeSecretsFromUrl(blobSourceInfo.getBlobPath());
             if (blobSourceInfo.getRawSizeInBytes() > 0L) {
-                ingestionBlobInfo.rawDataSize = blobSourceInfo.getRawSizeInBytes();
+                ingestionBlobInfo.setRawDataSize(blobSourceInfo.getRawSizeInBytes());
```
AsafMah, line 90: I get it's "the java way", but aren't you the one that said ".append" is less readable than "+="? 
Then surely "set" is less readable than "="?

On the other hand, if there was a time to make this change it's now, let's discuss.

yihezkel: I am!

But here this is about OO fundamentals, which trump readability any day.

PR #213, opened 2022-01-23
Support exceptions that don't follow oneapi

AsafMah:
-If exception doesn't follow oneapi throw dataserviceexception
-Don't throw oneapi for wrong exception

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -7,8 +7,15 @@ import com.microsoft.azure.kusto.data.auth.CloudInfo;
 import com.microsoft.azure.kusto.data.exceptions.DataClientException;
 import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
 import com.microsoft.azure.kusto.data.exceptions.DataWebException;
+import com.microsoft.azure.kusto.data.exceptions.WebException;
+
 import org.apache.commons.lang3.StringUtils;
-import org.apache.http.*;
+import org.apache.http.Header;
+import org.apache.http.HttpEntity;
+import org.apache.http.HttpHeaders;
+import org.apache.http.HttpResponse;
+import org.apache.http.HttpStatus;
+import org.apache.http.StatusLine;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
@@ -22,13 +29,18 @@ import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.util.EntityUtils;
 import org.jetbrains.annotations.NotNull;
 import org.json.JSONException;
+import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
-import java.net.*;
+import java.net.MalformedURLException;
+import java.net.SocketTimeoutException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -162,11 +162,13 @@ private static DataServiceException createExceptionFromResponse(String url, Http
                     return new DataServiceException(url, message, formattedException,
                             formattedException.getApiError().isPermanent());
                 } catch (Exception ignored) {
+                    // Exception doesn't follow the OneAPI error format
+                    return new DataServiceException(url, errorFromResponse, false);
```
ArielYehezkely, line 166: 1. Usually it's not a healthy pattern to communicate between code parts using exceptions. i.e. we assume this is an API error then we try and fail - can we avoid this?
2. Can we try get the actual message instead of printing {"message": "..."}

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -162,11 +162,13 @@ private static DataServiceException createExceptionFromResponse(String url, Http
...
                 }
             }
             errorFromResponse = String.format("Http StatusCode='%s', ActivityId='%s'", httpResponse.getStatusLine().toString(), activityId);
             if (thrownException == null) {
-                thrownException = new DataWebException(errorFromResponse, httpResponse);
+                return new DataServiceException(url, errorFromResponse, false);
```
ArielYehezkely, line 171: Don't we miss the Http code here?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
AsafMah: 1. I changed the code. Now I parse the json and see from there, but we do still have this problem because you can't know if the json is valid until you parse it. I added a comment to explain.
2. Done

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
AsafMah: I changed the hierarchy (in a non-breaking way), now there is a `WebException` class for exceptions that aren't oneapi, and there's an api from DataServiceException to get the error code, if available.

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataWebException.java
```diff
@@ -6,41 +6,24 @@
...
+public class DataWebException extends WebException {
 
-    private final HttpResponse httpResponse;
-    private OneApiError apiError;
+    private final OneApiError apiError;
 
-    public HttpResponse getHttpResponse() { return httpResponse; }
+    public DataWebException(String message, HttpResponse httpResponse, Throwable cause) {
+        super(message, httpResponse, cause);
+        this.apiError = OneApiError.fromJsonObject(new JSONObject(getMessage()).getJSONObject("error"));
```
ArielYehezkely, line 15: This can throw - better to keep it in the method then throw from the c'tor

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -154,22 +166,35 @@ private static DataServiceException createExceptionFromResponse(String url, Http
...
-                String message;
-                DataWebException formattedException = new DataWebException(errorFromResponse, httpResponse);
                 try {
-                    message = String.format("%s, ActivityId='%s'", formattedException.getApiError().getDescription(), activityId);
-                    return new DataServiceException(url, message, formattedException,
-                            formattedException.getApiError().isPermanent());
-                } catch (Exception ignored) {
+                    JSONObject jsonObject = new JSONObject(errorFromResponse);
+                    if (jsonObject.has("error")) {
+                        formattedException = new DataWebException(errorFromResponse, httpResponse, thrownException);
```
ArielYehezkely, line 176: It's the same line for all cases

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -154,22 +166,35 @@ private static DataServiceException createExceptionFromResponse(String url, Http
...
+                        message = String.format("%s, ActivityId='%s'", ((DataWebException)formattedException).getApiError().getDescription(), activityId);
+                        isPermanent = ((DataWebException)formattedException).getApiError().isPermanent();
+                    } else if (jsonObject.has("message")) {
+                        formattedException = new WebException(errorFromResponse, httpResponse, thrownException);
+                        message = String.format("%s, ActivityId='%s'", jsonObject.getString("message"), activityId);
+                    } else {
+                        formattedException = new WebException(errorFromResponse, httpResponse, thrownException);
+                        message = String.format("%s, ActivityId='%s'", errorFromResponse, activityId);
```
ArielYehezkely, line 184: Same as the catch, you can reduce cases

ArielYehezkely (approved):

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/DataWebException.java
AsafMah: Done

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
AsafMah: I reduced a lot of the logic, let me know if that's ok.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
AsafMah: I reduced a lot of the logic, let me know if that's ok.

ArielYehezkely (approved):

ArielYehezkely (approved):

PR #214, opened 2022-01-30
Remove mapping requirement, since mappings are no longer required

yihezkel:
#### Pull Request Description

Remove mapping requirement, since mappings are no longer required

---

#### Future Release Comment
**Features:**
- Remove mapping requirement, since mappings are no longer required

### DIFF
--- a/quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
+++ b/quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
@@ -360,30 +360,18 @@ public class KustoSampleApp {
     }
 
     private static boolean createIngestionMappings(boolean useExistingMapping, Client kustoClient, String databaseName, String tableName, String mappingName, String mappingValue, IngestionProperties.DataFormat dataFormat) {
-        if (!useExistingMapping) {
-            // TODO Deprecated: this restriction is likely to be removed soon
-            if ((dataFormat == IngestionProperties.DataFormat.JSON || dataFormat == IngestionProperties.DataFormat.MULTIJSON || dataFormat == IngestionProperties.DataFormat.SINGLEJSON || dataFormat == IngestionProperties.DataFormat.AVRO) && StringUtils.isBlank(mappingValue)) {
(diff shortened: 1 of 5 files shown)


AsafMah:
Do we want to add a test specifically for ingesting json without a mapping?

yihezkel:
@AsafMah I think it would be appropriate if we use tests as documentation. We don't, but it's a laudable goal, so I'll add :-)

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -360,30 +360,18 @@ private static void alterBatchingPolicy(Client kustoClient, String databaseName,
     }
 
     private static boolean createIngestionMappings(boolean useExistingMapping, Client kustoClient, String databaseName, String tableName, String mappingName, String mappingValue, IngestionProperties.DataFormat dataFormat) {
-        if (!useExistingMapping) {
-            // TODO Deprecated: this restriction is likely to be removed soon
-            if ((dataFormat == IngestionProperties.DataFormat.JSON || dataFormat == IngestionProperties.DataFormat.MULTIJSON || dataFormat == IngestionProperties.DataFormat.SINGLEJSON || dataFormat == IngestionProperties.DataFormat.AVRO) && StringUtils.isBlank(mappingValue)) {
-                System.out.printf("The data format '%s' requires a mapping, but configuration indicates to not use an existing mapping and no mapping was provided. Skipping this ingestion.%n", dataFormat.getKustoValue());
-                return false;
-            }
+        if (!useExistingMapping && StringUtils.isNotBlank(mappingValue)) {
```
AsafMah, line 363: Nit: you can invert the if and do an early return

AsafMah (approved):
Looks good

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
yihezkel: My thought process was that this structure is more intuitive to understand the intended logic.
Also, it effectively returns early because the implicit `else` immediately returns `true`.

PR #215, opened 2022-02-06
Support ingesting with ignoring first record in both SDK and Quickstart

yihezkel:
#### Pull Request Description

Support ingesting with ignoring first record in both SDK and Quickstart

---

#### Future Release Comment
**Features:**
- Support ingesting with ignoring first record in both SDK and Quickstart

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -9,7 +9,6 @@ import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
 import com.microsoft.azure.kusto.data.exceptions.DataWebException;
 import com.microsoft.azure.kusto.data.exceptions.OneApiError;
 import com.microsoft.azure.kusto.data.exceptions.WebException;
-
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
@@ -57,15 +56,14 @@ class Utils {
         // Hide constructor, as this is a static utility class
     }
 
-    static String post(CloseableHttpClient httpClient, String url, String payload, InputStream stream, long timeoutMs,
-      Map<String, String> headers, boolean leaveOpen) throws DataServiceException, DataClientException {
-        URI uri = parseUriFromUrlString(url);
+    static String post(CloseableHttpClient httpClient, String urlStr, String payload, InputStream stream, long timeoutMs, Map<String, String> headers, boolean leaveOpen) throws DataServiceException, DataClientException {
+        URI url = parseUriFromUrlString(urlStr);
 
         try (InputStream ignored = (stream != null && !leaveOpen) ? stream : null) {
-            HttpPost request = setupHttpPostRequest(uri, payload, stream, headers);
+            HttpPost request = setupHttpPostRequest(url, payload, stream, headers);
             int requestTimeout = timeoutMs > Integer.MAX_VALUE ?
-              Integer.MAX_VALUE :
-              Math.toIntExact(timeoutMs);
+                    Integer.MAX_VALUE :
+                    Math.toIntExact(timeoutMs);
             RequestConfig requestConfig = RequestConfig.custom().setSocketTimeout(requestTimeout).build();
             request.setConfig(requestConfig);
 
@@ -79,13 +77,13 @@ class Utils {
                 if (statusLine.getStatusCode() == 200) {
                     return responseContent;
                 } else {
-                    throw createExceptionFromResponse(url, response, null, responseContent);
+                    throw createExceptionFromResponse(urlStr, response, null, responseContent);
                 }
             }
         } catch (SocketTimeoutException e) {
-            throw new DataServiceException(url, "Timed out in post request:" + e.getMessage(), false);
+            throw new DataServiceException(urlStr, "Timed out in post request:" + e.getMessage(), false);
         } catch (JSONException | IOException e) {
-            throw new DataClientException(url, "Error in post request:" + e.getMessage(), e);
+            throw new DataClientException(urlStr, "Error in post request:" + e.getMessage(), e);
         }
         return null;
     }
@@ -144,7 +142,7 @@ class Utils {
             if (shouldPostToOriginalUrlDueToRedirect(redirectCount, responseStatusCode)) {
                 Optional<Header> redirectLocation = Arrays.stream(httpResponse.getHeaders(HttpHeaders.LOCATION)).findFirst();
                 if (redirectLocation.isPresent() && !redirectLocation.get().getValue().equals(url)) {
-                    return postToStreamingOutput(httpClient, redirectLocation.get().getValue(), payload, timeoutMs, headers,redirectCount + 1);
+                    return postToStreamingOutput(httpClient, redirectLocation.get().getValue(), payload, timeoutMs, headers, redirectCount + 1);
                 }
             }
         } catch (IOException ex) {
@@ -183,8 +181,7 @@ class Utils {
                     } else if (jsonObject.has("message")) {
                         message = jsonObject.getString("message");
                     }
-                }
-                catch (JSONException ex) {
+                } catch (JSONException ex) {
                     // It's not ideal to use an exception here for control flow, but we can't know if it's a valid JSON until we try to parse it
                 }
             } else {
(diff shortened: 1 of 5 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -101,6 +105,14 @@ public void setFlushImmediately(boolean flushImmediately) {
         this.flushImmediately = flushImmediately;
     }
 
+    public boolean getIgnoreFirstRecord() {
```
yihezkel, line 108: I'm going to change this to `isActive()` in keeping with Java's conventions for `boolean` getters.

yihezkel:
We would like to keep all Quickstarts (currently on Python and Java) functionally the same, so I'm making the same change there.

However, I'm pushing this first because:
1) It's best to first agree on the approach in 1 before doing them all.
2) We could theoretically hold up all PRs for each feature until a corresponding PR exists in each SDK, but that seems inefficient. I personally prefer "eventual consistency". In any case regarding the Quickstart, no feature is effectively released until the UI links to the relevant commit, so we can still release the feature at the same time even if we push this first.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
yihezkel, line 108: I'm going to change this to `isIgnoreFirstRecord()` in keeping with Java's conventions for `boolean` getters.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientTest.java
```diff
@@ -133,12 +134,29 @@ void IngestFromBlob_NullBlobSourceInfo_IllegalArgumentException() {
...
+
+        queuedIngestClient.ingestFromBlob(blobSourceInfo, ingestionProperties);
+
+        verify(azureStorageClientMock, atLeast(1)).postMessageToQueue(anyString(), captor.capture());
+        assertTrue ((captor.getValue()).contains("\"ignoreFirstRecord\":\"true\""));
+
+        ingestionProperties.setIgnoreFirstRecord(false);
+        queuedIngestClient.ingestFromBlob(blobSourceInfo, ingestionProperties);
+        verify(azureStorageClientMock, atLeast(1)).postMessageToQueue(anyString(), captor.capture());
+        assertTrue ((captor.getValue()).contains("\"ignoreFirstRecord\":\"false\""));
```
yihezkel, line 159: My approach is to prefer balance between UTs for each unit, and sometimes logically grouping UTs. In this case, I test both setting the field to true and false. (I could also check when the field isn't set at all, but I don't see much of a point.)

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -57,15 +56,14 @@ private Utils() {
         // Hide constructor, as this is a static utility class
     }
 
-    static String post(CloseableHttpClient httpClient, String url, String payload, InputStream stream, long timeoutMs,
-      Map<String, String> headers, boolean leaveOpen) throws DataServiceException, DataClientException {
-        URI uri = parseUriFromUrlString(url);
+    static String post(CloseableHttpClient httpClient, String urlStr, String payload, InputStream stream, long timeoutMs, Map<String, String> headers, boolean leaveOpen) throws DataServiceException, DataClientException {
```
AsafMah, line 59: This line is way too long, break it to one line per parameter.

AsafMah, line 59: Other than the rename of the parameter, there are no actual changes to this file, right?

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/KustoSampleApp.java
```diff
@@ -389,16 +397,16 @@ private static void ingest(Map<String, String> file, IngestionProperties.DataFor
...
-            ingestFromBlob(ingestClient, databaseName, tableName, uri, dataFormat, mappingName);
+            ingestFromBlob(ingestClient, databaseName, tableName, uri, dataFormat, mappingName, ignoreFirstRecord);
         } else {
             System.out.printf("Unknown source '%s' for file '%s'%n", sourceType, uri);
         }
     }
 
-    private static void ingestFromFile(IngestClient ingestClient, String databaseName, String tableName, String filePath, IngestionProperties.DataFormat dataFormat, String mappingName) {
-        IngestionProperties ingestionProperties = createIngestionProperties(databaseName, tableName, dataFormat, mappingName);
+    private static void ingestFromFile(IngestClient ingestClient, String databaseName, String tableName, String filePath, IngestionProperties.DataFormat dataFormat, String mappingName, boolean ignoreFirstRecord) {
```
AsafMah, line 408: Also some very long line here.
I think 160 chars is a good breaking point

ingest/src/test/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientTest.java
AsafMah: Maybe add an e2e test to make sure the parameter we send actually does someting?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
yihezkel: Most projects by big companies have strict conventions/enforcement, including for how long a line can be. We don't, and if you think we should then we should discuss that.

I personally don't see the benefit of a strict maximum line length, and only care about readability. This personally doesn't feel too long to me - it all fits on my laptop screen, and I'd still consider it readable if just the `throws` didn't fit.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientTest.java
AsafMah: You could say the same about testing with ingestion maps, but these tests already caught issues with how we send the parameters (in node), so I wouldn't say it's oos

yihezkel: Just to be clear, we're talking about tests that fail if and only if there is a bug in the external service (in this case, the Kusto service).
These tests you reference that helped you catch a bug - the bug was in the external service?

I maintain that **on principle** such tests should never be included. It then becomes a slippery and weakens the service's responsibility to properly test itself.

AsafMah: The bug wasn't on the external service, it was with how we sent the data.
An e2e test ensures that we have used the parameter correctly from the client's side.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
yihezkel: I've used maven plugins that enforce rules like this. As above, I agree in theory that we should agree on a standard as is common in big companies, but we haven't **discussed** a standard yet and I think that should be a prerequisite to **enforcing** a standard.
(In such a discussion, I would suggest standardizing on the maximum length that fits and is easily readable on the smallest of our screens, though of course it's OSS and it's not just us reading the code. I also would be in favor of maximum parameters per line until the char limit we decide upon instead of 1 - Java already tends to be somewhat sparse.)

To convey why discussing this first would make sense, search our Java SDK with the following regex:
`.{160}`
(Mea culpa that I think all of the longest lines in the search above are mine :-) )

May I call you today to discuss standardizing?

AsafMah: Sure, so I guess it shouldn't be a blocker for this PR, so I'll resolve.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientTest.java
yihezkel: The UTs I added verify that the Java SDK code sends the property when it should.

And I verified that this property as sent by the Java SDK is appropriate as expected by the service, and has the expected impact on the service (per https://docs.microsoft.com/azure/data-explorer/ingestion-properties).

But once we know that the property we send is appropriate, retesting that during every build/CI means to me that we're testing the service, not the SDK. It's that principle, less than the fact that E2Es are expensive, that makes me uneasy with adding a Kusto service test to the SDK.

ohadbitt (approved):

PR #216, opened 2022-02-13
Added formatting with spotless

AsafMah:
Adding formatting check for the code with spotless, if the formatting is incorrect, the build will fail.

The format chosen is https://github.com/palantir/palantir-java-format, which builds on the standard google format but allows modern features such as longer line limits (120 chars) and better lambda formatting.

To format with it, you can use the new spotless:apply maven command, or just set your IDE settings to it.


AsafMah:
Closing this, as we decided to use another formatting plugin.

PR #229, opened 2022-02-24
Added validation policy

### LINKED ISSUE #222: Add ValidationPolicy in IngestionProperties
Add ValidationPolicy options in IngestionProperties client, that indicates which validations to run/stop during ingestion (ValidationOptions, ValidationImplications) and what to do when that is violated.

AsafMah:
Added validation policy to ingestion properties

Fixes #222

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/ValidationPolicy.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/ValidationPolicy.java
@@ -1,10 +1,11 @@
 package com.microsoft.azure.kusto.ingest.result;
 
+import com.fasterxml.jackson.annotation.JsonValue;
 import java.io.Serializable;
 
 public class ValidationPolicy implements Serializable {
-    private ValidationOptions validationOptions = ValidationOptions.DoNotValidate;
-    private ValidationImplications validationPolicyType = ValidationImplications.BestEffort;
+    private ValidationOptions validationOptions = ValidationOptions.DO_NOT_VALIDATE;
+    private ValidationImplications validationPolicyType = ValidationImplications.VALIDATION_IMPLICATIONS;
 
     public ValidationPolicy() {
     }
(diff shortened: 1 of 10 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/ValidationPolicy.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.Serializable;
+
+public class ValidationPolicy implements Serializable {
+    private ValidationOptions validationOptions = ValidationOptions.DoNotValidate;
+    private ValidationImplications validationPolicyType = ValidationImplications.BestEffort;
+
+    public ValidationPolicy() {
+    }
+
+    public ValidationPolicy(ValidationOptions validationOptions, ValidationImplications validationPolicyType) {
```
yihezkel, line 12: What if they want to provide just 1 or the other?

yihezkel: I guess they should then just use the default constructor followed by the appropriate setter, NBD.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/ValidationPolicy.java
```diff
@@ -0,0 +1,68 @@
...
+    public ValidationImplications getValidationPolicyType() {
+        return validationPolicyType;
+    }
+
+    public void setValidationPolicyType(ValidationImplications validationPolicyType) {
+        this.validationPolicyType = validationPolicyType;
+    }
+
+    public enum ValidationOptions {
+        DoNotValidate("DoNotValidate"),
```
yihezkel, line 39: Java standard is enums in ALL CAPS. The value in the parens allow using the case our service expects.

yihezkel (changes requested):
2 minor changes requested (1 change x 2 enums)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/ValidationPolicy.java
AsafMah: Weird that IDEA ignores that by default, thanks!

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/ValidationPolicy.java
AsafMah: So I'm resolving.

yihezkel (approved):

Issue #218, opened 2022-02-15
Calling AAD access token endpoint for every ADX query call (not caching access token)


andrejpk:
**Describe the bug**
When I use the ADX Kusto ClientImpl and call .execute, there's a request every time to the AAD Token endpoint to get a new access token before the actual AAD call. This should be  fetched the first time then cached. 

**To Reproduce**
I have a modified sample here: https://github.com/andrejpk/azure-kusto-java/blob/apk/not-caching-msal-token/samples/src/main/java/Query.java 

This:
- removes outdated reference that no longer works for 3.x (so it works) see #219 
- adds Log4J to get logging output from MSAL
- loops to call the query 3x times

Run using the instructions in the samples/README

**Expected behavior**
I expect the call to the token endpoint on the first loop but subsequent calls should use a cached token.

This wasn't the case in a prior POC project where we used the 2.x client. 

**Screenshots**
Here are the logs emitted. You can see the `Token not found in the cache` messages and subsequent calls to fetch a new one each time.

```
[ForkJoinPool.commonPool-worker-19] DEBUG com.microsoft.aad.msal4j.ConfidentialClientApplication - [Correlation ID: 82859f65-e184-4589-8373-235d41b0b9d1] Execution of class com.microsoft.aad.msal4j.AcquireTokenSilentSupplier failed.
com.microsoft.aad.msal4j.MsalClientException: Token not found in the cache
	at com.microsoft.aad.msal4j.AcquireTokenSilentSupplier.execute(AcquireTokenSilentSupplier.java:98)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:69)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:18)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1700)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1692)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
[...]

PR #230, opened 2022-02-24
Assure cloud info is only initialized once

AsafMah:
A recent changed caused `setRequiredMembersBasedOnCloudInfo` to be called for every query, which meant creating a new token provider, and thus ruined token caching.

Solves #218 

This PR also improves E2E  tests- shorter timer and run after the check for faster tests

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
@@ -5,27 +5,29 @@ package com.microsoft.azure.kusto.data.auth;
 
 import com.microsoft.aad.msal4j.ConfidentialClientApplication;
 import com.microsoft.aad.msal4j.IClientCertificate;
-import com.microsoft.azure.kusto.data.exceptions.DataClientException;
-
-import org.jetbrains.annotations.NotNull;
-
+import com.microsoft.aad.msal4j.IConfidentialClientApplication;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
+import org.jetbrains.annotations.NotNull;
 
 public class ApplicationCertificateTokenProvider extends ConfidentialAppTokenProviderBase {
     private final IClientCertificate clientCertificate;
 
-    ApplicationCertificateTokenProvider(@NotNull String clusterUrl, @NotNull String applicationClientId, @NotNull IClientCertificate clientCertificate, String authorityId) throws URISyntaxException {
+    ApplicationCertificateTokenProvider(
+            @NotNull String clusterUrl,
+            @NotNull String applicationClientId,
+            @NotNull IClientCertificate clientCertificate,
+            String authorityId)
+            throws URISyntaxException {
         super(clusterUrl, applicationClientId, authorityId);
         this.clientCertificate = clientCertificate;
     }
 
     @Override
-    protected void setClientApplicationBasedOnCloudInfo() throws DataClientException {
-        try {
-            clientApplication = ConfidentialClientApplication.builder(applicationClientId, clientCertificate).authority(aadAuthorityUrl).validateAuthority(false).build();
-        } catch (MalformedURLException e) {
-            throw new DataClientException(clusterUrl, ERROR_INVALID_AUTHORITY_URL, e);
-        }
+    protected IConfidentialClientApplication getClientApplication() throws MalformedURLException {
+        return clientApplication = ConfidentialClientApplication.builder(applicationClientId, clientCertificate)
+                .authority(aadAuthorityUrl)
+                .validateAuthority(false)
+                .build();
     }
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
@@ -13,12 +13,7 @@ import org.jetbrains.annotations.NotNull;
 public class ApplicationCertificateTokenProvider extends ConfidentialAppTokenProviderBase {
     private final IClientCertificate clientCertificate;
 
-    ApplicationCertificateTokenProvider(
-            @NotNull String clusterUrl,
-            @NotNull String applicationClientId,
-            @NotNull IClientCertificate clientCertificate,
-            String authorityId)
-            throws URISyntaxException {
+    ApplicationCertificateTokenProvider(@NotNull String clusterUrl, @NotNull String applicationClientId, @NotNull IClientCertificate clientCertificate, String authorityId) throws URISyntaxException {
         super(clusterUrl, applicationClientId, authorityId);
         this.clientCertificate = clientCertificate;
     }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
@@ -1,0 +1,52 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+package com.microsoft.azure.kusto.data.auth;
+
+import com.microsoft.azure.kusto.data.exceptions.DataClientException;
+import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
+import java.net.URISyntaxException;
+import java.util.HashSet;
+import java.util.Set;
+import org.jetbrains.annotations.NotNull;
+
+public abstract class CloudDependantTokenProviderBase extends TokenProviderBase {
+    private static final String ERROR_INVALID_SERVICE_RESOURCE_URL =
+            "Error determining scope due to invalid Kusto Service Resource URL";
+    protected final Set<String> scopes = new HashSet<>();
+    protected CloudInfo cloudInfo = null;
+
+    CloudDependantTokenProviderBase(@NotNull String clusterUrl) throws URISyntaxException {
+        super(clusterUrl);
+    }
+
+    protected void onCloudInfoInitialized() throws DataClientException, DataServiceException {
+        try {
+            scopes.add(cloudInfo.determineScope());
+        } catch (URISyntaxException e) {
+            throw new DataServiceException(clusterUrl, ERROR_INVALID_SERVICE_RESOURCE_URL, e, true);
+        }
+    }
+
+    protected void initializeCloudInfo() throws DataClientException, DataServiceException {
+        if (cloudInfo != null) {
+            return;
+        }
+        synchronized (this) {
+            if (cloudInfo != null) {
+                return;
+            }
+
+            cloudInfo = CloudInfo.retrieveCloudInfoForCluster(clusterUrl);
+            onCloudInfoInitialized();
+        }
+    }
+
+    @Override
+    public String acquireAccessToken() throws DataServiceException, DataClientException {
+        initializeCloudInfo();
+        return acquireAccessTokenAfterCloudInfo();
+    }
+
+    public abstract String acquireAccessTokenAfterCloudInfo() throws DataServiceException, DataClientException;
+}
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependentTokenProviderBase.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependentTokenProviderBase.java
@@ -40,8 +40,8 @@ public abstract class CloudDependentTokenProviderBase extends TokenProviderBase 
     @Override
     public String acquireAccessToken() throws DataServiceException, DataClientException {
         initialize();
-        return acquireAccessTokenAfterCloudInfo();
+        return acquireAccessTokenImpl();
     }
 
-    public abstract String acquireAccessTokenAfterCloudInfo() throws DataServiceException, DataClientException;
+    protected abstract String acquireAccessTokenImpl() throws DataServiceException, DataClientException;
 }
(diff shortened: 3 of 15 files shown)


ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -240,6 +238,10 @@ private void assertRowCount(int expectedRowsCount, boolean checkViaJson) {
                     mainTableResult.next();
                     actualRowsCount = mainTableResult.getInt(0) - currentCount;
                 }
+
+
+                Thread.sleep(3000);
+                timeoutInSec -= 3;
```
alonadam, line 244: I would move these two lines back to the start of the loop -  this change results in a potential infinite loop when an exception is thrown from the code inside the IF statement.

alonadam: Also, I suggest you replace the while loop with a for loop as you know exactly how many times (max) you want to loop.
This way you don't need to manage timeoutInSec and it helps you skipping the thread.sleep in the first iteration.

data/src/main/java/com/microsoft/azure/kusto/data/auth/TokenProviderBase.java
```diff
@@ -24,7 +25,10 @@
         this.clusterUrl = UriUtils.setPathForUri(clusterUrl, "");
     }
 
-    protected void initializeCloudInfo() throws DataServiceException {
+    protected void setRequiredMembersBasedOnCloudInfo() throws DataClientException, DataServiceException {
+    }
+
+    protected void initializeCloudInfo() throws DataClientException, DataServiceException {
```
alonadam, line 31: Why is this not part of TokenProviderFactory? The class is not usable without CloudInfo so for me this is a perfect match for a factory.
I would change the C'tor to accept CloudInfo instead of clusterUrl and move setRequiredMembersBasedOnCloudInfo logic to the C'tor.

AsafMah: Some Token Providers don't need cloud info - AccessTokenTokenProvider, CallbackTokenProvider so we don't need to do the web request every time.

And we do use clusterUrl in the derived classes, to give more info in the exception.

alonadam: "Some Token Providers don't need cloud info" and yet you have two functions (setRequiredMembersBasedOnCloudInfo,initializeCloudInfo) with could info in their name.
Also you have a function in the base class  (determineScope) which depends on cloud info.
I don't understand the design of this class... We can talk about in Teams if you would like.

alonadam:
@AsafMah 
A few suggestions (I hope they are all applicable as I am not familiar with the code):

- make determineScope() static and move it outside of the base class. You said that in some cases cloudInfo is null, making it even more reasonable to remove this code from the base class. Maybe put it in CloudInfo class? static String determineScope(CloudInfo cloudInfo) {...}
- setRequiredMembersBasedOnCloudInfo, initializeCloudInfo remove these functions entirely. The existence of this functions demonstrates the flaws in the design of the class.
-  For classes who depend on CloudInfo - update their C'tor to accept CloudInfo param and use the factory (TokenProviderFactory) to create an instance of CloudInfo and pass it to the relevant classes.

AsafMah:
The problem is that we don't want to create the CloudInfo too early.
Getting it requires a web request, and we don't want to do it before the user actually uses the token provider, so it can't be already initialized in the ctor.

I'm thinking about two options -
1. Make some lazy wrapper to `CloudInfo`, so that I can pass it in the ctor but it won't be actually retrieved until needed.
2. Add a derived class to `TokenProviderBase`, `CloudInfoTokenProvider` that will have all the cloudinfo stuff in it. 
So Token providers that don't need it will inherit from base, and ones that do will inherit from the new class.

alonadam (approved):

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
```diff
@@ -0,0 +1,52 @@
...
+import java.net.URISyntaxException;
+import java.util.HashSet;
+import java.util.Set;
+import org.jetbrains.annotations.NotNull;
+
+public abstract class CloudDependantTokenProviderBase extends TokenProviderBase {
+    private static final String ERROR_INVALID_SERVICE_RESOURCE_URL =
+            "Error determining scope due to invalid Kusto Service Resource URL";
+    protected final Set<String> scopes = new HashSet<>();
+    protected CloudInfo cloudInfo = null;
```
alonadam, line 17: Just a suggestion, not mandatory - can we make this private and change initializeCloudInfo() to getCloudInfo()?
Make it more "singleton" style. This can prevent derived classes from accessing this variable when it is NULL.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
```diff
@@ -0,0 +1,52 @@
...
+package com.microsoft.azure.kusto.data.auth;
+
+import com.microsoft.azure.kusto.data.exceptions.DataClientException;
+import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
+import java.net.URISyntaxException;
+import java.util.HashSet;
+import java.util.Set;
+import org.jetbrains.annotations.NotNull;
+
+public abstract class CloudDependantTokenProviderBase extends TokenProviderBase {
```
yihezkel, line 13: nit: dependant -> dependent

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
```diff
@@ -0,0 +1,52 @@
...
+            throw new DataServiceException(clusterUrl, ERROR_INVALID_SERVICE_RESOURCE_URL, e, true);
+        }
+    }
+
+    protected void initializeCloudInfo() throws DataClientException, DataServiceException {
+        if (cloudInfo != null) {
+            return;
+        }
+        synchronized (this) {
+            if (cloudInfo != null) {
```
yihezkel, line 36: We might've discussed this before, but I don't think double-checked locking works in Java without ensuring JRE optimizations didn't partially initialize `cloudInfo`.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
```diff
@@ -0,0 +1,52 @@
...
+
+    protected void onCloudInfoInitialized() throws DataClientException, DataServiceException {
+        try {
+            scopes.add(cloudInfo.determineScope());
+        } catch (URISyntaxException e) {
+            throw new DataServiceException(clusterUrl, ERROR_INVALID_SERVICE_RESOURCE_URL, e, true);
+        }
+    }
+
+    protected void initializeCloudInfo() throws DataClientException, DataServiceException {
```
yihezkel, line 31: I might be confusing things here, but I thought we talked about encapsulating any `CloudInfo` code into that class.
Do we need to maintain a `CloudInstance` here, as opposed to that being part of `CloudInfo`'s `Map` and we just retrieve what we need by indexing with our `clusterUrl`?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
```diff
@@ -0,0 +1,52 @@
...
+    private static final String ERROR_INVALID_SERVICE_RESOURCE_URL =
+            "Error determining scope due to invalid Kusto Service Resource URL";
+    protected final Set<String> scopes = new HashSet<>();
+    protected CloudInfo cloudInfo = null;
+
+    CloudDependantTokenProviderBase(@NotNull String clusterUrl) throws URISyntaxException {
+        super(clusterUrl);
+    }
+
+    protected void onCloudInfoInitialized() throws DataClientException, DataServiceException {
```
yihezkel, line 23: nit: "on" doesn't sound intuitive to me. I think "after" makes more sense.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
AsafMah: Ok, instead of having cloudInfo as a member, the init method now receives it as a parameter. 
So, it will not be null.

There is still the problem that fields that are dependent on cloudInfo initialization will still be null, but I'm not sure what we can do about this.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
```diff
@@ -5,27 +5,29 @@
...
 
 public class ApplicationCertificateTokenProvider extends ConfidentialAppTokenProviderBase {
     private final IClientCertificate clientCertificate;
 
-    ApplicationCertificateTokenProvider(@NotNull String clusterUrl, @NotNull String applicationClientId, @NotNull IClientCertificate clientCertificate, String authorityId) throws URISyntaxException {
+    ApplicationCertificateTokenProvider(
+            @NotNull String clusterUrl,
+            @NotNull String applicationClientId,
+            @NotNull IClientCertificate clientCertificate,
+            String authorityId)
```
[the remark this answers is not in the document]
AsafMah: Defering to the team

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
AsafMah: I removed the double-checking, now it's one synchronized method. WDYT?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
AsafMah: Cloudinfo is no longer a member, let me know if that's what you had in mind?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
AsafMah: Renamed to `initializeWithCloudInfo` now that it's a parameter. Is that ok?

data/src/main/java/com/microsoft/azure/kusto/data/auth/ApplicationCertificateTokenProvider.java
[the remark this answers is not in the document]
AsafMah: For now I removed the formatting change, we can discuss it in a later PR

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
yihezkel: Looks great! Your quick escape if not null should ensure solid performance.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
yihezkel: I don't recall the different approaches I considered, but this is a very good one. Most importantly, `CloudInfo` is not longer a member of these classes, which makes a lot more sense.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependantTokenProviderBase.java
yihezkel: Perfect!

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependentTokenProviderBase.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    @Override
+    public String acquireAccessToken() throws DataServiceException, DataClientException {
+        initialize();
+        return acquireAccessTokenAfterCloudInfo();
+    }
+
+    public abstract String acquireAccessTokenAfterCloudInfo() throws DataServiceException, DataClientException;
```
yihezkel, line 46: I don't love the naming, `acquireAccessTokenAfterCloudInfo()`. I get the issue - `acquireAccessToken()` is used by non-CloudInfo classes as well, so its name should be that.

I don't have any good suggestions. I'm leaning towards:
1) Rename to `acquireAccessTokenAfterCloudEnriched()`
2) Striking this method altogether, and expect subclasses to implement `acquireAccessToken()` by first calling `super()`, and then acquiring the token.

yihezkel, line 46: I don't love the naming, `acquireAccessTokenAfterCloudInfo()`. I get the issue - `acquireAccessToken()` is used by non-CloudInfo classes as well, so its name should be that.

I don't have any good suggestions. I'm leaning towards:
1) Rename to `acquireAccessTokenAfterCloudEnriched()`
2) Striking this method altogether, and expect subclasses to implement `acquireAccessToken()` by first calling `super()`, and then acquiring the token.

yihezkel (approved):
1 remaining comment, which is minor. I do suggest taking one of the suggestions I provide or a better one you might come up with, but I'd say it's merge-worthy even at this point.

PR #231, opened 2022-03-16
Add formatter plugin

AsafMah:
#### Pull Request Description

Adding plugin to maven to format and check the formatting of the code.

The plugin uses eclipse's xml style that we can edit.
See kusto-style.xml.
If you're using Intellij or Eclipse you can import the file into the IDE to use it with the built-in formatting.

The style file is duplicated in quick_start, since it's supposed to be stand-alone.
I don't like it, but I don't have a better idea.

---

**Fixes:**
- Code is now formatted and gate-checked before formatting.


data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -56,12 +56,16 @@ public ClientImpl(ConnectionStringBuilder csb, CloseableHttpClient httpClient) t
         Objects.requireNonNull(clusterUrlForParsing.getAuthority(), "clusterUri.authority");
         String auth = clusterUrlForParsing.getAuthority().toLowerCase();
         if (host == null && auth.endsWith(FEDERATED_SECURITY_SUFFIX)) {
-            csb.setClusterUrl(new URIBuilder().setScheme(clusterUrlForParsing.getScheme()).setHost(auth.substring(0, clusterUrlForParsing.getAuthority().indexOf(FEDERATED_SECURITY_SUFFIX))).toString());
+            csb
```
yogilad, line 59: 59-60 is an odd formatting decision

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -252,13 +272,25 @@ private CommandType determineCommandType(String command) {
         headers.put("x-ms-client-request-id", clientRequestId);
 
         UUID activityId = UUID.randomUUID();
-        String activityContext = String.format("%s%s/%s, ActivityId=%s, ParentId=%s, ClientRequestId=%s", JAVA_INGEST_ACTIVITY_TYPE_PREFIX, activityTypeSuffix, activityId, activityId, activityId, clientRequestId);
+        String activityContext = String
```
yogilad, line 275: 275-276 - Here as well

data/src/main/java/com/microsoft/azure/kusto/data/HttpClientFactory.java
```diff
@@ -47,18 +47,19 @@ static HttpClientFactory getInstance() {
      * @return a new Apache HTTP client
      */
     CloseableHttpClient create(HttpClientProperties providedProperties) {
-        final HttpClientProperties properties = Optional.ofNullable(providedProperties)
+        final HttpClientProperties properties = Optional
+                .ofNullable(providedProperties)
                 .orElse(HttpClientProperties.builder().build());
-        final HttpClientBuilder httpClientBuilder = HttpClientBuilder.create()
+        final HttpClientBuilder httpClientBuilder = HttpClientBuilder
```
yogilad, line 53: Can we make it add a new line when indentation decreases following a multi line command

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -275,7 +276,10 @@ public void setDataFormat(@NotNull String dataFormatName) {
         try {
             this.dataFormat = DataFormat.valueOf(dataFormatName.toUpperCase());
         } catch (IllegalArgumentException ex) {
-            log.warn("IngestionProperties.setDataFormat(): Invalid dataFormatName of {}. Per the API's specification, DataFormat property value wasn't set.", dataFormatName);
+            log
```
yogilad, line 279: 279-280 odd here as well

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionProperties.java
```diff
@@ -341,8 +345,12 @@ void validate() throws IngestionClientException {
             }
         } else { // a mapping was provided
             if (dataFormat.getIngestionMappingKind() != null && !dataFormat.getIngestionMappingKind().equals(ingestionMappingKind)) {
-                message.appendln("Wrong ingestion mapping for format '%s'; mapping kind should be '%s', but was '%s'.",
-                        dataFormat.getKustoValue(), dataFormat.getIngestionMappingKind().getKustoValue(), ingestionMappingKind != null ? ingestionMappingKind.getKustoValue() : "null");
+                message
```
yogilad, line 348: and here

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -147,15 +158,19 @@ IngestionResult ingestFromStream(StreamSourceInfo streamSourceInfo, IngestionPro
...
             log.debug("Executing streaming ingest");
-            this.streamingClient.executeStreamingIngest(ingestionProperties.getDatabaseName(),
-                    ingestionProperties.getTableName(),
-                    stream,
-                    clientRequestProperties,
-                    dataFormat.getKustoValue(),
-                    ingestionProperties.getIngestionMapping().getIngestionMappingReference(),
-                    !(streamSourceInfo.getCompressionType() == null || !streamSourceInfo.isLeaveOpen()));
+            this.streamingClient
+                    .executeStreamingIngest(
```
yogilad, line 166: Are we using tab=8 or is it a visual style of GitHub?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
yogilad: point is without the new line it would have been shorter.
How much control do we have here

AsafMah: Ok, I chose a new style which seems to be as good as it gets.

The downside is that I had to generate it directly from eclipse, so the comments are gone.
But now I know that their interface is fairly simple to understand and edit, so it's no big deal.


You can just look at the diff of the last commit to see if the changes are what you had in mind.

yogilad (approved):

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -94,9 +96,12 @@ public KustoOperationResult execute(String database, String command, ClientReque
...
             return new KustoOperationResult(response, clusterEndpoint.endsWith("v2/rest/query") ? "v2" : "v1");
         } catch (KustoServiceQueryError e) {
-            throw new DataServiceException(clusterEndpoint,
-                    "Error found while parsing json response as KustoOperationResult:" + e.getMessage(), e, e.isPermanent());
-        } catch (Exception e){
+            throw new DataServiceException(
+                    clusterEndpoint,
+                    "Error found while parsing json response as KustoOperationResult:" + e.getMessage(),
+                    e,
+                    e.isPermanent());
```
yihezkel, line 103: Is this formatting approach really an improvement?
I think sometimes it is, but I'd argue often it isn't, in which case I don't think it should be enforced.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -189,7 +205,10 @@ public InputStream executeStreamingQuery(String database, String command) throws
     }
 
     @Override
-    public InputStream executeStreamingQuery(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException {
+    public InputStream executeStreamingQuery(
+            String database,
+            String command,
+            ClientRequestProperties properties) throws DataServiceException, DataClientException {
```
yihezkel, line 211: Similar to method calls, I don't think this approach to method signatures is beneficial often enough to warrant being enforced.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -228,10 +249,10 @@ private CommandType determineCommandType(String command) {
...
     }
 
-    private Map<String, String> generateIngestAndCommandHeaders(ClientRequestProperties properties,
-                                                                String clientRequestIdPrefix,
-                                                                String activityTypeSuffix)
-            throws DataServiceException, DataClientException {
+    private Map<String, String> generateIngestAndCommandHeaders(
+            ClientRequestProperties properties,
+            String clientRequestIdPrefix,
+            String activityTypeSuffix) throws DataServiceException, DataClientException {
```
yihezkel, line 255: I think this *reduces* readability.

data/src/main/java/com/microsoft/azure/kusto/data/CloseParentResourcesStream.java
```diff
@@ -1,74 +1,74 @@
...
+    @Override
+    public synchronized void reset() throws IOException {
+        innerStream.reset();
+    }
+
+    @Override
+    public boolean markSupported() {
+        return innerStream.markSupported();
+    }
+}
```
yihezkel, line 74: This must be line endings, no?

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -325,4 +329,4 @@ public static ConnectionStringBuilder createWithAadManagedIdentity(String cluste
         csb.useManagedIdentityAuth = true;
         return csb;
     }
-}
\ No newline at end of file
+}
```
yihezkel, line 332: What's it doing here? Forcing a new line at the end of the file? If so, I don't know of any practical reasons why we'd want this.

data/src/main/java/com/microsoft/azure/kusto/data/CloseParentResourcesStream.java
yihezkel: Assuming so, would you please first do a PR that fixes the line endings, and then merge that into this PR so we can see the actual diff?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
AsafMah: The point of the formatting is not to be an improvement for every particular case, but to be *consistent*.
We decided on a limit of 160 characters per line, both this and the next sample are over it, that's why we split it.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
AsafMah: Again, it's consistent.
If you want it to be consistently like the old style we can discuss this again (though I tried to spark this discussion multiple times and nobody answered).
I don't care what it would be in the end, as long as it's one style.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
AsafMah: New lines at the end of files have always been part of formatting standard.
In C it's actually required by the language (though most compilers will ignore it unless you compile with pedantic).
It let's you `cat` files more easily, and gives you better diffs/blames with git.
And again, basically every code standard does them.

data/src/main/java/com/microsoft/azure/kusto/data/CloseParentResourcesStream.java
AsafMah: I'll try doing one just for the newlines and indentations.

AsafMah: https://github.com/Azure/azure-kusto-java/pull/234

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
yihezkel: I agree that there are many things we want to make consistent. I'm not sure all things need to be consistent, though.

In this case, there are 2 options if a line is more than 160 chars:
1) Spill the remainder over to 1 new line (assuming the remainder is less than 160 chars, else a 3rd line, etc)
2) Change the entire signature such that each parameter gets its own line

I'm saying that I don't think (2) is an improvement over (1).

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
yihezkel: Yeah I know about it being required in C and that many projects force it, but I'm asking if there are any practical reasons why we'd want this. Anything that affects us or our users such that forcing this adds value.

Diffing/blames shouldn't be relevant, as in Java it's uncommon to have multiple classes in 1 file, and I don't remember any of us ever adding a new class to the end of a file.
`cat`ting a file without a newline at the end just causes the prompt to remain on the previous line. NDB, IMHO. Plus, whomever among us/our users who develop in Linux would most likely use `more` instead of `cat`.

data/src/main/java/com/microsoft/azure/kusto/data/CloseParentResourcesStream.java
yihezkel: Approved

AsafMah: Now it's less polluted.

PR #237, opened 2022-03-30
Add retry mechanism for potentiality throttled operations.

alonadam:
#### Pull Request Description

Use external library (resilience4j-retry) to support retry of potentially throttled operations.

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -4,11 +4,7 @@
 package com.microsoft.azure.kusto.data;
 
 import com.microsoft.azure.kusto.data.auth.CloudInfo;
-import com.microsoft.azure.kusto.data.exceptions.DataClientException;
-import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
-import com.microsoft.azure.kusto.data.exceptions.DataWebException;
-import com.microsoft.azure.kusto.data.exceptions.OneApiError;
-import com.microsoft.azure.kusto.data.exceptions.WebException;
+import com.microsoft.azure.kusto.data.exceptions.*;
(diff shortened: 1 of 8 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -74,10 +70,13 @@ static String post(CloseableHttpClient httpClient, String urlStr, String payload
...
                 StatusLine statusLine = response.getStatusLine();
                 String responseContent = EntityUtils.toString(entity);
-                if (statusLine.getStatusCode() == 200) {
-                    return responseContent;
-                } else {
-                    throw createExceptionFromResponse(urlStr, response, null, responseContent);
+                switch (statusLine.getStatusCode()) {
+                    case HttpStatus.SC_OK:
+                        return responseContent;
+                    case HttpStatus.SC_TOO_MANY_REQUESTS:
```
ohadbitt, line 76: just make sure we really send this form dm

ohadbitt:
I dont think this is needed - we already have a sort of retry mechanism here
In case of failure - the next refresh would be 15 minutes from now .. 
https://github.com/Azure/azure-kusto-java/blob/ca04a5881ac52c1861775611a505cadf6cfa2308/ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java#L93

ohadbitt (approved):

ohadbitt (approved):

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
ohadbitt: also - make sure that the last exception (after 4 attempts ) will be thrown to the user as a DataServiceException with the http status code and all as if it was thrown before this change

Issue #241, opened 2022-04-11
KustoResultSetTable getTimestamp should respect ISO format


xatier:
**Describe the bug**

Kusto uses UTC in most of the places for timestamps, the SDK returns the timestamp string in the ISO format, e.g. `"2022-03-21T13:00:00Z"`, when using `executeToJsonResult`.

 However, when using `execute` with `KustoResultSetTable` and `getTimestamp`, it returns time parsed incorrectly in local time. The current implementation of `getTimestamp` should respect the `Z` postfix, which indicates it's in UTC.

https://github.com/Azure/azure-kusto-java/blob/master/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java#L256

```
return Timestamp.valueOf(StringUtils.chop(getString(columnIndex)).replace("T", " "));
```

workaround: use `getString` instead and parse with `joda.time`

```
import org.joda.time.Instant;

KustoResultSetTable kustoRestuls = kustoClient.execute(...).getPrimaryResults()
Instant.parse(kustoResults.getString("timestamp"))
```

**Expected behavior**

`getTimestamp` respect the `Z` postfix on the timestamp and use UTC instead of local time.

**Desktop (please complete the following information):**
 - OS: macOS
 - Version 3.1.0

PR #246, opened 2022-04-28
Update getKustoDateTime with more lenient formatter

breriksenms:
#### Pull Request Description

The `getKustoDateTime` function uses 2 potential formatters for the LocalDateTime that either include no digits for the fraction of a second or 7 digits for the fraction of a second. We ran into a situation where the value coming back from ADX had a different number of digits and would cause this method to fail. This code updates that function to just use the built-in `ISO_LOCAL_DATE_TIME` formatter, which allows for 0-9 digits in the fraction of a second, and appends the literal 'Z' to the end of the formatter to account for that value being present in the datetimes that ADX sends back.

---

#### Future Release Comment
Updating the `getKustoDateTime` function to use the built-in `ISO_LOCAL_DATE_TIME` formatter, which allows for 0-9 digits in the fraction of a second, and append the literal 'Z' to the end of the formatter to account for that value being present in the datetimes that ADX sends back.

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- Use a more lenient LocalDateTime formatter for parsing in `getKustoDateTime`

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
@@ -4,7 +4,6 @@
 package com.microsoft.azure.kusto.data;
 
 import com.microsoft.azure.kusto.data.exceptions.KustoServiceQueryError;
-import com.microsoft.azure.kusto.data.format.CslDateTimeFormat;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.time.FastDateFormat;
 import org.json.JSONArray;
@@ -37,6 +36,8 @@ public class KustoResultSetTable {
     private static final String ROWS_PROPERTY_NAME = "Rows";
     private static final String EXCEPTIONS_PROPERTY_NAME = "Exceptions";
     private static final String EXCEPTIONS_MESSAGE = "Query execution failed with multiple inner exceptions";
+    private static DateTimeFormatter kustoDateTimeFormatter = new DateTimeFormatterBuilder().parseCaseInsensitive()
+    .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME).appendLiteral('Z').toFormatter();
 
     private final List<List<Object>> rows;
     private String tableName;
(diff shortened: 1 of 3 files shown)


xatier:
Related to https://github.com/Azure/azure-kusto-java/issues/241

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -465,15 +465,9 @@ public LocalDateTime getKustoDateTime(int columnIndex) {
...
-            dateTimeFormatter = new DateTimeFormatterBuilder().parseCaseInsensitive()
-                    .append(DateTimeFormatter.ofPattern(CslDateTimeFormat.KUSTO_DATETIME_PATTERN_NO_FRACTIONS)).toFormatter();
-        } else {
-            dateTimeFormatter = new DateTimeFormatterBuilder().parseCaseInsensitive()
-                    .append(DateTimeFormatter.ofPattern(CslDateTimeFormat.KUSTO_DATETIME_PATTERN)).toFormatter();
-        }
-        return LocalDateTime.parse(getString(columnIndex), dateTimeFormatter);
+        DateTimeFormatter dateTimeFormatter = new DateTimeFormatterBuilder().parseCaseInsensitive()
+                .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME).appendLiteral('Z').toFormatter();
+        return LocalDateTime.parse(dateString, dateTimeFormatter);
```
ohadbitt, line 470: You can keep the formatter as static member

ohadbitt (approved):
Nice!

yogilad:
We need to add additional tests for fractions and partial values before completing this PR.
If the existing formatters are not required, they need to be deleted.
Also, there are additional formats which are (probably) used for user input which do not support timezone notation

breriksenms:
Do you have examples of the different datetime formats that you would want this method to support?

AsafMah:
@breriksenms This PR replaces both KUSTO_DATETIME_PATTERN_NO_FRACTIONS and KUSTO_DATETIME_PATTERN
Can you test it still works with both formats at least?

Also KUSTO_DATETIME_FORMATS in CslDateTimeFormat still doesn't seem to support timezones.

breriksenms:
I added new tests for the getKustoDateTime function that covers the different formats it supports. I'm not certain what you want to do with the KUSTO_DATETIME_FORMATS and timezones. I think that's more for sending a datetime out to Kusto, not for receiving it. Kusto only sends back datetimes in UTC timezone. The original getKustoDateTime function only supported UTC and then only in 2 formats. This PR just expands it to support a variable number of digits in the fraction of a second.

PR #253, opened 2022-06-28
Fix localhost handling

abestel:
#### Pull Request Description

[This PR](https://github.com/Azure/azure-kusto-java/pull/185/) added support for localhost in the Kusto client. This is extremely useful to mock ADX using Wiremock for example.

There seems to be one last issue however in `com.microsoft.azure.kusto.data.Utils.parseUriFromUrlString`. A typical input `url` would be `http://localhost:[port]/v2/rest/query`; however, when parsed, the `cleanUrl` host is `localhost` and not `http://localhost`.

The fix in this PR is to use the exact same logic as the ClientImpl constructor which is:

```java
aadAuthenticationHelper = clusterUrl.toLowerCase().startsWith(CloudInfo.LOCALHOST) ? null : TokenProviderFactory.createTokenProvider(csb, httpClient);
```

---

#### Future Release Comment

Full support of localhost in Kusto Data to effectively be able to mock ADX APIs.

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
@@ -253,7 +253,7 @@ class Utils {
     private static URI parseUriFromUrlString(String url) throws DataClientException {
         try {
             URL cleanUrl = new URL(url);
-            if ("https".equalsIgnoreCase(cleanUrl.getProtocol()) || cleanUrl.getHost().equalsIgnoreCase(CloudInfo.LOCALHOST)) {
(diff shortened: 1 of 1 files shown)


AsafMah:
Wouldn't it make more sense then to change the constant LOCALHOST to "localhost", and then we'll still parse the url and avoid weird scenarios like "abc.localhost/something"

abestel:
> Wouldn't it make more sense then to change the constant LOCALHOST to "localhost", and then we'll still parse the url and avoid weird scenarios like "abc.localhost/something"

I tried to make as little a change as possible. Honestly I don't know if the logic around `localhost` needs to be complex and handle cases like you mention. As far as I understand, the localhost case is only used to be able to mock ADX (?) and that should not require much complexity.

AsafMah:
I see, sure

AsafMah (approved):

PR #254, opened 2022-06-28
Send sourceMessageCreationTime, which can help troubleshoot delays

yihezkel:
It's already included in the C# SDK for the same reason.

#### Pull Request Description

Send sourceMessageCreationTime, which can help troubleshoot delays

---

#### Future Release Comment
**Features:**
- Send sourceMessageCreationTime


AsafMah (approved):
Great! luckily this already exists in all other SDKs, so we don't need to reimplement

PR #255, opened 2022-06-28
Improve wrong endpoint detection

yihezkel:
#### Pull Request Description

Improve wrong endpoint detection:
- Improve exception handling: if `retrieveServiceType()` can't determine the endpoint's `ServiceType`, it shouldn't throw an (uncaught, no less) exception but instead should return `null`
- Improve message to be clearer and more actionable for the user
- Log warnings instead of errors if we can't suggest the right URL

---

#### Future Release Comment
**Features:**
- Improve wrong endpoint detection


AsafMah:
All the places we call it are inside exception handlers, which is why it throws:

```
        } catch (StorageException e) {
            throw new IngestionServiceException("Failed to ingest from blob", e);
        } catch (IOException | URISyntaxException e) {
            throw new IngestionClientException("Failed to ingest from blob", e);
        } catch (IngestionServiceException e) {
            validateEndpointServiceType(connectionDataSource, EXPECTED_SERVICE_TYPE);
            throw e;
        }
```

I don't think it should just be a warning, personally - the user might not notice it.

But if you do make it so, maybe change the calling sites too to use the return values.

yihezkel:
> All the places we call it are inside exception handlers, which is why it throws:
> 
> ```
>         } catch (StorageException e) {
>             throw new IngestionServiceException("Failed to ingest from blob", e);
>         } catch (IOException | URISyntaxException e) {
>             throw new IngestionClientException("Failed to ingest from blob", e);
>         } catch (IngestionServiceException e) {
>             validateEndpointServiceType(connectionDataSource, EXPECTED_SERVICE_TYPE);
>             throw e;
>         }
> ```
> 
> I don't think it should just be a warning, personally - the user might not notice it.
> 
> But if you do make it so, maybe change the calling sites too to use the return values.

It's a little confusing, but the calling context isn't within a `try`, but within a `catch` (see https://stackoverflow.com/questions/3526167/exception-handling-try-catch-inside-catch)
If we throw within `validateEndpointServiceType()`/`retrieveServiceType()`, the exception handling flow designed here in `ingestFromBlob()` is broken.

The log item I'm proposing changing to a warning is notifying the user that we can't determine the suggested corrected endpoint. We still notify the user as an error *that* they're using the wrong endpoint in any case (see https://github.com/Azure/azure-kusto-java/blob/master/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java#L35).

AsafMah:
The link you sent says it's completely fine.
In fact, we throw an exception from every other branch in the catch, too.

The only thing this method does right now, is instead of copy pasting its logic to every catch clause it unifies them.

But I can go with the warning solution, as long as if someone encounters it and can't figure it out it's on you 😛

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -48,16 +55,16 @@ protected String generateEndpointSuggestion(String existingSuggestedEndpointUri,
                 existingEndpoint = new URIBuilder(dataSource);
                 endpointUriToSuggestStr = emendEndpointUri(existingEndpoint);
             } catch (URISyntaxException e) {
-                log.error("Couldn't parse dataSource '{}', so no suggestion can be made.", dataSource, e);
+                log.warn("Couldn't parse dataSource '{}', so no suggestion can be made.", dataSource, e);
```
ohadbitt, line 58: I think this message can be unclear to the user he didnt ask for any suggestion- lets say  : After failing the request - client tried to parse the URL and failed.

yihezkel:
> The link you sent says it's completely fine. In fact, we throw an exception from every other branch in the catch, too.
> 
> The only thing this method does right now, is instead of copy pasting its logic to every catch clause it unifies them.
> 
> But I can go with the warning solution, as long as if someone encounters it and can't figure it out it's on you 😛

@AsafMah What that link says is completely fine is having a `try` within a `catch`, which is rarely but sometimes necessary. Here, that wouldn't be the most logical approach. If we just add a `try`/`catch` which then ignores that exception, why did we throw an exception in the first place?

I'm not sure how we got to that point, but we shouldn't be throwing unless we catch it in a deliberate way. So any other paths where we are in a `catch` and that code might `throw`, but nowhere up the chain do we properly handle it, we should fix that as well.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
yihezkel: Updated message

AsafMah:
We don't handle it because up until now, we wanted the user to see this exception directly.
If we want to use logs it isn't needed

yihezkel:
> We don't handle it because up until now, we wanted the user to see this exception directly. If we want to use logs it isn't needed

We wanted the user to directly see an exception that, when generating a different exception that the wrong endpoint was provided, we couldn't generate a suggestion?!

AsafMah (approved):
Discussed offline - approved (check with ohad though)

PR #256, opened 2022-06-28
Black: Roll back requiring there be no newline after a field's annotation

yihezkel:
#### Pull Request Description

Black: Roll back requiring there be no newline after a field's annotation. Alternatively, require that there *be* a newline after a field's annotation.

As discussed previously, my perspective is that if one code style isn't clearly better than alternatives, the best approach is to let the devs decide. We don't need all aspects of code style to be 100% conforming in all ways, and taking away freedom unnecessarily from devs makes devs sad.

---

#### Future Release Comment
**Fixes:**
- Roll back requiring there be no newline after a field's annotation

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/WebException.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/WebException.java
@@ -4,7 +4,8 @@ import org.apache.http.HttpResponse;
 import org.jetbrains.annotations.Nullable;
 
 public class WebException extends Exception {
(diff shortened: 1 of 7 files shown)


AsafMah:
Can you give an example to what this allows?

AsafMah:
P.S you should find my old email that lets you run the formatting command on commit, it will save you a lot of headaches

AsafMah:
Sure, makes sense.

AsafMah (approved):
Fix the build before merging

PR #251, opened 2022-06-16
Dependency Bumps and Apache httpclient5 migration

The-Funk:
#### Pull Request Description

-updated all dependencies to latest
-migrated apache httpclient to version 5

---

#### Future Release Comment
This upgrades the version of the Apache httpclient used by the library to version 5 for incoming support of asynchronous/non-blocking ADX queries.

**Breaking Changes:**
There are 2 test failures total.

1. createExceptionFromBlankErrorMessage() in UtilitiesTest in kusto-data module
2. E2ETest in kusto-ingest

There is a TODO in the code for the max connections parameter. It does not appear that version 5 of the Apache http client supports a max connections parameter.

**Features:**
Everything is up to date and new Apache client appears to work

**Fixes:**
NA

PR #257, opened 2022-07-03
Update to client5

AsafMah:
#### Pull Request Description

Taken from #251, but I re-did the merge so now it's here.

Pull Request Description
-updated all dependencies to latest
-migrated apache httpclient to version 5

I'm pretty sure that changing the dependency like that is breaking, but we have to if we want to support async in the future.


The-Funk:
Any update to this?

The-Funk:
Hey, just checking in again to see if this is still in the works.

AsafMah:
We have decided to close this for now.
Since we have decided to move to java 11, we will probably use the new standard http client when we will make an async interface.
Thank you for the contribution.

PR #258, opened 2022-07-10
QueueRequestOptions setter

ohadbitt:
Setter for QueueRequestOptions to be used by queued ingest client users.
New interface for queued ingest client


AsafMah (approved):
Looks good!

PR #259, opened 2022-07-11
Formatting throws like jetbrains

AsafMah:
#### Pull Request Description

Jetbrains doesn't have an equivalent setting for how the "throw clause" in our formatting was set up,
so this PR changes the formatting to fit into the IDE.

Now it's harder to break your builds by accident.


ohadbitt (approved):
trust you

PR #260, opened 2022-07-25
Java 11 and extract Process name

ohadbitt:
Send process name and process user as default application name and user name on headers


ohadbitt:
This includes an alternative way to extract real process name for windows but should remove next commit
 
Once removed  process name is  extracted from executed file name

AsafMah:
The process name will always be "java.exe", so we need a method to get the main "package"/class/something.

We might want to upgrade to java 11 anyway in our next breaking change.

ohadbitt:
Decided on making this the main module name

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -58,6 +53,16 @@ private ConnectionStringBuilder(String clusterUrl) {
...
         this.clientVersionForTracing = null;
         this.applicationNameForTracing = null;
+        this.initProcessNameForTracing();
+    }
+
+    private void initProcessNameForTracing() {
+        this.processNameForTracing = System.getProperty("sun.java.command");
+        this.userNameForTracing = System.getProperty("user.name");
+        if (processNameForTracing != null) {
+            String[] p = processNameForTracing.split(" ")[0].split("\\.");
```
AsafMah, line 63: Comment explaining what's going on here.
The '\\.' is suspicious, it looks windows specific, and path specific?

Plus, make the code more defensive.
You are doing a lot of checks that can throw, and we don't want to not to be able to create a KSCB if the format of the property isn't correct, or it doesn't exist.

AsafMah (changes requested):

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
ohadbitt: What throws here

ohadbitt: split always returns at least one element

data/src/main/java/com/microsoft/azure/kusto/data/UriUtils.java
```diff
@@ -67,4 +67,22 @@ public static String[] getSasAndEndpointFromResourceURL(String url) throws URISy
...
+            if (processNameForTracing.equalsIgnoreCase("jar") && splitCommand.length > 1) {
+                processNameForTracing = splitCommand[splitCommand.length - 2];
+            }
+            int lastDirSeparator = processNameForTracing.lastIndexOf(File.separatorChar);
+            if (lastDirSeparator > 0) {
+                processNameForTracing = processNameForTracing.substring(lastDirSeparator + 1);
+            }
+        }
+
+        return processNameForTracing;
```
AsafMah, line 88:
```suggestion
        if (cmdLine == null) {
            return null;
        }

        String[] splitCommand = cmdLine.split(" ")[0].split("\\.");
        String processNameForTracing = splitCommand[splitCommand.length - 1];
        if (processNameForTracing.equalsIgnoreCase("jar") && splitCommand.length > 1) {
            processNameForTracing = splitCommand[splitCommand.length - 2];
        }
        int lastDirSeparator = processNameForTracing.lastIndexOf(File.separatorChar);
        if (lastDirSeparator > 0) {
            processNameForTracing = processNameForTracing.substring(lastDirSeparator + 1);
        }

        return processNameForTracing;
```

I think you can reduce the nesting here and make it more readable

AsafMah (approved):

thewisenerd:
apologies for I am uninitiated, but is there a valid reason for breaking 1.8 compat with this change?

PR #263, opened 2022-08-08
Verify endpoints 

ohadbitt:
Validate endpoint validity before talking to service
Copy of .Net client classes and json


data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/FastSuffixMatcher.java
```diff
@@ -0,0 +1,94 @@
...
+     * @return true if at least one of the rules matched.
+     */
+    public Boolean isMatch(String candidate) {
+        return Match(candidate).isMatch;
+    }
+
+    public MatchResult Match(String candidate) {
+        Ensure.argIsNotNull(candidate, "candidate");
+
+        if (candidate.length() >= suffixLength) {
```
AsafMah, line 79: early return

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/FastSuffixMatcher.java
```diff
@@ -0,0 +1,94 @@
...
+            List<MatchRule> matchRules = m_rules.getOrDefault(StringUtils.GetStringTail(candidate, suffixLength),
```
AsafMah, line 80: You can save on an allocation by using the regular get and doing a null check

data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
```diff
@@ -0,0 +1,33 @@
+package com.microsoft.azure.kusto.data;
+
+public class StringUtils
+{
+    public static boolean startsWithIgnoreCase(String str, String prefix)
+    {
+        return str.regionMatches(true, 0, prefix, 0, prefix.length());
+    }
+
+    public static boolean endsWithIgnoreCase(String str, String suffix)
```
AsafMah: Both of these are available in import org.apache.commons.lang3.StringUtils; which we use elsewhere in the code

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/FastSuffixMatcher.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public static FastSuffixMatcher Create(List<MatchRule> rules) {
+        Ensure.argIsNotNull(rules, "rules");
+        int minRuleLength = rules.stream().min(Comparator.comparing(MatchRule::getSuffixLength))
+                .map(MatchRule::getSuffixLength).orElse(0);
+
+        Ensure.isTrue(minRuleLength > 0 && minRuleLength != Integer.MAX_VALUE, "Cannot have a match rule " +
+                "whose length is zero");
+
+        Map<String, List<MatchRule>> processedRules = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
AsafMah, line 27: Any specific reason to use a treemap over an hashmap? why does it matter if they're ordered? especially since we don't sort them

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/FastSuffixMatcher.java
```diff
@@ -0,0 +1,94 @@
...
+        for (MatchRule rule : rules) {
```
AsafMah, line 28: You can one-liner this:
`        Map<String, List<MatchRule>> processedRules =  rules.stream().collect(Collectors.groupingBy(rule -> StringUtils.GetStringTail(rule.suffix,
minRuleLength)));`

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/FastSuffixMatcher.java
```diff
@@ -0,0 +1,94 @@
...
+            String suffix = StringUtils.GetStringTail(rule.suffix, minRuleLength);
+            List<MatchRule> list = processedRules.computeIfAbsent(suffix, k -> new ArrayList<>());
+            list.add(rule.Clone());
+        }
+
+        return new FastSuffixMatcher(minRuleLength, processedRules);
+    }
+
+    private FastSuffixMatcher(int suffixLength, Map<String, List<MatchRule>> rules) {
+        this.suffixLength = suffixLength;
```
AsafMah, line 38: naming consistency

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
```diff
@@ -0,0 +1,144 @@
...
+public class KustoTrustedEndpoints {
+    private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+    public static boolean enableWellKnownKustoEndpointsValidation = true;
+    private static final Map<String, FastSuffixMatcher> matchers;
+    private static FastSuffixMatcher additionalMatcher;
+    private static Predicate<String> overrideMatcher; // We could unify this with matchers, but separating makes debugging easier
+
+    static {
+        matchers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
AsafMah, line 30: Again, is it important that it's sorted?

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
```diff
@@ -0,0 +1,144 @@
...
+
+        WellKnownKustoEndpointsData.getInstance().AllowedEndpointsByLogin.forEach((key, value) -> {
```
AsafMah, line 32: Here you can also use streams, but it's up to you if it's more readable:
```java
        matchers = WellKnownKustoEndpointsData.getInstance().AllowedEndpointsByLogin.entrySet().stream()
                .collect(Collectors.toMap(Map.Entry::getKey, e -> FastSuffixMatcher.Create(
                        Stream.concat(
                                e.getValue().AllowedKustoHostnames.stream().map(hostname -> new MatchRule(hostname, true)),
                                e.getValue().AllowedKustoSuffixes.stream().map(suffix -> new MatchRule(suffix, false)))
                                .collect(Collectors.toList())
                )));
```

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
```diff
@@ -0,0 +1,144 @@
...
+
+    private static void ValidateHostnameIsTrusted(String hostname, String loginEndpoint) throws KustoClientInvalidConnectionStringException {
+        // The loopback is unconditionally allowed (since we trust ourselves)
+        if (CloudInfo.isLocalAddress(hostname)) {
+            return;
+        }
+
+        // Either check the override matcher OR the matcher:
+        if (overrideMatcher != null) {
+            if (overrideMatcher.test(hostname)) {
```
AsafMah, line 121: Should the additional matcher still be on w/ the override policy?

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/WellKnownKustoEndpointsData.java
```diff
@@ -0,0 +1,46 @@
...
+
+    // For Deserialization
+    public WellKnownKustoEndpointsData() {
+    }
+
+    private static WellKnownKustoEndpointsData readInstance() {
+        try {
+            // Beautiful !
+            ObjectMapper objectMapper = new ObjectMapper();
+            try (InputStream resourceAsStream = WellKnownKustoEndpointsData.class.getResourceAsStream(
```
AsafMah, line 36: Make sure it's compiled into the jar and doesn't need to exist on the filesystem

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/WellKnownKustoEndpointsData.java
```diff
@@ -0,0 +1,46 @@
...
+        public ArrayList<String> AllowedKustoHostnames;
+    }
+
+    @JsonIgnore
+    public ArrayList<String> _Comments;
+    public HashMap<String, AllowedEndpoints> AllowedEndpointsByLogin;
+    private static WellKnownKustoEndpointsData instance = null;
+
+    public static WellKnownKustoEndpointsData getInstance() {
+        if (instance == null)
```
AsafMah, line 22: thread safety?

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Adds the rules that determine if a hostname is a valid/trusted Kusto endpoint
+     * (extends existing rules).
+     *
+     * @param rules   - A set of rules
+     * @param replace - If true nullifies the last added rules
+     */
+    public synchronized static void AddTrustedHosts(List<MatchRule> rules, boolean replace) {
```
AsafMah, line 101: It's not synchronized when you use it

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
```diff
@@ -0,0 +1,144 @@
...
+
+        additionalMatcher = null;
+        overrideMatcher = null;
+    }
+
+    /**
+     * @param matcher Rules that determine if a hostname is a valid/trusted Kusto endpoint
+     *  (replaces existing rules). Null means "use the default policy".
+     */
+    public static void SetOverridePolicy(Predicate<String> matcher) {
```
AsafMah, line 47: Shouldn't this be locked/synchronized too?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -38,12 +38,35 @@ public class CloudInfo {
             DEFAULT_FIRST_PARTY_AUTHORITY_URL);
     public static final String LOCALHOST = "http://localhost";
 
-    private static final Map<String, CloudInfo> cache = new HashMap<>();
-
     static {
         cache.put(LOCALHOST, DEFAULT_CLOUD);
     }
 
+    public static boolean isLocalAddress(String host)
```
AsafMah: move this to UriUtils or something?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -136,19 +137,25 @@ public String executeToJsonResult(String database, String command, ClientRequest
         command = command.trim();
         CommandType commandType = determineCommandType(command);
         long timeoutMs = determineTimeout(properties, commandType);
+        // TODO save the uri once - no need to format everytime
         String clusterEndpoint = String.format(commandType.getEndpoint(), clusterUrl);
```
AsafMah, line 141: Where do we format again?

AsafMah (changes requested):

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
ohadbitt: we reach this code for each command executed - this can be done once

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
ohadbitt: no

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
ohadbitt: what do you mean

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/WellKnownKustoEndpointsData.java
ohadbitt: will synchronize the method

pom.xml
```diff
@@ -45,7 +45,7 @@
         <httpclient.version>4.5.13</httpclient.version>
         <httpcore.version>4.4.15</httpcore.version>
         <json.version>20201115</json.version>
-        <fasterxml.jackson.core.version>[2.11.0,2.13.0]</fasterxml.jackson.core.version>
+        <fasterxml.jackson.core.version>2.12.5</fasterxml.jackson.core.version>
```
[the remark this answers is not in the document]
ohadbitt: because we have a conflict with azure core which uses 2.15.5

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/FastSuffixMatcher.java
```diff
@@ -0,0 +1,94 @@
...
+    public static FastSuffixMatcher Create(FastSuffixMatcher existing, List<MatchRule> rules) {
+        if (existing == null || existing.m_rules.size() == 0) {
+            return Create(rules);
+        }
+
+        if (rules == null || rules.isEmpty()) {
+            return existing;
+        }
+
+        List<MatchRule> list = new ArrayList<>();
```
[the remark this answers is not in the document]
ohadbitt: It seemed a little too much but will take

pom.xml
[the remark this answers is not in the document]
AsafMah: Ok

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
AsafMah: validateHostnameIsTrusted is not synchronized

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
AsafMah: What if someone calls validateHostnameIsTrusted at the same time as this?

ohadbitt: then he will have either the new matcher or the old matcher.

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/FastSuffixMatcher.java
ohadbitt: this is not case insensitive tho

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
AsafMah: 1.        if (overrideMatcher != null) {
2.           if (overrideMatcher.test(hostname)) {

1 passes, anthoer thread changes overrideMatcher to null, 2 runs, you have an NPE

data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
AsafMah: You resolved this, but the code is still here

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/FastSuffixMatcher.java
AsafMah: Fair enough. I would have saved/fetched with toLower, but you can keep as is.

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
ohadbitt: it matters that its case insensitive

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
ohadbitt: i don't like using it just for the sake of using it - this example is too much for me

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
ohadbitt: i dont think its critical - whats important about the override matcher is that if it succeeds it will be return a match

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/WellKnownKustoEndpointsData.java
ohadbitt: its a resource

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
ohadbitt: how about i change it to 
`
    public static void AddTrustedHosts(List<MatchRule> rules, boolean replace) {

        if (rules.isEmpty())
        {
            if (replace)
            {
                additionalMatcher = null;
            }
            return;
        }

        additionalMatcher = FastSuffixMatcher.Create(replace ? null : additionalMatcher, rules);
    }
`

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/FastSuffixMatcher.java
ohadbitt: just for case insensitivity

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
ohadbitt: :(
fixed

AsafMah (approved):
nice

PR #267, opened 2022-08-16
Validate endpoint once

ohadbitt:
As the cluster url never changes - only need to validate once (if valid)


data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -42,6 +43,7 @@ public class ClientImpl implements Client, StreamingClient {
     private final String applicationNameForTracing;
     private final String userNameForTracing;
     private final CloseableHttpClient httpClient;
+    private boolean _endpoint_validated = false;
```
AsafMah, line 46: Camelcase

yihezkel (approved):
Approved, with minor change requested

PR #268, opened 2022-08-18
Automatic conversion of query/ingestion urls in ingest clients

AsafMah:
Changes:

* Removed complex endpoint detection system (as per discussion)
* Automatic conversion of query/ingestion urls in ingest clients - adds or removes "ingest-"
* Changed terminology to "ingestion endpoint" and "query endpoint" where possible
* Deprecated un-needed ctors

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -11,68 +11,31 @@ import java.lang.invoke.MethodHandles;
 import java.net.URISyntaxException;
 
 public abstract class IngestClientBase {
+    static final String INGEST_PREFIX = "://ingest-";
+    static final String NO_INGEST_PREFIX = "://";
+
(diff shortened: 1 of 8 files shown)


ohadbitt:
Did you test with proxies ?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -1,78 +1,31 @@
...
 
-    protected abstract String retrieveServiceType();
-
-    protected abstract String emendEndpointUri(URIBuilder existingEndpoint);
-
-    static boolean shouldCompress(CompressionType sourceCompressionType, IngestionProperties.DataFormat dataFormat) {
-        return (sourceCompressionType == null) && (dataFormat == null || dataFormat.isCompressible());
+    static String getQueryEndpoint(String clusterUrl) {
+        if (clusterUrl.contains(INGEST_PREFIX)) {
+            return clusterUrl.replaceFirst(INGEST_PREFIX, "");
```
ohadbitt, line 25: you can just do replace first - no need contains and else if

ohadbitt (approved):
Seems OK

PR #269, opened 2022-08-28
Support tables with special characters .-_

yihezkel:
#### Pull Request Description

Support tables with special characters .-_

---

#### Future Release Comment
**Fixes:**
- Support tables with special characters .-_
- Rename method `StringUtils.getStringTail` to conform to Java nomenclature
- Unify StringUtils (the popular Apache Commons one) with our proprietary one by the same name (in com.microsoft.azure.kusto.data), by inheriting and adding our own static methods.

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
@@ -17,7 +17,15 @@ public class StringUtils extends org.apache.commons.lang3.StringUtils {
         return val.substring(val.length() - minRuleLength);
     }
 
-    public static String quoteEntityName(String name) {
-        return "['" + name + "']";
+    public static String normalizeEntityName(String name) {
+        if (StringUtils.isBlank(name)) {
+            return name;
+        } else if (name.startsWith("[")) {
+            return name;
+        } else if (!name.contains("'")) {
+            return "['" + name + "']";
+        } else {
+            return "[\"" + name + "\"]";
+        }
     }
 }
--- a/data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
@@ -1,7 +1,11 @@
 package com.microsoft.azure.kusto.data;
 
-public class StringUtils {
-    public static String GetStringTail(String val, int minRuleLength) {
+public class StringUtils extends org.apache.commons.lang3.StringUtils {
+    private StringUtils() {
+        // Hide constructor for static class
+    }
+
+    public static String getStringTail(String val, int minRuleLength) {
         if (minRuleLength <= 0) {
             return "";
         }
@@ -13,6 +17,7 @@ public class StringUtils {
         return val.substring(val.length() - minRuleLength);
     }
 
-    private StringUtils() {
+    public static String quoteEntityName(String name) {
+        return "['" + name + "']";
     }
 }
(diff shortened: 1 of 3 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
```diff
@@ -1,7 +1,11 @@
 package com.microsoft.azure.kusto.data;
 
-public class StringUtils {
-    public static String GetStringTail(String val, int minRuleLength) {
+public class StringUtils extends org.apache.commons.lang3.StringUtils {
```
yihezkel, line 3: There are downsides to this approach, but we only want to add additional methods to the existing Apache Commons suite, so extending makes the most sense, especially to prevent name collisions/needing to add a package-qualified prefix to each call.

data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
```diff
@@ -13,6 +17,7 @@ public static String GetStringTail(String val, int minRuleLength) {
         return val.substring(val.length() - minRuleLength);
     }
 
-    private StringUtils() {
+    public static String quoteEntityName(String name) {
+        return "['" + name + "']";
```
AsafMah, line 21: What if the table is already quoted? 
Because we don't need to escape chars, we can assume that if it has the prefix/suffix we don't need to add it:
`StringUtils.appendIfMissing(StringUtils.prependIfMissing(str, "['"), "']");`

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/SampleApp.java
```diff
@@ -375,7 +375,7 @@ private static void preIngestionQuerying(ConfigJson config, Client kustoClient)
      * @param tableSchema  Table Schema
      */
     private static void alterMergeExistingTableToProvidedSchema(Client kustoClient, String databaseName, String tableName, String tableSchema) {
-        String command = String.format(".alter-merge table %s %s", tableName, tableSchema);
+        String command = String.format(".alter-merge table %s %s", StringUtils.quoteEntityName(tableName), tableSchema);
```
AsafMah, line 378: If this method is only used for SampleApp, should it be on the utils files in this package?
(So people downloading it could see how it's implemented)

data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
ohadbitt: in spark we have this code and i really prefer this logic:
def normalizeTableName(table: String): String = {
    val tableName = table.replace("-", "_")

    if (tableName.startsWith("[")) {
      tableName
    }
    else if (!tableName.contains("'")) {
      "['" + tableName + "']"
    }
    else {
      "[\"" + tableName + "\"]"
    }
  }

quickstart/src/main/java/com/microsoft/azure/kusto/quickstart/SampleApp.java
yihezkel: I put it in the main package because we may want to expand its use, and suspect we will (this is an initial step).

They can still see what it does to the table name, but it doesn't have to be duplicated to the quickstart package because the intent of the Sample app is to get them started, not to be too minutiae-oriented.

data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
yihezkel: I think the intent with the quickstart is that they provide just the table name, and the documentation illustrates that as well. But I do agree, because I really intended this to be expanded for use outside quickstart as well, so I'll add some more logic to cover cases where it's already escaped.

PR #287, opened 2023-02-09
IngestionStatusInTableDescription to support serialization

asaharn:
#### Pull Request Description
- TableClient defaults to null at time of serialization
- Rebuilding TableClient in case of deserialization

---

**Fixes:**
- Fixes serialization for IngestionStatusInTableDescription

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
@@ -128,6 +128,7 @@ public class QueuedIngestClientImpl extends IngestClientBase implements QueuedIn
                 ingestionStatusInTable.setPartitionKey(ingestionBlobInfo.getId().toString());
                 ingestionStatusInTable.setRowKey(ingestionBlobInfo.getId().toString());
                 ingestionBlobInfo.setIngestionStatusInTable(ingestionStatusInTable);
+                ingestionStatusInTable.setTableUri(statusTable.getUri());
                 azureStorageClient.azureTableInsertEntity(statusTable.getTable(), new TableEntity(id, id).setProperties(status.getEntityProperties()));
                 tableStatuses.add(ingestionBlobInfo.getIngestionStatusInTable());
             }
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
@@ -128,7 +128,6 @@ public class QueuedIngestClientImpl extends IngestClientBase implements QueuedIn
                 ingestionStatusInTable.setPartitionKey(ingestionBlobInfo.getId().toString());
                 ingestionStatusInTable.setRowKey(ingestionBlobInfo.getId().toString());
                 ingestionBlobInfo.setIngestionStatusInTable(ingestionStatusInTable);
-                ingestionStatusInTable.setTableUri(statusTable.getUri());
                 azureStorageClient.azureTableInsertEntity(statusTable.getTable(), new TableEntity(id, id).setProperties(status.getEntityProperties()));
                 tableStatuses.add(ingestionBlobInfo.getIngestionStatusInTable());
             }
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
@@ -27,7 +27,6 @@ import com.microsoft.azure.kusto.ingest.utils.IngestionUtils;
 import com.microsoft.azure.kusto.ingest.utils.SecurityUtils;
 import com.microsoft.azure.kusto.ingest.utils.TableWithSas;
 import com.univocity.parsers.csv.CsvRoutines;
-import org.apache.http.client.utils.URIBuilder;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/IngestionStatusInTableDescription.java
```diff
@@ -46,6 +49,14 @@ public TableClient getTableClient() {
         return tableClient;
     }
 
+    public void setTableUri(String tableUri) {
```
ohadbitt, line 52: you dont need all these 
use the connection string (add a getTableConnectionString)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/TableReportIngestionResult.java
```diff
@@ -17,10 +19,13 @@ public TableReportIngestionResult(List<IngestionStatusInTableDescription> descri
     }
 
     @Override
-    public List<IngestionStatus> getIngestionStatusCollection() {
+    public List<IngestionStatus> getIngestionStatusCollection() throws URISyntaxException {
         List<IngestionStatus> results = new LinkedList<>();
         for (IngestionStatusInTableDescription descriptor : descriptors) {
             TableClient table = descriptor.getTableClient();
+            if (table == null) {
```
ohadbitt, line 26: I'd rather let the IngestionStatusInTableDescription.getTableClient class handle the null case - also it can then save it for later uses

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
```diff
@@ -128,6 +128,7 @@ public IngestionResult ingestFromBlob(BlobSourceInfo blobSourceInfo, IngestionPr
                 ingestionStatusInTable.setPartitionKey(ingestionBlobInfo.getId().toString());
                 ingestionStatusInTable.setRowKey(ingestionBlobInfo.getId().toString());
                 ingestionBlobInfo.setIngestionStatusInTable(ingestionStatusInTable);
+                ingestionStatusInTable.setTableUri(statusTable.getUri());
```
ohadbitt, line 131: as explained below - no need

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/IngestionStatusInTableDescription.java
```diff
@@ -1,63 +1,61 @@
...
+    }
+
+    public void setRowKey(String rowKey) {
+        this.rowKey = rowKey;
+    }
+
+    public TableClient getTableClient() throws URISyntaxException {
+        if (tableClient == null) {
+            tableClient = TableWithSas.TableClientFromUrl(getTableConnectionString(), null);
+        }
```
ohadbitt, line 54: nit: new line after if

ohadbitt (approved):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/IngestionStatusInTableDescription.java
```diff
@@ -1,52 +1,61 @@
...
+
+    public String getRowKey() {
+        return rowKey;
+    }
+
+    public void setRowKey(String rowKey) {
+        this.rowKey = rowKey;
+    }
+
+    public TableClient getTableClient() throws URISyntaxException {
```
AsafMah, line 51: This is a breaking change. I think the best we can do is to catch the exception inside and log, and return null

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/IngestionStatusInTableDescription.java
```diff
@@ -1,52 +1,61 @@
...
+
+    public String getPartitionKey() {
+        return partitionKey;
+    }
+
+    public void setTableConnectionString(String tableConnectionString) {
+        this.tableConnectionString = tableConnectionString;
+    }
+
+    public String getTableConnectionString() {
```
AsafMah, line 35: Get comes before set

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/IngestionStatusInTableDescription.java
```diff
@@ -1,52 +1,61 @@
...
+
+    @JsonProperty
+    private String partitionKey;
+
+    @JsonProperty
+    private String rowKey;
+
+    @JsonIgnore
+    private transient TableClient tableClient;
+
```
AsafMah, line 26: Do you need both jsonignore and transient? geniunely asking, I'm not sure how this works

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
```diff
@@ -1,293 +1,293 @@
...
-import com.azure.storage.blob.models.BlobStorageException;
-import com.azure.storage.common.policy.RequestRetryOptions;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.microsoft.azure.kusto.data.*;
-import com.microsoft.azure.kusto.data.auth.ConnectionStringBuilder;
-import com.microsoft.azure.kusto.ingest.exceptions.IngestionClientException;
-import com.microsoft.azure.kusto.ingest.exceptions.IngestionServiceException;
-import com.microsoft.azure.kusto.ingest.result.IngestionResult;
-import com.microsoft.azure.kusto.ingest.result.IngestionStatus;
-import com.microsoft.azure.kusto.ingest.result.IngestionStatusInTableDescription;
```
AsafMah, line 16: Are there any changes in this file?

asaharn: Ah, seems Line break type changed from LF to CRLF

ingest/src/main/java/com/microsoft/azure/kusto/ingest/result/IngestionStatusInTableDescription.java
asaharn: as per my understanding, Looks like transient can be enough.  transient will ignore for serialization and get the default value at time of deserialization. which basically completes the functionality of jsonignore

AsafMah (approved):
Please use the logger to log the error.
Otherwise approve once you fix the build.

ag-ramachandran (approved):

PR #292, opened 2023-04-04
Replace non-ascii characters in headers to be in line with the service.

AsafMah:
### Fixed
* Replace non-ascii characters in headers to be in line with the service.

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -308,12 +308,7 @@ class ClientImpl implements Client, StreamingClient {
         headers.put("x-ms-activitycontext", activityContext);
 
         // replace non-ascii characters in header values with '?'
-        for (Map.Entry<String, String> entry : headers.entrySet()) {
-            String value = entry.getValue();
-            if (value != null) {
-                entry.setValue(value.replaceAll("[^\\x00-\\x7F]", "?"));
-            }
-        }
+        headers.replaceAll((_i, v) -> v == null ? null : v.replaceAll("[^\\x00-\\x7F]", "?"));
         return headers;
     }
 
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -307,6 +307,13 @@ class ClientImpl implements Client, StreamingClient {
                 activityId, activityId, activityId, clientRequestId);
         headers.put("x-ms-activitycontext", activityContext);
 
+        // replace non-ascii characters in header values with '?'
+        for (Map.Entry<String, String> entry : headers.entrySet()) {
+            String value = entry.getValue();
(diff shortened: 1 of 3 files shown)


ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -122,6 +122,7 @@ public static void setUp() throws IOException {
 
         ConnectionStringBuilder engineCsb = ConnectionStringBuilder.createWithAadApplicationCredentials(System.getenv("ENGINE_CONNECTION_STRING"), appId,
                 appKey, tenantId);
+        engineCsb.setUserNameForTracing("Java_E2ETest_ø");
```
yihezkel, line 125: This is great for testing that sending a non-ASCII char doesn't cause ingestion to **fail**. But can we use a getter to ensure in this test that it properly translates the char to a "?"?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -307,6 +307,13 @@ private Map<String, String> generateIngestAndCommandHeaders(ClientRequestPropert
                 activityId, activityId, activityId, clientRequestId);
         headers.put("x-ms-activitycontext", activityContext);
 
+        // replace non-ascii characters in header values with '?'
+        for (Map.Entry<String, String> entry : headers.entrySet()) {
```
yihezkel, line 311: nit: I like how you did it as well, but I think this is a great scenario for using the stream API.

yihezkel (approved):
Approving with suggestions (though 1 is a nit, and I didn't verify the other is even possible)

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -307,6 +307,13 @@ private Map<String, String> generateIngestAndCommandHeaders(ClientRequestPropert
...
+            String value = entry.getValue();
+            if (value != null) {
+                entry.setValue(value.replaceAll("[^\\x00-\\x7F]", "?"));
```
yihezkel, line 314: Suggest using `[^\\x20-\\x7E]`, per https://www.asciitable.com/

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
AsafMah: Discussed offline

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
AsafMah: Done

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
AsafMah: The point was to align this to the behavior of the c# SDK.
They don't touch other character, so we won't either.

PR #293, opened 2023-04-16
Bug fix - timeout as string

ohadbitt:
Timeout seconds were not computed correctly, failed when serializing to string and then from string


AsafMah (approved):
Ok looks good

PR #295, opened 2023-04-16
Dependency bumps, jackson, commons features, httpclient, and msal4j

The-Funk:
changed some minor dependency versions to get rid of some CVEs and Intellij warnings.
ran all tests, everything but E2E passes (E2E did not pass without modification to the main kusto branch either, error was the same in both scenarios)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,18 +41,18 @@
         <!-- Versions below are for several dependencies we're using in the data & ingest modules -->
         <!-- Compile dependencies -->
         <slf4j.version>1.8.0-beta4</slf4j.version>
-        <commons-lang3.version>3.11</commons-lang3.version>
+        <commons-lang3.version>3.12.0</commons-lang3.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -41,18 +41,18 @@
         <!-- Versions below are for several dependencies we're using in the data & ingest modules -->
         <!-- Compile dependencies -->
         <slf4j.version>1.8.0-beta4</slf4j.version>
```
ohadbitt, line 43: Can you add a comment that we like to align versions with some databricks runtime dependency list 
it actually looks the same as https://learn.microsoft.com/en-us/azure/databricks/release-notes/runtime/13.0
:)

ohadbitt (approved):

The-Funk:
@ohadbitt sorry I clicked resolve conversation instead of add comment. I have added the requested comment in the pom file.

PR #296, opened 2023-04-17
Added autocrrectendpoint usage

ronmonetaMicro:
### Added
support for autoCorrectEndpoint boolean flag for url converstions

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -56,6 +56,7 @@ public abstract class IngestClientBase implements IngestClient {
 
         return isLocalFlagg || isIPFlag || authority.equalsIgnoreCase("onebox.dev.kusto.windows.net");
     }
+
     protected abstract IngestionResult ingestFromFileImpl(FileSourceInfo fileSourceInfo, IngestionProperties ingestionProperties)
             throws IngestionClientException, IngestionServiceException;
 
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -1,6 +1,10 @@
 package com.microsoft.azure.kusto.ingest;
 
+import com.microsoft.azure.kusto.data.UriUtils;
 import com.microsoft.azure.kusto.ingest.source.CompressionType;
+
+import java.net.URI;
+import java.util.regex.Pattern;
 
 public abstract class IngestClientBase {
     static final String INGEST_PREFIX = "ingest-";
@@ -13,19 +17,33 @@ public abstract class IngestClientBase {
     }
 
     static String getIngestionEndpoint(String clusterUrl) {
-        if (clusterUrl.contains(INGEST_PREFIX)) {
+        if (clusterUrl == null || clusterUrl.contains(INGEST_PREFIX) || isReservedHostname(clusterUrl)) {
             return clusterUrl;
-        } else {
+        }
+        if (clusterUrl.contains(PROTOCOL_SUFFIX)) {
             return clusterUrl.replaceFirst(PROTOCOL_SUFFIX, PROTOCOL_SUFFIX + INGEST_PREFIX);
         }
+        return INGEST_PREFIX + clusterUrl;
     }
 
     static String getQueryEndpoint(String clusterUrl) {
-        if (clusterUrl.contains(INGEST_PREFIX)) {
-            return clusterUrl.replaceFirst(INGEST_PREFIX, "");
-        } else {
-            return clusterUrl;
-        }
+        return isReservedHostname(clusterUrl) ? clusterUrl : clusterUrl.replaceFirst(INGEST_PREFIX, "");
     }
 
+    static boolean isReservedHostname(String rawUri) {
+        URI uri = URI.create(rawUri);
+        if (!uri.isAbsolute()) {
+            return true;
+        }
+        var host = uri.getHost().toLowerCase();
+        Pattern IPV4_PATTERN = Pattern.compile("^(25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)(\\.(25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)){3}$");
+
+        Pattern IPV6_STD_PATTERN = Pattern.compile("^(?:[0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}$");
+
+        Pattern IPV6_HEX_COMPRESSED_PATTERN = Pattern.compile("^((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)::((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)$");
+
+        var isIPFlag = IPV4_PATTERN.matcher(host).matches() || IPV6_STD_PATTERN.matcher(host).matches() || IPV6_HEX_COMPRESSED_PATTERN.matcher(host).matches();
+
+        return UriUtils.isLocalAddress(host) || isIPFlag || host.equalsIgnoreCase("onebox.dev.kusto.windows.net");
+    }
 }
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -1,6 +1,5 @@
 package com.microsoft.azure.kusto.ingest;
 
-import com.microsoft.azure.kusto.data.UriUtils;
 import com.microsoft.azure.kusto.ingest.source.CompressionType;
 import org.apache.http.conn.util.InetAddressUtils;
 
@@ -45,9 +44,9 @@ public abstract class IngestClientBase implements IngestClient {
         if (!uri.isAbsolute()) {
             return true;
         }
-        var authority = uri.getAuthority().toLowerCase();
-        var isIPFlag = InetAddressUtils.isIPv4Address(authority) || InetAddressUtils.isIPv6Address(authority);
-        var isLocalFlagg = false;
+        String authority = uri.getAuthority().toLowerCase();
+        boolean isIPFlag = InetAddressUtils.isIPv4Address(authority) || InetAddressUtils.isIPv6Address(authority);
+        boolean isLocalFlagg;
         try {
             isLocalFlagg = InetAddress.getByName(authority).isLoopbackAddress();
         } catch (UnknownHostException e) {
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -50,7 +50,7 @@ public abstract class IngestClientBase implements IngestClient {
         try {
             isLocalFlagg = InetAddress.getByName(authority).isLoopbackAddress();
         } catch (UnknownHostException e) {
-            throw new RuntimeException(e);
+            isLocalFlagg = false;
         }
 
         return isLocalFlagg || isIPFlag || authority.equalsIgnoreCase("onebox.dev.kusto.windows.net");
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -36,7 +36,7 @@ public abstract class IngestClientBase implements IngestClient {
     }
 
     static String getQueryEndpoint(String clusterUrl) {
-        return isReservedHostname(clusterUrl) ? clusterUrl : clusterUrl.replaceFirst(INGEST_PREFIX, "");
+        return (clusterUrl == null || isReservedHostname(clusterUrl)) ? clusterUrl : clusterUrl.replaceFirst(INGEST_PREFIX, "");
     }
 
     static boolean isReservedHostname(String rawUri) {
@@ -46,14 +46,9 @@ public abstract class IngestClientBase implements IngestClient {
         }
         String authority = uri.getAuthority().toLowerCase();
         boolean isIPFlag = InetAddressUtils.isIPv4Address(authority) || InetAddressUtils.isIPv6Address(authority);
-        boolean isLocalFlagg;
-        try {
-            isLocalFlagg = InetAddress.getByName(authority).isLoopbackAddress();
-        } catch (UnknownHostException e) {
-            isLocalFlagg = false;
-        }
+        boolean isLocalFlag = authority.contains("localhost");
 
-        return isLocalFlagg || isIPFlag || authority.equalsIgnoreCase("onebox.dev.kusto.windows.net");
+        return isLocalFlag || isIPFlag || authority.equalsIgnoreCase("onebox.dev.kusto.windows.net");
     }
 
     protected abstract IngestionResult ingestFromFileImpl(FileSourceInfo fileSourceInfo, IngestionProperties ingestionProperties)
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -2,9 +2,11 @@ package com.microsoft.azure.kusto.ingest;
 
 import com.microsoft.azure.kusto.data.UriUtils;
 import com.microsoft.azure.kusto.ingest.source.CompressionType;
+import org.apache.http.conn.util.InetAddressUtils;
 
+import java.net.InetAddress;
 import java.net.URI;
-import java.util.regex.Pattern;
+import java.net.UnknownHostException;
 
 public abstract class IngestClientBase {
     static final String INGEST_PREFIX = "ingest-";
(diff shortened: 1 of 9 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientFactory.java
```diff
@@ -32,8 +32,8 @@ public static QueuedIngestClient createClient(ConnectionStringBuilder csb) throw
      * @return a new queued ingest client
      * @throws URISyntaxException if the connection string is invalid
      */
-    public static QueuedIngestClient createClient(ConnectionStringBuilder csb, @Nullable HttpClientProperties properties) throws URISyntaxException {
-        return new QueuedIngestClientImpl(csb, properties);
+    public static QueuedIngestClient createClient(ConnectionStringBuilder csb, @Nullable HttpClientProperties properties, boolean autoCorrectEndpoint) throws URISyntaxException {
```
AsafMah, line 35: These are all breaking changes, I'm afraid you'll have to do more overloads (cause java doesn't have default params)

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -92,21 +92,20 @@ class E2ETest {
...
     public static void setUp() throws IOException {
-        appKey = System.getenv("APP_KEY");
-        if (appKey == null) {
-            String secretPath = System.getProperty("SecretPath");
-            if (secretPath == null) {
-                throw new IllegalArgumentException("SecretPath is not set");
-            }
-            appKey = Files.readAllLines(Paths.get(secretPath)).get(0);
-        }
+//        appKey = System.getenv("APP_KEY");
```
AsafMah, line 95: Revert your changes in this file

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
```diff
@@ -50,14 +50,15 @@ public class QueuedIngestClientImpl extends IngestClientBase implements QueuedIn
...
     private final AzureStorageClient azureStorageClient;
 
-    QueuedIngestClientImpl(ConnectionStringBuilder csb, @Nullable HttpClientProperties properties) throws URISyntaxException {
-        this(csb, properties == null ? null : HttpClientFactory.create(properties));
+    QueuedIngestClientImpl(ConnectionStringBuilder csb, @Nullable HttpClientProperties properties, boolean autoCorrectEndpoint) throws URISyntaxException {
+        this(csb, properties == null ? null : HttpClientFactory.create(properties),autoCorrectEndpoint);
     }
 
-    QueuedIngestClientImpl(ConnectionStringBuilder csb, CloseableHttpClient httpClient) throws URISyntaxException {
+    QueuedIngestClientImpl(ConnectionStringBuilder csb, CloseableHttpClient httpClient, boolean autoCorrectEndpoint) throws URISyntaxException {
```
AsafMah, line 57: Where is the second half of the PR? (Adding checks to not replace the url in certain cases)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientFactory.java
```diff
@@ -22,7 +22,7 @@ private IngestClientFactory() {
      * @throws URISyntaxException if the connection string is invalid
      */
     public static QueuedIngestClient createClient(ConnectionStringBuilder csb) throws URISyntaxException {
-        return createClient(csb, (HttpClientProperties) null);
+        return createClient(csb, (HttpClientProperties) null, true);
```
AsafMah, line 25: Think of tests you can add in general

AsafMah (changes requested):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -138,27 +143,53 @@
...
+    public ManagedStreamingIngestClient(ConnectionStringBuilder ingestionEndpointConnectionStringBuilder,
+            ConnectionStringBuilder queryEndpointConnectionStringBuilder,
+            @Nullable HttpClientProperties properties, boolean autoCorrectEndpoint) throws URISyntaxException {
+        log.info("Creating a new ManagedStreamingIngestClient from connection strings");
+        queuedIngestClient = new QueuedIngestClientImpl(ingestionEndpointConnectionStringBuilder, properties, autoCorrectEndpoint);
+        streamingIngestClient = new StreamingIngestClient(queryEndpointConnectionStringBuilder, properties, autoCorrectEndpoint);
+        exponentialRetryTemplate = new ExponentialRetry(ATTEMPT_COUNT);
+    }
+
+    ManagedStreamingIngestClient(ConnectionStringBuilder connectionStringBuilder,
```
AsafMah, line 155: Why is this not public?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -138,27 +143,53 @@
...
+            @Nullable HttpClientProperties properties, boolean autoCorrectEndpoint) throws URISyntaxException {
+        log.info("Creating a new ManagedStreamingIngestClient from connection strings");
+        queuedIngestClient = new QueuedIngestClientImpl(connectionStringBuilder, properties, autoCorrectEndpoint);
+        streamingIngestClient = new StreamingIngestClient(connectionStringBuilder, properties, autoCorrectEndpoint);
+        exponentialRetryTemplate = new ExponentialRetry(ATTEMPT_COUNT);
+    }
+
+    public ManagedStreamingIngestClient(ConnectionStringBuilder connectionStringBuilder,
```
AsafMah, line 163: Docstring for all methods here

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -138,27 +143,53 @@
      * This constructor should only be used for advanced cases. If your endpoints are standard, or you do not know, use
      * {@link #ManagedStreamingIngestClient(ConnectionStringBuilder, HttpClientProperties)})} instead.
      */
+    public ManagedStreamingIngestClient(ConnectionStringBuilder ingestionEndpointConnectionStringBuilder,
```
AsafMah, line 146: Autocorrectendpoint is not documented

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -13,19 +17,33 @@
...
+        return isReservedHostname(clusterUrl) ? clusterUrl : clusterUrl.replaceFirst(INGEST_PREFIX, "");
     }
 
+    static boolean isReservedHostname(String rawUri) {
+        URI uri = URI.create(rawUri);
+        if (!uri.isAbsolute()) {
+            return true;
+        }
+        var host = uri.getHost().toLowerCase();
+        Pattern IPV4_PATTERN = Pattern.compile("^(25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)(\\.(25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)){3}$");
```
AsafMah, line 39: I told you not to import new libraries, but we already have com.apache.http.componenets, so you can use InetAddressUtils.isIPv6Address/InetAddressUtils.isIPv4Address

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -13,19 +17,33 @@
...
+
+        Pattern IPV6_STD_PATTERN = Pattern.compile("^(?:[0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}$");
+
+        Pattern IPV6_HEX_COMPRESSED_PATTERN = Pattern.compile("^((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)::((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)$");
+
+        var isIPFlag = IPV4_PATTERN.matcher(host).matches() || IPV6_STD_PATTERN.matcher(host).matches() || IPV6_HEX_COMPRESSED_PATTERN.matcher(host).matches();
+
+        return UriUtils.isLocalAddress(host) || isIPFlag || host.equalsIgnoreCase("onebox.dev.kusto.windows.net");
```
AsafMah, line 47: Not yours, but isLocalAddress can also be made better by using the standard java tools (InetAddress.getName().IsLoopback or something, though we still need to support "localhost")

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -386,6 +388,43 @@ private boolean canAuthenticate(ConnectionStringBuilder engineCsb) {
         return true;
     }
 
+    private void assertUrlCompare(String connectionDataSource, String clusterUrl, boolean autoCorrectEndpoint, boolean isQueued) {
```
AsafMah, line 391: Not nearly enough tests.
We need unit tests, with all edge cases (ip addresses, loopbacks, what not)

AsafMah (changes requested):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
ronmonetaMicro: There's this function, isn't it all together better?

https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#isLinkLocalAddress--

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -20,19 +26,34 @@ static boolean shouldCompress(CompressionType sourceCompressionType, IngestionPr
...
         }
+        return INGEST_PREFIX + clusterUrl;
     }
 
     static String getQueryEndpoint(String clusterUrl) {
-        if (clusterUrl.contains(INGEST_PREFIX)) {
-            return clusterUrl.replaceFirst(INGEST_PREFIX, "");
-        } else {
-            return clusterUrl;
+        return isReservedHostname(clusterUrl) ? clusterUrl : clusterUrl.replaceFirst(INGEST_PREFIX, "");
```
AsafMah, line 39: You don't check if it's null here?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -20,19 +26,34 @@ static boolean shouldCompress(CompressionType sourceCompressionType, IngestionPr
...
+    }
+
+    static boolean isReservedHostname(String rawUri) {
+        URI uri = URI.create(rawUri);
+        if (!uri.isAbsolute()) {
+            return true;
+        }
+        String authority = uri.getAuthority().toLowerCase();
+        boolean isIPFlag = InetAddressUtils.isIPv4Address(authority) || InetAddressUtils.isIPv6Address(authority);
+        boolean isLocalFlagg;
```
AsafMah, line 49: flagg

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -20,19 +26,34 @@ static boolean shouldCompress(CompressionType sourceCompressionType, IngestionPr
...
+        try {
```
AsafMah, line 50: Using exceptions for controlflow is the only way to do it?
Also, wouldn't this actually send a DNS request to any of the addresses it gets? if so, this is bad.

Just check against the string "localhost", it's good enough.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientFactory.java
```diff
@@ -171,13 +229,44 @@ public static ManagedStreamingIngestClient createManagedStreamingIngestClientFro
     public static ManagedStreamingIngestClient createManagedStreamingIngestClientFromDmCsb(ConnectionStringBuilder dmConnectionStringBuilder,
             @Nullable HttpClientProperties properties)
             throws URISyntaxException {
-        return ManagedStreamingIngestClient.fromDmConnectionString(dmConnectionStringBuilder, properties);
+        return ManagedStreamingIngestClient.fromDmConnectionString(dmConnectionStringBuilder, properties, true);
+    }
+
+    /**
```
AsafMah, line 235: Never make new code that is deprecated.
The whole point of deprecated is not supported.
Don't add any variants to deprecated code.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -126,40 +128,106 @@ public static ManagedStreamingIngestClient fromEngineConnectionString(Connection
...
      * @param properties                               - Additional properties to configure the http client
+     * @param autoCorrectEndpoint - Flag to indicate whether to correct the endpoint URI or not
      * @throws URISyntaxException if the connection string is invalid
      * @deprecated - This method is slated to be private.  Use
      * {@link IngestClientFactory#createManagedStreamingIngestClient(ConnectionStringBuilder, ConnectionStringBuilder, HttpClientProperties)} instead.
      * This constructor should only be used for advanced cases. If your endpoints are standard, or you do not know, use
      * {@link #ManagedStreamingIngestClient(ConnectionStringBuilder, HttpClientProperties)})} instead.
      */
+    public ManagedStreamingIngestClient(ConnectionStringBuilder ingestionEndpointConnectionStringBuilder,
+            ConnectionStringBuilder queryEndpointConnectionStringBuilder,
```
AsafMah, line 157: Again, you don't need all of these ctors (becuase of the deprecated factory methods), and if you do, prefer delegating to the main one.
Also don't make them public, we want people to use the factory.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -126,40 +128,106 @@ public static ManagedStreamingIngestClient fromEngineConnectionString(Connection
...
+     * @param connectionStringBuilder - Client connection string
+     * @param httpClient - HTTP client
+     * @param autoCorrectEndpoint - Flag to indicate whether to correct the endpoint URI or not
+     * @throws URISyntaxException if the connection string is invalid
+     */
+    public ManagedStreamingIngestClient(ConnectionStringBuilder connectionStringBuilder,
+            @Nullable CloseableHttpClient httpClient, boolean autoCorrectEndpoint) throws URISyntaxException {
+        log.info("Creating a new ManagedStreamingIngestClient from connection strings");
+        queuedIngestClient = new QueuedIngestClientImpl(connectionStringBuilder, httpClient, autoCorrectEndpoint);
+        streamingIngestClient = new StreamingIngestClient(connectionStringBuilder, httpClient);
```
AsafMah, line 189: forgot to pass here

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -429,6 +435,59 @@ private boolean canAuthenticate(ConnectionStringBuilder engineCsb) {
...
+        if (!autoCorrectEndpoint || clusterUrl.contains(INGEST_PREFIX) || isReservedHostname(clusterUrl)) {
```
AsafMah, line 439: Don't put logic like this in tests, it makes the test useless.
Put the desired result for each url.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -429,6 +435,59 @@ private boolean canAuthenticate(ConnectionStringBuilder engineCsb) {
...
+                assertEquals(clusterUrl, connectionDataSource);
+            }
+        } else {
+            String compareString = isQueued ? clusterUrl.replaceFirst(PROTOCOL_SUFFIX, PROTOCOL_SUFFIX + INGEST_PREFIX) : clusterUrl;
+            assertEquals(compareString, connectionDataSource);
+        }
+    }
+
+    private static Stream<Arguments> provideStringsForAutoCorrectEndpointTests() {
+        return Stream.of(
```
AsafMah, line 454: Positive examples? Edge cases?

Testing against System.getenv("ENGINE_CONNECTION_STRING") is not recommended, have just tests that pass

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -429,6 +435,59 @@ private boolean canAuthenticate(ConnectionStringBuilder engineCsb) {
...
+                Arguments.of("https://127.0.0.1"),
+                Arguments.of("https://localhost"),
+                Arguments.of("https://onebox.dev.kusto.windows.net"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("provideStringsForAutoCorrectEndpointTests")
+    void testQueuedUseAutoCorrectEndpoint(String csb) {
+        ConnectionStringBuilder engineCsb = ConnectionStringBuilder.createWithAadApplicationCredentials(csb, appId, appKey, tenantId);
+        try {
```
AsafMah, line 467: Why are these e2e tests anyway?
They don't belong here.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -429,6 +435,59 @@ private boolean canAuthenticate(ConnectionStringBuilder engineCsb) {
...
+            QueuedIngestClient queuedIngestClient = IngestClientFactory.createClient(engineCsb, null, true);
+            assertUrlCompare(((QueuedIngestClientImpl) queuedIngestClient).connectionDataSource, engineCsb.getClusterUrl(), true, true);
+            queuedIngestClient = IngestClientFactory.createClient(engineCsb, null, false);
+            assertUrlCompare(((QueuedIngestClientImpl) queuedIngestClient).connectionDataSource, engineCsb.getClusterUrl(), false, true);
```
AsafMah, line 471: Where are the tests for managedstreaming, which is the most complicated one?
Are you testing all the code you entered? all the new ctors?

AsafMah (changes requested):

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
ronmonetaMicro: WDYM?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
ronmonetaMicro: Not sure what you're referring to - these are edge cases which we discussed here on the PR
furthermore - all of these tests pass, so again not quite sure what you are referring to

AsafMah: I meant tests that pass the check, and have their url be autocorrected.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
AsafMah: You are remaking the logic of what you are testing inside E2ETest.
That makes the test only check if the logic is the same in both places, and not if it is correct.
Instead, compare each url against the expected result, hardcoded.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
ronmonetaMicro: so isn't System.getenv("ENGINE_CONNECTION_STRING") just that? or would you prefer it being a hardcoded URL instead?

AsafMah: There are many cases that do have a replacement but close to cases to aren't (for example - an address with a lot of numbers that isn't an ip address). 
You only test ONCE, on an env variable provided by the user, which could be anything, even one of the ones that is exempted.

Test against known urls, so you can know what the result should be.

Write at least 10 cases that do the substitution, with something different for each of them.

AsafMah (approved):

PR #297, opened 2023-04-20
makes AbstractSourceInfo.java public for use in bounded generic methods. Resolves #294

### LINKED ISSUE #294: Feature Request: Make AbstractSourceInfo "public"
I'm creating a wrapper library for a microservice framework and utilizing azure-kusto-java. The ingestion operations all have similar signatures, so I'd like to write a wrapper method that looks like this:

`public <T extends AbstractSourceInfo> ingestionMethod(T sourceInfo, IngestionProperties) {}`

Unfortunately, the AbstractSourceInfo class is not public, so I'm stuck writing a wrapper for each type of source info (4 types right now)

**Describe the solution you'd like**
Make the AbstractSourceInfo class public

**Describe alternatives you've considered**
Alternatively, the kusto library could do this work for me and write their own generic method to call the appropriate ingestion operation based on what sourceInfo is an instanceof (this is my intention with the wrapper). This is not super performant however it's an ingestion operation and a singular check, so I think the performance cost is worth the simplification for the user.

The-Funk:
Makes AbstractSourceInfo.java public for use in bounded generic methods.

Example:

`
public <T extends AbstractSourceInfo> ingest(T sourceInfo, IngestionProperties props) {
    if (sourceInfo instanceof FileSourceInfo.class) {
       // etc...
    }
}
`

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/AbstractSourceInfo.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/AbstractSourceInfo.java
@@ -5,7 +5,7 @@ package com.microsoft.azure.kusto.ingest.source;
 
 import java.util.UUID;
 
-abstract class AbstractSourceInfo implements SourceInfo {
+public abstract class AbstractSourceInfo implements SourceInfo {
 
     private UUID sourceId;
 


The-Funk:
I'm not sure I can link the issue to the PR when my branch is across forks but let me know if this is alright.

yogilad:
See the replay to your issue.
We do not think this change is required to accomplish what you need.

If you still feel this change is wanted, let us know.

The-Funk:
@yogilad I see, I think I glossed over that there was a public interface. Yes, that should work for my needs. Thank you! I will go ahead and close this.

PR #307, opened 2023-06-15
Guess who's back - Java 8 is back again - tell a friend 

ohadbitt:
Also updating bom version and msal

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,7 @@ jobs:
     name: Java ${{ matrix.java }}
     steps:
       - uses: actions/checkout@v2
-      - name: Setup java 8
+      - name: Setup java ${{ matrix.java }}
         uses: actions/setup-java@v3
         with:
           distribution: 'adopt'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,14 +8,17 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    name: Java 11
+    strategy:
+      matrix:
+        java: [ '8' ]
+    name: Java ${{ matrix.java }}
     steps:
       - uses: actions/checkout@v2
-      - name: Setup java 11
+      - name: Setup java 8
         uses: actions/setup-java@v3
         with:
           distribution: 'adopt'
-          java-version: 11
+          java-version: ${{ matrix.java }}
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -8,14 +8,17 @@ on:
...
+    strategy:
+      matrix:
+        java: [ '8' ]
+    name: Java ${{ matrix.java }}
     steps:
       - uses: actions/checkout@v2
-      - name: Setup java 11
+      - name: Setup java 8
         uses: actions/setup-java@v3
         with:
```
AsafMah, line 19: ${{ matrix.java }} if you're gonna use it

.travis.yml
```diff
@@ -1,3 +1,3 @@
 language: java
 jdk:
-  - oraclejdk11
+  - oraclejdk8
```
AsafMah, line 3: I don't think we need this file at all anymore (we don't use travisci)

pom.xml
```diff
@@ -32,19 +32,19 @@
     </developers>
 
     <properties>
-        <revision>4.0.4</revision> <!-- CHANGE THIS to adjust project version-->
+        <revision>4.0.5</revision> <!-- CHANGE THIS to adjust project version-->
```
AsafMah, line 35: This is atleast a 4.1.0 no?

data/src/main/java/com/microsoft/azure/kusto/data/ClientDetails.java
```diff
@@ -16,13 +16,13 @@
 public class ClientDetails {
 
     public static final String NONE = "[none]";
-    private static ConcurrentInitializer<String> defaultApplication = new LazyInitializer<>() {
+    private static ConcurrentInitializer<String> defaultApplication = new LazyInitializer() {
```
AsafMah, line 19: Pretty sure <> also works in java 8

AsafMah (commented):
Nothing major

data/src/main/java/com/microsoft/azure/kusto/data/ClientDetails.java
ohadbitt: it fails

pom.xml
ohadbitt: 5.0.0

PR #308, opened 2023-06-29
Queued Ingestion Client now supports retries

AsafMah:
Also exposed ResourceManager for spark.

Usage:

```java
ResourceHelper manager = ingestionClient.getResourceManager();
for (ContainerWithSas container = manager.getContainers()) {
// do your upload
boolean result = ...

manager.reportIngestionResult(container, result); // report the result in case of failure/success
}
```

One thing to note:

 

At first I named the interface "ResourceManager", and changed the name of the class to "ResourceManagerImpl".

This caused the diff to be huge and unreadable.

 

So for now I reverted the name of the class, and the interface is called "ResourceHelper". The idea is when the PR is approved I could change the name back.

Or if you have a better name for the interface, without changing the class name - please let me know


ag-ramachandran (approved):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public RankedStorageAccount(String accountName) {
+        this.accountName = accountName;
+    }
+
+    public void addResult(boolean success) {
+
+    }
+
+    public double getSuccessRate() {
```
yogilad, line 14: GetRank

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -346,16 +372,121 @@ protected String retrieveServiceType() {
...
 
+    @Override
+    public void reportIngestionResult(ResourceWithSas<?> resource, boolean success) {
+        if (storageAccountSet == null) {
+            log.warn("StorageAccountSet is null");
+        }
+        storageAccountSet.addResultToAccount(resource.getAccountName(), success);
+    }
+
+    private int calculateRetries(int size) {
```
ohadbitt, line 383: @yogilad  - i think we changed or are changing this in .Net - arent we?

ag-ramachandran (approved):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceHelper.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.microsoft.azure.kusto.ingest.exceptions.IngestionClientException;
+import com.microsoft.azure.kusto.ingest.exceptions.IngestionServiceException;
+import com.microsoft.azure.kusto.ingest.utils.ContainerWithSas;
+import com.microsoft.azure.kusto.ingest.utils.ResourceWithSas;
+
+import java.util.List;
+
+public interface ResourceHelper {
+    List<ContainerWithSas> getContainers() throws IngestionClientException, IngestionServiceException;
```
yogilad, line 11: Need rename and doc string

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
yogilad: Though I still prefer get-rank

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccountSet.java
```diff
@@ -0,0 +1,60 @@
...
+    @Nullable
+    public RankedStorageAccount getAccount(String accountName) {
+        return accounts.get(accountName);
+    }
+
+    public <T> List<T> getShuffledResources(Map<String, List<T>> resources) {
+        List<RankedStorageAccount> accounts = getShuffledAccounts();
+
+        // Get valid resources for each account
+        List<List<T>> validResources = accounts.stream().map(x -> resources.get(x.getAccountName())).filter(r -> r != null && !r.isEmpty())
```
yogilad, line 46: I have no idea what this works.
Also, this seems like a logic which ought to sit in ResourceManager

ingest/src/main/java/com/microsoft/azure/kusto/ingest/utils/ContainerWithSas.java
```diff
@@ -32,7 +32,18 @@ public BlobContainerClient getContainer() {
         return container;
     }
 
-    public String getEndpoint() {
-        return container.getBlobContainerUrl() + sas;
+    @Override
+    public String getEndpointWithoutSas() {
```
yogilad, line 36: Is there a ToString which does this as well?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/utils/QueueWithSas.java
```diff
@@ -35,4 +36,19 @@ public QueueClient getQueue() {
     public String getEndpoint() {
         return queue.getQueueUrl() + sas;
     }
+
+    @Override
+    public String getEndpointWithoutSas() {
```
yogilad, line 41: Same as ContainersWithSas

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -346,16 +372,121 @@ protected String retrieveServiceType() {
...
+        }, "ResourceManager.uploadLocalFileWithRetries");
+    }
+
+    String uploadLocalFileWithRetries(AzureStorageClient azureStorageClient, File file, String blobName, boolean shouldCompress)
+            throws IngestionClientException, IngestionServiceException {
+        return resourceActionWithRetries(getContainers(), container -> {
+            azureStorageClient.uploadLocalFileToBlob(file, blobName, container.getContainer(), shouldCompress);
+            return (container.getContainer().getBlobContainerUrl() + "/" + blobName + container.getSas());
+        }, "ResourceManager.uploadLocalFileWithRetries");
+    }
```
yogilad, line 447: This ought to be a _utils/IngestionUtils.java_ API which takes the resource manager in or a shuffled resource list and does the work.
Here it just clusters up the file.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -346,16 +372,121 @@ protected String retrieveServiceType() {
...
+            log.warn("StorageAccountSet is null");
+        }
+        storageAccountSet.addResultToAccount(resource.getAccountName(), success);
+    }
+
+    private int calculateRetries(int size) {
+        return size < 3 ? size : size / 2 + 1; // Calculation is based on dotnet
+    }
+
+    private <TInner, TWrapper extends ResourceWithSas<TInner>, TOut> TOut resourceActionWithRetries(List<TWrapper> resources,
```
yogilad, line 387: Same as below comment

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
yogilad: Yes. Should be default 3 with policy override

ingest/src/main/java/com/microsoft/azure/kusto/ingest/utils/ContainerWithSas.java
AsafMah: Why? explicit methods are better

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccountSet.java
AsafMah: There is a comment that explains it. 
I don't know how to make it simpler -> it loops over the shuffled account, for each one it looks up a resource in the resources map, and returns the valid ones.

Also, in c# the logic lives there.
You didn't want much logic in ResourceManager anyway.

Plus, this is just the most dead simple logic (just pick the valid ones).
When we will have a more complicated logic (taking into account the rank), it will be needed to be done here anyway.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceHelper.java
AsafMah: Done

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
AsafMah: Done

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
AsafMah: Done

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
```diff
@@ -0,0 +1,89 @@
...
+            List<TWrapper> resources, FunctionOneException<TOut, TWrapper, Exception> action, String actionName)
+            throws IngestionClientException {
+
+        if (resources.isEmpty()) {
+            throw new IngestionClientException("No resources found");
+        }
+
+        for (int i = 0; i < RETRY_COUNT; i++) {
+            if (i >= resources.size()) {
+                throw new IngestionClientException("Retry count exceeded");
```
yogilad, line 36: The problem with this exception is that it is contextless. (which operation, which resource are we trying to act on, etc.)
Pass the context message to the exception.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
```diff
@@ -0,0 +1,89 @@
...
+    private static <TInner, TWrapper extends ResourceWithSas<TInner>, TOut> TOut resourceActionWithRetries(ResourceManager resourceManager,
+            List<TWrapper> resources, FunctionOneException<TOut, TWrapper, Exception> action, String actionName)
+            throws IngestionClientException {
+
+        if (resources.isEmpty()) {
+            throw new IngestionClientException("No resources found");
+        }
+
+        for (int i = 0; i < RETRY_COUNT; i++) {
+            if (i >= resources.size()) {
```
yogilad, line 35: 1) You can run on max(resources, retry_count)
2) suggest you loop over none the less

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
```diff
@@ -0,0 +1,89 @@
...
+                throw new IngestionClientException("Retry count exceeded");
+            }
+            TWrapper resource = resources.get(i);
+            try {
+
```
yogilad, line 40: nit: misplaced new line

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
```diff
@@ -0,0 +1,89 @@
...
+                        resourceManager.reportIngestionResult(resource, true);
+                        return result;
+                    } catch (Exception e) {
+                        span.addException(e);
+                        throw e;
+                    }
+                }, actionName, attributes);
+            } catch (Exception e) {
+                log.warn(String.format("Error during retry %d of %d for %s", i + 1, RETRY_COUNT, actionName), e);
+                resourceManager.reportIngestionResult(resource, false);
```
yogilad, line 58: nit: move this to the inner exception for clarity

yogilad, line 58: Any built in backoff?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
```diff
@@ -0,0 +1,89 @@
...
+                throw new IngestionClientException("Retry count exceeded");
+            }
+            TWrapper resource = resources.get(i);
+            try {
+
+                Map<String, String> attributes = new HashMap<>();
+                attributes.put("resource", resource.getEndpointWithoutSas());
+                attributes.put("account", resource.getAccountName());
+                attributes.put("type", resource.getClass().getName());
+                attributes.put("retry", String.valueOf(i));
```
yogilad, line 45: Should this not include some context about the resource posted to queue or uploaded to container?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
```diff
@@ -234,19 +215,15 @@ protected IngestionResult ingestFromStreamImpl(StreamSourceInfo streamSourceInfo
...
-            BlobContainerClient blobContainerClient = container.getContainer();
-            // trace uploadStreamToBlob
-            Map<String, String> attributes = new HashMap<>();
-            attributes.put("container", blobContainerClient.getBlobContainerName());
-            MonitoredActivity.invoke((SupplierTwoExceptions<Void, IOException, URISyntaxException>) () -> {
-                azureStorageClient.uploadStreamToBlob(streamSourceInfo.getStream(), blobName,
-                        blobContainerClient, shouldCompress);
-                return null;
-            }, getClientType().concat(".uploadStreamToBlob"), attributes);
-            String blobPath = blobContainerClient.getBlobContainerUrl() + "/" + blobName + container.getSas();
```
yogilad, line 247: Note you seem to loose some context in this implementation (may need to be passed down to the API)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/utils/ContainerWithSas.java
yogilad: Because if you print this class it will print the SAS

ingest/src/main/java/com/microsoft/azure/kusto/ingest/utils/StreamUtils.java
```diff
@@ -0,0 +1,18 @@
+package com.microsoft.azure.kusto.ingest.utils;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class StreamUtils {
```
yogilad, line 10: Why the separate class (sounds like a resource manager capability)?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceHelper.java
```diff
@@ -0,0 +1,25 @@
+package com.microsoft.azure.kusto.ingest;
+
+import com.microsoft.azure.kusto.ingest.exceptions.IngestionClientException;
+import com.microsoft.azure.kusto.ingest.exceptions.IngestionServiceException;
+import com.microsoft.azure.kusto.ingest.resources.ContainerWithSas;
+import com.microsoft.azure.kusto.ingest.resources.ResourceWithSas;
+
+import java.util.List;
+
+public interface ResourceHelper {
```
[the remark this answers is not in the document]
AsafMah: Done

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
AsafMah: Added context

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
AsafMah: Added context

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
AsafMah: Added some context

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
AsafMah: Fixed

ingest/src/main/java/com/microsoft/azure/kusto/ingest/utils/StreamUtils.java
AsafMah: Fixed

ingest/src/main/java/com/microsoft/azure/kusto/ingest/AzureStorageClient.java
```diff
@@ -27,7 +27,7 @@
 import java.nio.file.Files;
 import java.util.zip.GZIPOutputStream;
 
-class AzureStorageClient {
+public class AzureStorageClient {
```
yihezkel, line 30: If making this class public so it can be used, I suggest adding javadocs for the class and the API (`public`) methods.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -375,4 +374,14 @@ public void close() throws IOException {
         queuedIngestClient.close();
         streamingIngestClient.close();
     }
+
+    @Override
```
yihezkel, line 378: Maybe I just need to understand the design better, but I'm not used to seeing both composition (`QueuedIngestClientImpl this.queuedIngestClient`) and inheritance (`implements QueuedIngestClient`) of the such similar classes.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+    private ResourceAlgorithms() {
+    }
+
+    private static <TInner, TWrapper extends ResourceWithSas<TInner>, TOut> TOut resourceActionWithRetries(ResourceManager resourceManager,
+            List<TWrapper> resources, FunctionOneException<TOut, TWrapper, Exception> action, String actionName, Map<String, String> additionalAttributes)
+            throws IngestionClientException {
+
+        if (resources.isEmpty()) {
```
yihezkel, line 35: Can you use Apache Commons' `ArrayUtils`?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
```diff
@@ -46,13 +30,15 @@
...
 import java.net.URISyntaxException;
 import java.time.Instant;
-import java.util.*;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.UUID;
 
 public class QueuedIngestClientImpl extends IngestClientBase implements QueuedIngestClient {
 
+    public static final String QUEUED_INGEST_CLIENT_IMPL = "QueuedIngestClientImpl";
```
yihezkel, line 39: We don't really need to do it like this, do we?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -375,4 +374,14 @@ public void close() throws IOException {
...
+    public void setQueueRequestOptions(RequestRetryOptions queueRequestOptions) {
+        queuedIngestClient.setQueueRequestOptions(queueRequestOptions);
+    }
+
+    @Override
+    public IngestionResourceManager getResourceManager() {
```
yihezkel, line 384: Though I don't recall the description 100%, I think my comment is going to be related:
There's a bit of inconsistency with naming here, and I think we should get it right in this 1 PR. if it'll mess up the diff, maybe that can be the last step after PR approval?

1) If there's a class that implements an interface, the former should be more specific than the latter. Yet the naming implies the opposite: `class ResourceManager implements IngestionResourceManager`.
2) The type of this class' member is `IngestionResourceManager`, yet its name is `resourceManager`.
3) The name of this getter should align with both.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
```diff
@@ -0,0 +1,136 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.InputStream;
+import java.lang.invoke.MethodHandles;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class ResourceAlgorithms {
```
yihezkel, line 24: I'm sure I'm just missing something about the rationale for this design, but as far as I can tell, `ResourceAlgorithms` provides methods that I would expect to be a part of either `ResourceManager`, or some parent or inherited class that exposes the subset of methods that should be exposed publicly.

I generally regard (effectively) static classes as a smell - `resourceActionWithRetries`, `postToQueueWithRetries`, `uploadStreamToBlobWithRetries` and `uploadLocalFileWithRetries` accept a `ResourceManager`, and based on these methods' names, at first glance I'd expect them to be called thusly, for example: `resourceManager.postToQueueWithRetries(...)`

As for the remaining methods, `roundRobinNestedList`, `getShuffledResources` and `groupResourceByAccountName`, I'd expect them to be in a `ResourceGroup<T>` class.

Again, I assume I'm misunderstanding something, but at first glance the design of this class is a smell to me.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -38,60 +41,33 @@
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
-class ResourceManager implements Closeable {
+public class ResourceManager implements Closeable, IngestionResourceManager {
+    public static final String SERVICE_TYPE_COLUMN_NAME = "ServiceType";
+    private static final long REFRESH_INGESTION_RESOURCES_PERIOD = 1000L * 60 * 60; // 1 hour
```
yihezkel, line 46: All of these would IMHO be more readable as:
`private static final Long REFRESH_INGESTION_RESOURCES_PERIOD_MILLIS = TimeUnit.HOURS.toMillis(1);`

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccountSet.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public RankedStorageAccountSet() {
+        this.accounts = new HashMap<>();
+    }
+
+    public void addResultToAccount(String accountName, boolean success) {
+        RankedStorageAccount account = accounts.get(accountName);
+        if (account != null) {
+            account.addResult(success);
+        }
```
yihezkel, line 22: No action (e.g. exception, trace) needed if `accountName` wasn't found?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class RankedStorageAccount {
+    private final String accountName;
+
+    public RankedStorageAccount(String accountName) {
+        this.accountName = accountName;
+    }
+
+    public void addResult(boolean success) {
+
```
yihezkel, line 11: Was this mistakenly unfinished?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    public double getRank() {
+        return 1;
```
yihezkel, line 15: Was this mistakenly unfinished?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -346,16 +358,55 @@ protected String retrieveServiceType() {
...
+            throw new IllegalArgumentException(resourceTypeName);
+        }
+
+        String getResourceTypeName() {
+            return resourceTypeName;
+        }
+    }
+
     private static class IngestionResource<T> {
+        final ResourceType resourceType;
```
yihezkel, line 397: Is `T` not the `ResourceType`?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/AzureStorageClient.java
AsafMah: It's not really meant for the consumption of users, it's only public because it has to be used in a subpackage
Do you have another idea.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: https://en.wikipedia.org/wiki/Delegation_pattern

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
AsafMah: Out of scope for this PR.

This name is for tracing, it's either that, doing reflection every time, or hardcoding the name in the function.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: I planned on changed the name after the PR ended to avoid having a huge diff.
Yochai suggested this name as a solution, but I still agree with you, I'll see what we can do.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
AsafMah: They were, @yogilad wanted me to move them because they "bloated the ResourceManager class".

I'm not sure what to do when people have radically different ideas of arranging code.

Maybe we can all discuss together.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
AsafMah: I think this api is ugly but it's probably better.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
AsafMah: This PR is just about setting the pieces, since we didn't decide the logic yet.

For now, the shuffling of accounts is completely random.

The idea is to decide on an alogirthm in the future without needing to redo the code.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
AsafMah: It looks like this entire class is now useless with the new algorithm.

In fact there's seem to be a lot of dead code in ResourceManager, i'll remove it. Good catch.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/AzureStorageClient.java
yihezkel: Way out of scope, but one way I've done this is to have an api package with everything that's fit for public use being `public`, + a core package with classes like this.

Resolving :-)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
yihezkel: Yeah, OOS, but too ugly for me to ignore :-)

`this.getClass().getSimpleName()` is cleaner.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
AsafMah: Upon second read - this class is actually important. ResourceType is an enum with the name of the resource, and exists here to map between the name (we get from kusto) to the resource itself. T is the actual resource stored in the array. We can't really change it without a full re-write

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
AsafMah: What's the advantage over using a native method?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccountSet.java
AsafMah: Fixed

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
AsafMah: Done

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: Fixed as we said

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccountSet.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Nullable
+    public RankedStorageAccount getAccount(String accountName) {
+        return accounts.get(accountName);
+    }
+
+    @NotNull
+    public List<RankedStorageAccount> getRankedShuffledAccounts() {
+        List<RankedStorageAccount> accounts = new ArrayList<>(this.accounts.values());
```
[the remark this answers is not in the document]
AsafMah: This PR is only for the infrastructure and to provide an interface - the actual alogrithm is not chosen yet.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public static <T extends ResourceWithSas<?>> List<T> getShuffledResources(List<RankedStorageAccount> shuffledAccounts, List<T> resourceOfType) {
+        Map<String, List<T>> accountToResourcesMap = groupResourceByAccountName(resourceOfType);
+
+        List<List<T>> validResources = shuffledAccounts.stream()
+                // For each shuffled account, get the resources for that account
+                .map(account -> accountToResourcesMap.get(account.getAccountName()))
+                // Remove nulls and empty lists
+                .filter(resourceList -> resourceList != null && !resourceList.isEmpty())
```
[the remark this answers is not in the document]
AsafMah: These are all of the accounts we have, and we select one type of resource from them - so if we select queues, we could have an account which only has containers, so it will be empty. Null I don't think they can, but worth checking if we're already checking empty.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
yihezkel: Includes `null` check (not sure if relevant here)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
yihezkel, line 123: For example here, the `ArrayUtils`' `null` check would help.

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/ArrayUtils.html#isNotEmpty-T:A-

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -367,12 +366,22 @@ protected IngestionResult ingestFromStreamImpl(StreamSourceInfo streamSourceInfo
 
     @Override
     protected String getClientType() {
-        return MANAGED_STREAMING_INGEST_CLIENT;
+        return CLASS_NAME;
```
yihezkel, line 369: Can't the parent class define this, without the need to `override`, since it's always going to return `CLASS_NAME``?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
yihezkel: Maybe add a comment to this effect, though.

yihezkel (approved):
Approved, with a few suggestions

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
AsafMah: Arrayutils doesn't work since it's a list, and we currently don't have the ListUtils dependency.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
AsafMah: See from before

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: You can't inherit static types, so it will have to be non-static, and the string will have to be created every time.
Since it uses reflection for the class name, it's wasteful to do every time, so I'm leaving it as is.

PR #310, opened 2023-07-13
fixed jackson transient serialization bug and handled HTTP status 204 in HTTPResponseWrapper

tanmaya-panda1:
### Fixed
1. Handling bug from Jackson: Bug from Jackson(https://github.com/FasterXML/jackson-databind/issues/3948) After upgrading jackson-databind 2.13+, properties were being exposed after serialization that was set to @JsonIngore and shouldn't be.
2. In HTTP, a response with a status code of 204 (No Content) indicates that the request was successful, but there is no content to return in the response body. Since there is no response body, there won't be an HTTP entity associated with the response and hence response.getEntity() returns IllegalArgumentException(HTTP Response entity doesn't exist) when the HTTP response code is 204 when the ingestionReportMethod is set to TABLE. As a fix handled HTTP status code 204 empty response handling in the existing HTTPResponseWrapper of azure-kusto-sdk.

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/HttpResponseWrapper.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/HttpResponseWrapper.java
@@ -83,9 +83,9 @@ public class HttpResponseWrapper extends HttpResponse implements IHttpResponse {
 
         if (body == null) {
             try {
-                if(response.getStatusLine().getStatusCode() == HttpStatus.SC_NO_CONTENT){
+                if (response.getStatusLine().getStatusCode() == HttpStatus.SC_NO_CONTENT) {
                     body = new byte[0];
-                }else{
+                } else {
                     body = EntityUtils.toByteArray(response.getEntity());
                 }
             } catch (IOException ignored) {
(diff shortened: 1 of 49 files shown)


AsafMah (approved):
Great!

data/src/main/java/com/microsoft/azure/kusto/data/auth/HttpResponseWrapper.java
```diff
@@ -82,7 +83,11 @@ public Mono<byte[]> getBodyAsByteArray() {
 
         if (body == null) {
             try {
-                body = EntityUtils.toByteArray(response.getEntity());
+                if (response.getStatusLine().getStatusCode() == HttpStatus.SC_NO_CONTENT) {
+                    body = new byte[0];
```
ag-ramachandran, line 87: Minor nit, do you want to a constant EMPTY_BYTE_ARRAY

ag-ramachandran (approved):

data/src/main/java/com/microsoft/azure/kusto/data/auth/HttpResponseWrapper.java
tanmaya-panda1: agreed Ram, EMPTY_BYTE_ARRAY is a cleaner way of handling empty bye array.

PR #312, opened 2023-07-23
Dont use timer if closed

ohadbitt:


yogilad:
Can you please add a description?
This is also so the change list for the version can be easily composed

yogilad (approved):

AsafMah (approved):

AsafMah (changes requested):
Add to changelog.md

ohadbitt:
> This is also so the change list for the version c

The description wouldve been a copy of the title

data/src/main/java/com/microsoft/azure/kusto/data/auth/PublicAppTokenProviderBase.java
```diff
@@ -28,6 +28,11 @@ public abstract class PublicAppTokenProviderBase extends MsalTokenProviderBase {
         super(clusterUrl, authorityId, httpClient);
     }
 
+    @Override
+    protected synchronized String acquireAccessTokenImpl() throws DataServiceException, DataClientException {
```
AsafMah, line 32: Add a comment explaining this.

AsafMah (approved):
Approved - just add the comment, and make sure it actually does what it should (not making the thing pop up twice)

PR #328, opened 2023-09-28
Fix method name capitalization and other non-functional changes

yihezkel:
### Fixed
Method name capitalization and other non-functional changes


AsafMah (approved):
Awesome effort!

AsafMah (changes requested):
Needs that method fixed

AsafMah (approved):

PR #331, opened 2023-11-21
Transient IOError

ohadbitt:
### Fixed
* IOException is now checked against known subclasses that should be retried or checked for string "time out" to decide for permanency
 Following exception was raised and considered permanent
shaded.msdataflow.com.microsoft.azure.kusto.data.exceptions.DataServiceException: IOError when trying to retrieve CloudInfo
      at shaded.msdataflow.com.microsoft.azure.kusto.data.auth.CloudInfo.retrieveCloudInfoForCluster(CloudInfo.java:122)
..
Caused by: java.net.ConnectException: Connection timed out (Connection timed out)
..
Caused by: org.apache.http.conn.HttpHostConnectException: Connect to ingest-pfclientquality.centralus.kusto.windows.net:443 [ingest-pfclientquality.centralus.kusto.windows.net/52.185.106.107] failed: Connection timed out (Connection timed out)

### Aadded
* Retry cloud info fetch 

* Split Utils to HttpPostUtils and Utils, so it now makes sense this class has the post() methods... 
 Wanted to move all http related classes to a new http package as i found it too crowded - but couldnt move HttpClientProperties as its used  by client and therefore a breaking change, so its half way now and added todo to finish on next major


data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -127,8 +127,10 @@ public static CloudInfo retrieveCloudInfoForCluster(String clusterUrl,
                         ((Closeable) localHttpClient).close();
                     }
                 }
-            } catch (IOException | URISyntaxException ex) {
-                throw new DataServiceException(clusterUrl, "IOError when trying to retrieve CloudInfo", ex, true);
+            }catch (URISyntaxException e) {
+                throw new DataServiceException(clusterUrl, "URISyntaxException when trying to retrieve CloudInfo:" + e.getMessage(), e, true);
+            } catch (IOException ex) {
+                throw new DataServiceException(clusterUrl, "IOException when trying to retrieve CloudInfo:" + ex.getMessage(), ex, true);
```
AsafMah, line 133: Why is IOExeception permanent here but not before?

ohadbitt: wierd - not pushed right

ingest/pom.xml
```diff
@@ -272,11 +272,6 @@
             <artifactId>resilience4j-retry</artifactId>
             <version>${resilience4j.version}</version>
         </dependency>
-        <dependency>
```
AsafMah, line 275: How do we know it's ok to remove?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -90,49 +92,69 @@ public static CloudInfo retrieveCloudInfoForCluster(String clusterUrl,
                 return cloudInfo;
             }
 
-            CloudInfo result;
+            for (int i = 0;; i++) {
```
AsafMah, line 95: The "ExponentialRetry" class did not fit here?

ingest/pom.xml
ohadbitt: it was here only because the verify phase complained

AsafMah (approved):

PR #332, opened 2023-11-21
Test more java versions

AsafMah:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,7 @@
         <profile>
             <id>java8</id>
             <activation>
-                <jdk>[1.8,)</jdk>
+                <jdk>[1.8,11)</jdk>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -112,6 +123,7 @@
                 <artifactId>formatter-maven-plugin</artifactId>
                 <version>2.16.0</version>
                 <configuration>
+                    <!--suppress UnresolvedMavenProperty -->
```
yihezkel, line 126: What's this for?
I assume there's an unresolved maven prop, but which one and why would we include any such property?

AsafMah: The one on the next line - jetbrains complains

ohadbitt (approved):

PR #333, opened 2023-11-28
bumps dependencies and removes vulnerability caused by apache httpclient

The-Funk:
Excluded vulnerable commons-codec version from old Apache client and overrode with new version
Moved Jackson to 2.15.2, current version supported by Databricks 14.2 GA
Moved slf4j back to 1.7.36 (newer and better maintained than 1.8 series)
Bump other dependencies within minor version release train

### DIFF
--- a/data/pom.xml
+++ b/data/pom.xml
@@ -169,6 +169,10 @@
                     <artifactId>httpcore</artifactId>
                     <groupId>org.apache.httpcomponents</groupId>
                 </exclusion>
+                <exclusion>
+                    <artifactId>commons-codec</artifactId>
+                    <groupId>org.apache.commons</groupId>
+                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
@@ -201,6 +205,11 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-text</artifactId>
             <version>${commons-text.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+            <version>${commons-codec.version}</version>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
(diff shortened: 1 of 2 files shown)


yogilad:
@ag-ramachandran , can you have a look at this contrinuted version bump PR?
Does this affect our compatibility with Synapse, Databricks, Fabric Spark?

ohadbitt:
> @ag-ramachandran , can you have a look at this contrinuted version bump PR? Does this affect our compatibility with Synapse, Databricks, Fabric Spark?

I'm OK with it - Jackson 2.15.2 also updated in latest ADB

ohadbitt (approved):

ohadbitt:
Seems like you need to run the formatter plugin

The-Funk:
> Seems like you need to run the formatter plugin

Can do! How do I run the plugin?

AsafMah:
@The-Funk 
If you have intellij you can access it from the maven menu:
image


Otherwise you can run (with the same maven you run other commands):
```bash
mvn formatter:format
``

The-Funk:
I ran the formatter plugin however git is indicating no changes. The Java 8 build above is failing due to a dependency issue in the data module however the only thing that was changed directly in the data module was to use the newer commons-codec to replace the one that had the vulnerability.

I will look at the output of the build above a little more closely. The project compiles on my system and all the tests except E2E pass.

yogilad:
Hi @ag-ramachandran , 

Would this PR cause any conflict with one or more of the Spark connector flavors?
If it does not, we'd like to approve it?
Otherwise, let's talk.

ag-ramachandran:
The version looks okay @yogilad , it should be compatible

yihezkel:
@ohadbitt @AsafMah
This PR seems to be triggering maven-dependency-plugin failures on:
Unused declared dependencies found: commons-codec:commons-codec:jar:1.16.0:compile

Are you sure it's being used in kusto-data?

In the meantime, I'm going to remove it in a PR I'm working on (https://github.com/Azure/azure-kusto-java/pull/341), and if you aren't aware of why we might need it, I'll proceed with merging that PR with the dependency removed.

The-Funk:
@yihezkel it was used in kusto-ingest previously but might not be used anymore at all.

The Apache HTTP Client brings in an old version of it as a transitive dependency, so I had explicitly replaced it because the old version has some associated CVEs.

PR #335, opened 2023-12-10
Removes apache http libraries and replace with azure core httpclient

The-Funk:
This PR is ready for review. Updated Notes 3/2/2024.

**Notes:**

- The removal of the entity stuff was difficult as expected. There was a lot of use of Entity and a lot of use of the URIBuilder from Apache core. I simplified this to some extent and I'd like to open a new PR in the future to remove the usage of URIBuilder completely.

- The old Apache Client comes with an exception called ParseException that was being used to indicate parsing failures. The version of said exception from the V4 client library, extends RuntimeException. To replace it I created a KustoParseException that extends RuntimeException for the sake of compatibility. This should work as a drop in replacement. **After refactoring, this class is only used in TESTS now.**

- Rather than using the Netty HttpClient implementation from core, I used the HttpClient interface itself. This way, if someone wants to replace the Netty client they can do so easily by adding a different classname to their GlobalConfig or excluding the Netty client jar from their dependencies and replacing it with their own jar. For instance, I have a personal interest in utilizing the [azure-core-http-vertx](https://mvnrepository.com/artifact/com.azure/azure-core-http-vertx) client once it comes out of Beta. The vertx client better supports native image building with GraalVM. The Netty client implementation is the default implementation included with azure core. **We should avoid using anything with Netty specifically in the name in order to avoid becoming locked to the Netty client.**

- Moved all request building logic (some from HttpPostUtils and some from ClientImpl) to a request builder class in order to provide better testability and separation of concerns and reduce the number of parameters that need passed between methods.

- Moved all HTTP Request Tracing parameters to a specific builder class for better testability, separation of concerns, etc, same as above.

- Moved remainder of HttpPostUtils into a BaseClient class that the ClientImpl extends, removing the need to pass the HTTP client to a separate class and promoting long lived HTTP Clients.

- I deleted the HttpResponseWrapper. Since the Apache client is removed, this class should no longer be needed.

**BREAKING CHANGES:**

- In the ClientRequestProperties class the proxy option used HttpHost from the Apache libraries directly. This is exposed directly to the users of the library, so in order to remove the Apache dependencies I needed to remove HttpHost. I replaced this with the ProxyOptions from Azure core but this would be a breaking change for users of the library that specify a custom proxy. They would need to rework a few lines of code, no more than 3 or 4 max.
- HttpClientFactory now accepts any client implementing azure-core's HttpClient interface instead of an Apache CloseableHttpClient.

**In Tests:**

- The UtilitiesTest in the data module made use of BasicHttpResponse from the Apache libraries. Because the HTTPResponse class from Azure core is an abstract class, creating new instances of it manually is not simple. This is probably by design. But in the event that we want to create a fake HTTPResponse, for instance in tests, we need a way to do so. There is a utility class in Tests now that will allow for the creation of fake HttpResponses. I created a simple implementation for it including only what had been used previously, and a builder to make it easier to instantiate new instances.

- CloudInfo tests now use the RequestBuilder


ohadbitt:
Hey 
Sounds like a very well thought process really like the usage of HttpClient - looks like a great fit.
- Reactor is pretty much the same as mutiny (not familiar with the latter but it looks pretty similar approach ) but i could help with overcoming any problems you got trying to use it - you intend that we overcome the Todos in this PR ? As my intentions were that the Async part will be handled in the next PR and now only have simple sync impl - please tell me which part would you like me to take in this PR i would do it first thing this week or the next.
- ProxyOptions is a very good implementation that we want anyway.
- Regarding closeable - IngestClient has to be closeable, and only the STreamingIngestClient will have it as a noOp

The-Funk:
@ohadbitt 

You're right about the closeable.

I like your plan of a separate PR to work on the async implementation. That's a lot more changes.

There's also one or two other design things that I think might make sense while we're performing surgery around the HTTP Clients. I think it would be nice to get rid of HttpPostUtils and move the post methods onto an abstract class that all of the KustoClients should extend. That way the enclosed HTTP client can be a protected property of the parent abstract class, as well as the post methods. That would mean less passing around of the HTTP client, meaning shorter method signatures. I saw there was also a comment about making the base URL string a class property that can be set one time upon KustoClient creation and then reused. I liked that idea as well.

ohadbitt:
@The-Funk Sounds good - maybe better @AsafMah will look as well at all the planning

AsafMah:
I'm not sure about KustoParseException extending RuntimeException, don't we want our exceptions to be checked?

ohadbitt:
I too am against RuntimeExceptions - @The-Funk , sounds like you should go ahead with the other changes, when ready hit the "ready for review"

The-Funk:
Speaking of unchecked exceptions, I had extra throws for that exception that I didn't need, I just pushed that.

In this case, I think KustoParseException made sense because the original ParseException that needed replaced was unchecked, so to get the same behavior without changing the method signatures, I replaced it with a new unchecked exception.

If you would like I could use the new Apache HTTP core's ParseException which is a checked exception in V5.

Setting ready for review now.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -67,7 +65,7 @@ public class ResourceManager implements Closeable, IngestionResourceManager {
     private IngestionResource<QueueWithSas> successfulIngestionsQueues;
     private IngestionResource<QueueWithSas> failedIngestionsQueues;
 
-    ResourceManager(Client client, long defaultRefreshTime, long refreshTimeOnFailure, @Nullable CloseableHttpClient httpClient) {
+    ResourceManager(Client client, long defaultRefreshTime, long refreshTimeOnFailure, @Nullable HttpClient httpClient) {
         this.defaultRefreshTime = defaultRefreshTime;
         this.refreshTimeOnFailure = refreshTimeOnFailure;
         this.client = client;
```
ohadbitt, line 71: No need the HttpClientWrapper(httpClient) here  :)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
```diff
@@ -183,6 +184,7 @@ protected IngestionResult ingestFromFileImpl(FileSourceInfo fileSourceInfo, Inge
         } catch (IOException e) {
             throw new IngestionClientException("Failed to ingest from file", e);
         } catch (IngestionServiceException e) {
+            // Todo: Determine if this is needed since the exception is immediately rethrown w/ no type change
```
ohadbitt, line 187: can remove

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpPostUtils.java
```diff
@@ -194,65 +189,67 @@ public static DataServiceException createExceptionFromResponse(String url, HttpR
...
+            if (httpResponse != null) {
+                httpResponse.close();
             }
         }
     }
 
     private static boolean shouldPostToOriginalUrlDueToRedirect(int redirectCount, int status) {
-        return (status == HttpStatus.SC_MOVED_TEMPORARILY || status == HttpStatus.SC_TEMPORARY_REDIRECT)
-                && redirectCount + 1 <= MAX_REDIRECT_COUNT;
+        return (status == 302 || status == 307) && redirectCount + 1 <= MAX_REDIRECT_COUNT;
```
ohadbitt, line 202: here too- consts ate still available

ohadbitt (commented):
Went through it all. Main issue is testing this properly, we need some long query, big data, multi threaded stress tests to really know we are ok here as we never had such a change. Will you take this as well ?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -3,32 +3,30 @@
...
-import org.apache.http.HttpHeaders;
-import org.apache.http.ParseException;
-import org.apache.http.client.utils.URIBuilder;
-import org.apache.http.entity.AbstractHttpEntity;
-import org.apache.http.entity.ContentType;
-import org.apache.http.entity.InputStreamEntity;
-import org.apache.http.entity.StringEntity;
-import org.apache.http.impl.client.CloseableHttpClient;
+
+import org.apache.hc.core5.http.ContentType;
```
AsafMah, line 25: Why do we need to have this dependency? Isn't the whole point is to get rid of the apache http server?

I don't think (from what I've seen in the code) we need to use the "Entity" abstraction - the BinaryData abstraction in the Azure package should be good enough.

And URIBuilder, we can't achieve this otherwise?

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
```diff
@@ -29,63 +22,23 @@ public class HttpClientFactory {
...
-    /**
-     * A custom connection keep-alive strategy that uses the server instructions set in the {@code Keep-Alive}
-     * response header; if the response doesn't contain a {@code Keep-Alive} header, the client will use a configurable
-     * keep-alive period.
-     */
-    static class CustomConnectionKeepAliveStrategy implements ConnectionKeepAliveStrategy {
-
-        private final int defaultKeepAlive;
+        // Is the per route connection maximum needed?
+        // Todo: missing keepalive and per-route connections, is keepalive handled by default in the azure clients?
```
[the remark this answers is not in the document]
The-Funk: I spent a lot of time looking at ConnectionProvider and think it would be a good addition for further HTTPClient customization.

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpPostUtils.java
```diff
@@ -45,51 +38,49 @@ private HttpPostUtils() {
...
+        HttpRequest request = setupHttpPostRequest(url, body, headers);
+        int requestTimeout = timeoutMs > Integer.MAX_VALUE ? Integer.MAX_VALUE : Math.toIntExact(timeoutMs);
+
+        // Todo: Handle custom socket timeout settings
+        // RequestConfig requestConfig = RequestConfig.custom().setSocketTimeout(requestTimeout).build();
+        // request.setConfig(requestConfig);
+
+        // Execute and get the response
+        // Fixme: Send requests asynchronously instead of synchronously
+        try (HttpResponse response = httpClient.sendSync(request, new Context(new Object(), null))) {
```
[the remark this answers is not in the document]
The-Funk: Resolved

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -122,41 +118,45 @@ private static CloudInfo fetchImpl(String clusterUrl, @Nullable HttpClient given
...
                     if (errorFromResponse.isEmpty()) {
-                        errorFromResponse = response.getStatusLine().getReasonPhrase();
+                        // Fixme: Missing reason phrase to add to exception. Potentially want to use an enum.
+                        errorFromResponse = "";
                     }
-                    throw new DataServiceException(clusterUrl,
-                            "Error in metadata endpoint, got code: " + statusCode + "\nWith error: " + errorFromResponse,
-                            statusCode != HttpStatus.SC_TOO_MANY_REQUESTS);
+                    throw new DataServiceException(clusterUrl, "Error in metadata endpoint, got code: " + statusCode +
+                            "\nWith error: " + errorFromResponse, statusCode != 429);
```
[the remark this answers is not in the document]
The-Funk: The original const values that were used were from the Apache library but if we want to look into removing Apache HttpCore as well the HttpClient, we'll need to use constants from somewhere else. 

Those same const values from Apache are provided in another 3rd party library that the Kusto library uses (nimbus) but I don't think it would be a good idea to rely on that dependency being around forever. If Nimbus removes Apache or nimbus needs removed from Kusto dependencies we'll run into the same problem again.

So what I've done is create my own HttpStatus class with public static final int values. I put this class under the http package in data.

The-Funk:
@ohadbitt  For the const values, I created a specific class called HttpStatus in the http package and put const values in it.

The reason is, the other status codes constants come from the Apache library, and if we want to remove even the core library eventually, we'll need to migrate away from the Apache codes.

Those same status Apache status codes are provided as a transitive dependency of another dependency in the project (nimbus) but it probably wouldn't be a good idea to rely on them being there since we wouldn't be including the codes directly anymore if we get rid of Apache HttpCore.

The-Funk:
> Went through it all. Main issue is testing this properly, we need some long query, big data, multi threaded stress tests to really know we are ok here as we never had such a change. Will you take this as well ?

I have a development cluster that I can run some long running queries against. I can also try some test ingestions as well. I usually ingest data as JSON though. The part that worries me is ingestions involving file sources and streams.

For scale I can use JMeter and compare/contrast my service RAM/CPU with the current version of the library. 

I don't know if I can give you really good in-depth insights though. That might be beyond me.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
The-Funk: @AsafMah I removed references to entity and replaced it with BinaryData from the Azure Core library. The URIBuilder is a little tricky. I will need to dig into how it is implemented before I can remove it, but I will do that next.

I know this is supposed to be a minimal implementation, but there's another issue related to HTTP request generation that I'd like to ask for permission to explore.

I'd really like to see the post methods exist on the KustoClient itself, that way the post methods could just use the KustoClient's inner HTTP client without the HTTP client needing passed around to the methods. This would help shorten method signatures.

I'd also really like to create an HttpRequestBuilder class and move the request building logic out of HttpPostUtils and out of the client implementations and into a request builder class. 

This would greatly shorten the post method signatures on HttpPostUtils and move request building logic out of the clients and into a class dedicated for that purpose, providing more logical separation of that phase.

e.g.

Phase 1: Parse/Validate Input and Generate an HTTP Request (builder class, separately testable)
Phase 2: Send request (monitored activity using KustoClient's inner HttpClient)
Phase 3: Parse/Validate Response and convert to KustoResult (potentially a separate builder?)

In this setup the post methods take a ready-made Request object, the Request object is then sent via the overarching client abstraction's inner HTTP client, and then the response is parsed.

With your and your team's permission I'd really like to explore that possibility in this PR. I think it would really help readability for newcomers to the codebase. I would of course adjust tests to account for the change.

The-Funk: I started working on the above just to see what it would do with the codebase and it looks really really nice and has a lot of positive impacts in the client class. I haven't committed any of it. If you wanted I could open a separate branch, but I feel **much** more confident about following HTTP request building logic with everything in one place. I'd really like to add this.

The-Funk:
@AsafMah @ohadbitt 

I left a comment above about my most recent push in response to Asaf. 

I can revert it if you do not like the changes.

I was getting very confused where everything was happening and wanted to break things down into phases.

To that end, there's a LOT of changes above, but I think they're very useful changes in terms of code readability.

I know this PR was originally only supposed to be a replacement of the HTTPClient but there was a lot of fragmentation in where headers were being set, where the request URI was being edited, etc, and I think now that problem is mostly solved.

There's still some scoping issues to hide things from the users of the library but that would involve moving files and I wanted to try to avoid that.

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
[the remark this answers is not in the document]
The-Funk: I spent some time looking into Keep-Alive. I believe we can set it per request using two specific headers.

I put some notes in HttpClientFactory regarding this specific property.

ohadbitt (commented):
So regarding the tests - i would take this on me and you can continue either here if its harder or in a new PR with the Async implementation

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
```diff
@@ -35,8 +41,23 @@ public static HttpClient create(HttpClientProperties providedProperties) {
             options.setProxyOptions(properties.getProxy());
         }
 
-        // Is the per route connection maximum needed?
-        // Todo: missing keepalive and per-route connections, is keepalive handled by default in the azure clients?
+        // Is the per route connection maximum needed anymore?
+
+        // Todo: missing keepalive and per-route connections.
+        // NOTE: Keep-Alive is provided by 2 headers that can be placed on any HTTP request or response.
+        // It looks like one of the headers, Connection: Keep-Alive, is already configured when setting up tracing
```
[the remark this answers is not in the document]
The-Funk: Thank you, the testing is definitely the part that I think would be harder for me to do on my side. I appreciate the help.

The tracing I was talking about, I think, was added by @enmoed 

There's some specific headers added for what looks like MDC (Mapped Diagnostic Context). There's a user, an app, an activity, and a request id. The request Id allows for an HTTP request to be traced from source to destination and back. Usually this is done for being able to trace a request through service logs. If you get a request with id SOME_GUID and it hits your API gateway, you can then check the logs for SOME_GUID to see that the gateway processed the request and then passed it on to its destination. Once the request reaches its destination, the request Id is then written into a response header, usually something like "x-correlation-id", indicating that you've finished the request phase, the request reached its destination and was processed, and now you can track the response on its way back to the requesting client via that correlation header. 

When adding these headers, the Connection header was also being set to "Keep-Alive", and because these HTTP tracing details were being added to every request, the Connection header was getting set on every request. However, as you mentioned, we can't really rely on Keep-Alive being set on the client because the server can just choose to ignore client Keep-Alive parameters if it's configured to do so.

For that reason, I don't know if we need to continue setting Keep-Alive details on the client-side however if we do want to continue setting those parameters, we could add the second Keep-Alive header in the request builder, so that every request gets Keep-Alive configured. We could also add a set of headers to the BaseClient which every request should integrate. Connection and Keep-Alive could be 2 of these universally applied headers.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
The-Funk: @ohadbitt you want me to use the HttpClientWrapper here?

The-Funk:
Updated the changelog in the initial PR description.

The-Funk:
I've marked this for review. For the record there are 2 failing tests. I've tracked the issue down to the ConnectionStringBuilder. Something in there doesn't like the colon : in certain URIs and replaces them with %3A.

I have yet to find the exact cause.

The-Funk:
@ohadbitt @AsafMah I was able to get this to finally build and all tests to pass! Including the ones that check weird URIs! 

The issue was the Apache URIBuilder in their new library.

I don't know if the behavior in the HTTPClient 4 URI builder is correct or the behavior in Apache HTTP5 URI builder is correct but using the older URIBuilder all tests pass and all results match what we expect. 

Again, I have never seen a URI like the one above, and I am positive I don't know my protocols and standards to the degree the Apache Devs do.

The-Funk:
@AsafMah @ohadbitt @yihezkel is there still interest in this? 

I've been aligning this with new PRs as they come in but don't want to keep working on it if there's no chance it will be merged. I haven't heard from y'all in a little while.

ohadbitt:
Hi @The-Funk 
Sorry for taking a lot of time , hurt my foot badly , couldnt walk - then was working on testing in the back.
To our matter - This looks all looks well but is failing for me now - ill try now and see why
com.microsoft.azure.kusto.data.exceptions.DataServiceException: IOException when trying to retrieve cluster metadata:Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens
 at [Source: (String)"\u001F�\u0008\u0000\u0000\u0000\u0000\u0000\u0004\u0000}R]O�0\u0014}G�?Ty�i�8\u001F�-�Lc�ijA{��5�H��v�\u0018��N\u0009Pi�%J��{�����l��\u007F\u0007\u000B�29�=��P�1�J_j�\u001B�}�'\u001B�{w>�w��n\u0015��\u0019���mr�\u0011�]�\u0015�\u001E�\u0005\u0011Вv\u000E���༹�\u0014h���U�'�UU.x�\u0016\\u0002\u0090qD\u001B\��Ъ��dP6o\u0004\u001F\u0016�@\u0004
�﬚�\u001Co��\u0018�\u000F!k�;�a\u0005�\u000C�Þy��\u0010�J��G��yt���
_�u�'���\u001D��X��l7�\u001Euf.\u001B.@\u0016\u0019�E\u0016�1L\u0011)\u0000�R�u\u0006$�˒&ѽ�ѣD�o�a��Eg\u00061e�\u001E�T\u007F"���\:�V@�m\u001C��%�!B1$d!�\u000Cr����CD��\u00084B���\u0014��}
>���\u0011��}��\u001F\u000C\u001E�\u0013O\u007F��'�E��\u0010<�5c\u0008s�@�d\u0015**�\u0008`Ij��OD��_�oٴBXp.
�\u0012\u001E���9��#\u0002G�v����(��(/3�pEj�pU#X����\u00027�܇vz��\u000F����G\u0003\u0000\u0000"; line: 1, column: 2]
	at com.microsoft.azure.kusto.data.auth.CloudInfo.lambda$retrieveCloudInfoForCluster$0(CloudInfo.java:105) ~[kusto-ingest-6.0.0-jar-with-de

The-Funk:
@ohadbitt   Sorry to hear about your foot, I hope you are doing better now. Those uXXXX numbers are unicode. I am wondering if this is a charset related issue. Looking at the changes I made in the CloudInfo file, there's no Accept-Charset header and there's no charset defined in the Accept header either. I am wondering if you just changed line 124 to:

```java
request.setHeader(HttpHeaderName.ACCEPT, "application/json;charset=UTF-8");
```

I'm not sure if that would fix it. Being external I don't have the secret key to run the E2ETest so it always fails for me and I'm kind of blind to issues there. If I could help more, I would.

I'm seeing in the build failures that the missing applicationId is what is causing the E2ETest to fail, and I assume that comes from this CloudInfo endpoint.

ohadbitt:
will push a fix

ohadbitt:
I see some different behavior here , we sometimes get "handshake timed out after 10000ms"
seems like its a different timeout  configuration then the timeout we use - maybe wise to configure value for that as well if the user provided client timeout 
https://github.com/reactor/reactor-netty/issues/1617

The-Funk:
> I see some different behavior here , we sometimes get "handshake timed out after 10000ms" seems like its a different timeout configuration then the timeout we use - maybe wise to configure value for that as well if the user provided client timeout . It seems the cause is just too many connections at a given time - solution is configuring cmd arg reactor.netty.ioWorkerCount or by making a fixed connection pool. Anyway user can configure this from outside or limit his concurrent requests. [reactor/reactor-netty#1617](https://github.com/reactor/reactor-netty/issues/1617)

Hey @ohadbitt is this an observation or something you'd like to see action taken to address? If you would like, I can add more options to the HttpClientProperties and HttpClientPropertiesBuilder to account for more of the options available in azure-core's HttpClientOptions.

Alternatively, or perhaps longer term in a separate PR, do you think it would be worthwhile to do away with the HttpClientProperties and HttpClientPropertiesBuilder in this project and instead expose the ability to the users to directly provide an instance of HTTPClientOptions from azure-core? That would be a breaking change for anyone providing custom HTTP options currently, but it would mean that you wouldn't have to update the HttpClientProperties class in the future. As it is, HttpClientProperties will just be acting as a wrapper around the azure-core HttpClientOptions, so doing away with it would just remove the middleman.

By the way, related to HTTP Tracing properties. I noticed HttpClientOptions in azure-core provides for that as well. They have various implementations there including open-telemetry implementations. It may very well be that the HttpTracing class I added and the work done to add telemetry here could be replaced by the azure-core implementations eventually, to further simplify things in this project. That's just an aside though, work for a different PR on a different day.

ohadbitt:
Hey @The-Funk - regarding your question - i think the properties are good as they are - the user can already give the client he wants to use - if he want to really customize it .

Regarding my perf tests - im actually uploading the jmeter file here 
https://github.com/Azure/azure-kusto-java/pull/354
I ran it in Azure Load Test resource with 2 instances , these are 5 tests that are running simultaneously
It seems to be showing a little worse performance - showing here version 5.0.4 vs funk pr  :)
I'm still thinking what to do next - but this doesn't necessarily mean we don't want to make the change as we can't really say if 
Kusto is not the one making the jitter.

percentile_ms_20 | percentile_ms_80 | percentile_ms_90 | percentile_ms_95 | count_ | avg_ms | t | testRun | Type
-- | -- | -- | -- | -- | -- | -- | -- | --
3,308 | 3,862 | 4,094 | 4,365 | 797 | 3821| funk | 3 | streaming query successfully
3,059 | 3,596 | 3,776 | 3,909 | 791 | 3332 | 504 | 3 | streaming query successfully
1,834 | 2,226 | 2,356 | 2,573 | 158 | 2051 | funk | 3 | streaming ingest successfully
1,705 | 2,019 | 2,133 | 2,222 | 160 | 1,878 | 504 | 3 | streaming ingest successfully
3,262 | 4,038 | 4,243 | 4,578 | 794 | 3862 | funk | 3 | simple query successfully
2,892 | 3,689 | 3,879 | 4,084 | 791 | 3412 | 504 | 3 | simple query successfully
721 | 1,059 | 1,161 | 1,733 | 158 | 1,077 | funk | 3 | queuing upload ingest successfully
797 | 991 | 1,092 | 1,149 | 160 | 1,056 | 504 | 3 | queuing upload ingest successfully
546 | 851 | 1,018 | 1,183 | 397 | 760 | funk | 3 | queue and update table '16' times succesfully
518 | 711 | 752 | 829 | 400 | 668 | 504 | 3 | queue and update table '16' times succesfully

The-Funk:
@ohadbitt I think the slightly worse performance is to be expected. I think the default azure-core client, Netty, is slightly slower than AHC because under the hood it uses NIO for async and there's some not-insignificant extra work there, per request, to support that async io. 

Whereas the AHC just spawns a thread per connection, up to a certain number of threads, which is a much simpler design, albeit it uses significantly more heap space. I think Netty doesn't start to shine until you ramp up the number of concurrent requests being made or the size of the data in the requests. [I saw this on medium about that a while back, and the sources were pretty good.](https://medium.com/@saswatraj/under-the-hood-of-http-clients-32a68279c5dd)

ohadbitt:
@The-Funk  
After some more changes and tests runs that also included using OKHttp client instead of netty (its basically user configureable)- results seems to be a little better and we agreed that we are ok with the now slight worse perf 
@AsafMah still wants to review this but after you (or us ?? )  resolve the merge from master we can go through with this change :)

The-Funk:
@ohadbitt @AsafMah I've updated the branch, let me know if there's anything in particular you'd like to see after review.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -114,54 +110,59 @@ public static CloudInfo retrieveCloudInfoForCluster(String clusterUrl,
...
     }
 
     private static CloudInfo fetchImpl(String clusterUrl, @Nullable HttpClient givenHttpClient) throws URISyntaxException, IOException, DataServiceException {
         CloudInfo result;
         HttpClient localHttpClient = givenHttpClient == null ? HttpClientFactory.create(null) : givenHttpClient;
         try {
-            HttpGet request = new HttpGet(UriUtils.appendPathToUri(clusterUrl, METADATA_ENDPOINT));
-            request.addHeader(HttpHeaders.ACCEPT_ENCODING, "gzip,deflate");
-            request.addHeader(HttpHeaders.ACCEPT, "application/json");
+            HttpRequest request = new HttpRequest(HttpMethod.GET, UriUtils.appendPathToUri(clusterUrl, METADATA_ENDPOINT));
```
AsafMah, line 120: We talked about which client we should use as the default.
Seems like out of our options, OKHttp has a huge advantage - it can automatically decompress "gzip", saving us this code.

It doesn't support "deflate" by default, but this implementation doesn't seem to either - so we should remove it from the accept encoding anyway.

If we do keep this though, it should really be somewhere where the logic can be re-used

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
```diff
@@ -23,69 +17,52 @@ public class HttpClientFactory {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(HttpClientFactory.class);
 
+    // Todo: for team consideration...
+    // With previous versions, this was configurable per request, now it's only configurable per client
+    // Because there is only 1 HTTP Client per KustoClient (ClientImpl) abstraction, this can only be configured universally for the instance
+    // For now I've put this value here as a placeholder, however this should be added to ClientOptions
+    private static final Long TIMEOUT = Duration.ofMinutes(10).toMillis();
```
AsafMah, line 24: We can actually still create a client per client, using the Builder instead of the Provider, and I think we may want that?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -66,21 +65,21 @@ class ResourceManager implements Closeable, IngestionResourceManager {
...
+    public ResourceManager(Client client, long defaultRefreshTime, long refreshTimeOnFailure, @Nullable HttpClient httpClient) {
         this.defaultRefreshTime = defaultRefreshTime;
         this.refreshTimeOnFailure = refreshTimeOnFailure;
         this.client = client;
         timer = new Timer(true);
         // Using ctor with client so that the dependency is used
         this.httpClient = httpClient == null
-                ? new NettyAsyncHttpClientBuilder().responseTimeout(Duration.ofMinutes(UPLOAD_TIMEOUT_MINUTES)).build()
-                : new HttpClientWrapper(httpClient);
+                ? new NettyAsyncHttpClientProvider().createInstance(new HttpClientOptions().setResponseTimeout(Duration.ofDays(UPLOAD_TIMEOUT_MINUTES)))
```
AsafMah, line 75: Instead of this, why not the default provider, or using the methods from data?

AsafMah (commented):
Some comments/thoughts

data/pom.xml
```diff
@@ -159,26 +159,10 @@
...
-                    <artifactId>commons-codec</artifactId>
-                    <groupId>org.apache.commons</groupId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.httpcomponents</groupId>
-            <artifactId>httpcore</artifactId>
-            <version>${httpcore.version}</version>
+            <version>${apache.httpclient.version}</version>
```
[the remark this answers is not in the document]
The-Funk: @AsafMah another reviewer asked if we could remove those lines in a different PR. Probably worth excluding them still. I will put those lines back.

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -0,0 +1,189 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.invoke.MethodHandles;
+import java.util.Optional;
+import java.util.zip.DeflaterInputStream;
+import java.util.zip.GZIPInputStream;
+
+public abstract class BaseClient implements Client, StreamingClient {
```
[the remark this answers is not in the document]
The-Funk: This was mainly to get rid of Utils and store the underlying HTTPClient in the "Kusto Client" so that we could stop passing the HTTPClient around in the code so much and so often. I made the class abstract so that there could be multiple Client implementations over time. 

If there's no plan to have other implementations we could move the logic from this class into the ClientImpl. I am wondering if we will want an AsyncClient implementation at some point or just add methods with async (flux/mono) signatures to the existing client implementation. If you're preference is the former, my recommendation would be that we keep it, if it's the latter my recommendation would be to move the logic from this abstract class into ClientImpl, and get rid of this class.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
The-Funk: I agree. I think we should avoid using specific providers in code and use default since default should fall back to the consumer's choice if Netty is explicitly excluded in the consumer's pom file.

data/pom.xml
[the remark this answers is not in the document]
The-Funk: [Readded in this commit](https://github.com/Azure/azure-kusto-java/pull/335/commits/c25960feb5c5a0ba84b3744564aa7fb2a5b262a3)

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
The-Funk: Moved this into the builder per request. [See this commit.](https://github.com/Azure/azure-kusto-java/pull/335/commits/abae5fb150f99e7e8ab5925a9b81fc73ad80ad78)

The-Funk: Added client provider to the builder while I was in there as well, although if you are open to the idea, I would like to discuss deprecating the HttpClientProperties class in the future. [See this comment.](https://github.com/Azure/azure-kusto-java/pull/335#issuecomment-1944673160)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
The-Funk: [Removed here. Also set timeout to 10 minutes, not 10 days. Not sure if that was intentional or not.](https://github.com/Azure/azure-kusto-java/pull/335/commits/286bbe10c5026acb825b609791de00ea42404f93)

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
The-Funk: It should be easy to swap in OkHttp at this point. All that should be needed now is to edit the pom file to exclude azure-core-http-netty and then add the azure-core-http-okhttp package. 👍

data/src/main/java/com/microsoft/azure/kusto/data/Client.java
```diff
@@ -27,6 +27,7 @@ public interface Client extends Closeable {
 
     KustoOperationResult executeMgmt(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException;
 
+    // Use this if the response is ignored or if you can use the result as unstructured string
     String executeToJsonResult(String database) throws DataServiceException, DataClientException;
```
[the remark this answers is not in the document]
The-Funk: Remove in [this commit.](https://github.com/Azure/azure-kusto-java/pull/335/commits/d9dc27c94003180ded71e30dff49a11d64e8b87e)

data/src/main/java/com/microsoft/azure/kusto/data/ClientFactory.java
```diff
@@ -43,12 +44,12 @@ public static Client createClient(ConnectionStringBuilder csb, HttpClientPropert
...
      *
      * @param csb the connection string builder
-     * @param client CloseableHttpClient client. It will not be closed when {@link Client#close} is called.
+     * @param client HttpClient client.
      * @return a fully constructed {@linkplain Client} instance
      * @throws URISyntaxException if the cluster URL is invalid
      */
-    public static Client createClient(ConnectionStringBuilder csb, CloseableHttpClient client) throws URISyntaxException {
-        return client == null ? createClient(csb, (HttpClientProperties) null) : new ClientImpl(csb, client, true);
+    public static Client createClient(ConnectionStringBuilder csb, HttpClient client) throws URISyntaxException {
```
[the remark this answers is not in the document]
The-Funk: [Finished in this commit.](https://learn.microsoft.com/en-us/java/api/com.azure.core.util.httpclientoptions?view=azure-java-stable)

The-Funk:
@AsafMah I think I addressed most points. All tests that I can run appear to be passing. I have not loaded in OkHttp but if you would like me to remove Netty and add OkHttp now, I can.

data/pom.xml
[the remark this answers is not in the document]
The-Funk: Reverted removing core in [this commit](https://github.com/Azure/azure-kusto-java/pull/335/commits/350dc1bb9ea59b11ba863c4f6c901ba04e55feb7) because the URIBuilder uses utilities out of core. Kept the exclusion of commons-codec though for better security posture.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
The-Funk: Removed deflate in [this commit.](https://github.com/Azure/azure-kusto-java/pull/335/commits/9370530077b2a12aa51279d3b7f4f491b63ef6f7)

ohadbitt (approved):
Lets go :)
@cole  Are you going to make the tests pass or should I?

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -0,0 +1,188 @@
...
+        try {
+
+            // Todo: Implement async version of this method
+
+            httpResponse = httpClient.sendSync(request, Context.NONE);
+
+            int responseStatusCode = httpResponse.getStatusCode();
+
+            if (responseStatusCode == HttpStatus.OK) {
+                InputStream contentStream = new EofSensorInputStream(new CloseParentResourcesStream(httpResponse), null);
```
AsafMah, line 82: Do we still need these two stream wrappers, even with the current http client?

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    protected String post(HttpRequest request) throws DataServiceException {
+
+        // Todo: Add async version of this method
+
+        // Execute and get the response
+        try (HttpResponse response = httpClient.sendSync(request, Context.NONE)) {
+            Optional<HttpHeader> contentEncoding = Optional.ofNullable(response.getHeaders().get(HttpHeaderName.CONTENT_ENCODING));
+            String responseBody = contentEncoding.isPresent() && (contentEncoding.get().getValue().contains("gzip")) ? response.getBodyAsInputStream()
```
AsafMah, line 43: Thought about it, I see we're handling gzip/deflate headers 3 times in 3 different ways (here, next method, cloudinfo). Let's extract this handling to a common code.

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
```diff
@@ -24,68 +21,49 @@ public class HttpClientFactory {
...
-     * Creates a new Apache HTTP client.
+     * Creates a new HTTP client.
      *
      * @param providedProperties custom HTTP client properties
-     * @return a new Apache HTTP client
+     * @return a new HTTP client
      */
-    public static CloseableHttpClient create(HttpClientProperties providedProperties) {
-        LOGGER.info("Creating new CloseableHttpClient client");
+    public static HttpClient create(HttpClientProperties providedProperties) {
```
AsafMah, line 29: The code in httpclientprovider can re-use the client if it has "null" as clientProperties.

Does it make sense to send null to the builder if providedProperties is null?

AsafMah (commented):
Let's resolve these and finally close this.

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
The-Funk: @AsafMah I am not sure. I kept them in place because I couldn't answer that myself. Would tests cover this if I changed it?

The-Funk: @AsafMah removed in [this commit](https://github.com/Azure/azure-kusto-java/pull/335/commits/54d1ec5d6f07ae58cc30711169f016f3de2a2fa1)

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
The-Funk: @AsafMah I just checked and I carried this over from 5.0.5 so the behavior is the same as in v5.

Currently, if we send null to the builder, then the builder returns some sensible default values for most properties. I think that makes sense but I also think the HttpClient in azure-core does the same thing. If you and your team want to control the defaults then it makes sense to call the builder and have it return its default values. If you'd rather let azure-core decide the defaults, then it makes sense to skip the builder and just go straight to returning a client with no properties.

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
The-Funk: @ohadbitt this was in postToStreamingOutput, so we'll probably want to double triple check that streaming queries are okay. I verified that all tests I can run still pass.

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
The-Funk: It's not perfect but these 3 should now be a little more aligned and I created and documented two static methods in Utils for identifying gzip compressed responses and getting specific stream implementations from a response stream.
Commit is[ here.](https://github.com/Azure/azure-kusto-java/pull/335/commits/5b6e2f7802384d4fd1363a443b087dbbdbff79d6)

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
The-Funk: @AsafMah I submitted a commit for this that just lets azure-core decide defaults. This makes us less dependent on the builder in this codebase which makes us less dependent on our wrapper. We may want to run the JMeter tests again to see what those defaults do to response times etc.
[Commit is here.](https://github.com/Azure/azure-kusto-java/pull/335/commits/5d3a6ed12e2ce53c7f3221910168ad4182b67c00)

The-Funk:
FYI @AsafMah @ohadbitt I fully expect that with this big of a change there will come challenges and potentially regressions. I am willing to help if issues arise.

AsafMah (approved):
Thank you for your patience, approved!

AsafMah:
@The-Funk 

What's next from here?

Are you going to work on the async api?

ohadbitt:
Thank you @The-Funk for your contribution, 
As Asaf mentioned- you did want to impl the async api as well, if its not relevant anymore do tell :)

The-Funk:
> @The-Funk
> 
> What's next from here?
> 
> Are you going to work on the async api?

Indeed!

PR #339, opened 2024-01-15
Throw KustoServiceQueryError when Kusto service result includes OneApiError(s)

yihezkel:
### Changed
Minor enhancements regarding reuse, added tests

### Fixed
Throw KustoServiceQueryError when Kusto service result includes OneApiError(s)


data/src/main/java/com/microsoft/azure/kusto/data/KustoOperationResult.java
```diff
@@ -18,6 +18,8 @@
 import java.lang.invoke.MethodHandles;
 import java.util.*;
 
+import static com.microsoft.azure.kusto.data.KustoResultSetTable.EXCEPTIONS_MESSAGE;
```
ohadbitt, line 21: please don't import static

data/src/main/java/com/microsoft/azure/kusto/data/KustoOperationResult.java
```diff
@@ -122,7 +128,24 @@ private void createFromV2Response(String response) throws KustoServiceQueryError
             if (jsonNode.isArray()) {
                 jsonArray = (ArrayNode) jsonNode;
                 for (JsonNode node : jsonArray) {
-                    if (node.has(FRAME_TYPE_PROPERTY_NAME) && node.get(FRAME_TYPE_PROPERTY_NAME).asText().equals(DATA_TABLE_FRAME_TYPE_PROPERTY_NAME)) {
+                    if (node.has(FRAME_TYPE_PROPERTY_NAME) && node.get(FRAME_TYPE_PROPERTY_NAME).asText().equals(DATA_SET_COMPLETION_FRAME_TYPE_PROPERTY_VALUE)
```
ohadbitt, line 131: Can do this once
if (node.has(FRAME_TYPE_PROPERTY_NAME) TableType = node.get(FRAME_TYPE_PROPERTY_NAME)

ohadbitt (approved):
Wow did not know that is an option .. is it new ?

data/src/main/java/com/microsoft/azure/kusto/data/KustoOperationResult.java
yihezkel: Even for constants? Why not?

yihezkel:
> Looking good Wow did not know that is an option .. is it new ?

I don't think it's new. Also, at least some consumers already throw an exception if the query response object includes a `OneApiError` - C# SDK (though they implement it a little differently) and Kusto Explorer. My next step is to check the other SDKs.

PR #340, opened 2024-01-16
Resource manager can return an empty list

ohadbitt:
Also added test for so it does not return old storage


ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -61,13 +60,9 @@ public class ResourceManager implements Closeable, IngestionResourceManager {
...
     private String identityToken;
-    private IngestionResource<ContainerWithSas> containers;
-    private IngestionResource<TableWithSas> statusTable;
-    private IngestionResource<QueueWithSas> queues;
-    private IngestionResource<QueueWithSas> successfulIngestionsQueues;
-    private IngestionResource<QueueWithSas> failedIngestionsQueues;
+    private IngestionResourceSet ingestionResourceSet;
 
-    ResourceManager(Client client, long defaultRefreshTime, long refreshTimeOnFailure, @Nullable CloseableHttpClient httpClient) {
+    public ResourceManager(Client client, long defaultRefreshTime, long refreshTimeOnFailure, @Nullable CloseableHttpClient httpClient) {
```
AsafMah, line 65: Why are you making it public? this means we are responsible to keep it not broken

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -409,4 +403,12 @@ boolean empty() {
             return this.resourcesList.size() == 0;
         }
     }
+
+    private static class IngestionResourceSet {
```
AsafMah, line 407: To be clear - the problem we had before is that we replaced the queues one at a time, and this fixes at by replacing them all at once?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
ohadbitt: for some tests
will revert

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
ohadbitt: Not exactly 
we were setting this.containers for example to a new empty list 
while it was not lock protected and could be read empty at this point and even not reverted back on failure too ...

AsafMah (approved):

PR #342, opened 2024-01-17
Add missing headers

yihezkel:
### Changed
- Include "charset=utf-8" in HttpHeaders.CONTENT_TYPE because that's best practice
- Remove "Fed" header, because it's to indicate "AAD federated authentication", but authentication should be dependent on the KCSB (see when FEDERATED_SECURITY_SUFFIX is stripped in com.microsoft.azure.kusto.data.ClientImpl.ClientImpl(ConnectionStringBuilder, CloseableHttpClient, boolean))


data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -461,8 +461,7 @@ private String generateCommandPayload(String database, String command, ClientReq
     }
 
     private void addCommandHeaders(Map<String, String> headers) {
-        headers.put(HttpHeaders.CONTENT_TYPE, "application/json");
-        headers.put("Fed", "True");
```
yihezkel, line 465: Am I mistaken that this is to indicate "AAD federated authentication", but authentication should be dependent on the KCSB?

AsafMah:
Why did you add the query_language header? we don't see it anywhere

yihezkel:
> Why did you add the query_language header? we don't see it used anywhere else

I see from the traces many clients that send this. But I don't think it helps anything, so I'm removing it for now.

AsafMah (approved):

PR #344, opened 2024-01-18
Fix excessive test runtime

yihezkel:
### Fixed
Fix excessive test runtime


ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -66,6 +69,8 @@ public class ResourceManager implements Closeable, IngestionResourceManager {
     private IngestionResource<QueueWithSas> queues;
     private IngestionResource<QueueWithSas> successfulIngestionsQueues;
     private IngestionResource<QueueWithSas> failedIngestionsQueues;
+    private TimerTask refreshIngestionAuthTokenTask;
+    private TimerTask refreshIngestionResourcesTask;
 
     ResourceManager(Client client, long defaultRefreshTime, long refreshTimeOnFailure, @Nullable CloseableHttpClient httpClient) {
```
yihezkel, line 75: We're putting a `CloseableHttpClient` in a wrapper that isn't closeable (`new HttpClientWrapper(httpClient)`)!
Now we can't close it!

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -99,11 +110,13 @@ class RefreshIngestionResourcesTask extends TimerTask {
             public void run() {
                 try {
                     refreshIngestionResources();
-                    timer.schedule(new RefreshIngestionResourcesTask(), defaultRefreshTime);
```
AsafMah, line 102: This changes the semantics, no?
What causes it to be scheduled more than once? (if there isn't an exception)?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -114,18 +127,22 @@ class RefreshIngestionAuthTokenTask extends TimerTask {
...
+                        refreshIngestionAuthTokenTask.cancel();
+                        refreshIngestionAuthTokenTask = newRefreshIngestionAuthTokenTask;
                     }
                 }
             }
         }
 
-        timer.schedule(new RefreshIngestionAuthTokenTask(), 0);
-        timer.schedule(new RefreshIngestionResourcesTask(), 0);
+        refreshIngestionAuthTokenTask = new RefreshIngestionAuthTokenTask();
```
ohadbitt, line 142: Really no need for that - task cant run after purge or cancel

What fixed it for you was just the closing of the client

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
ohadbitt: This is going to change soon 
also this is a client from the user - he should close it himself

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
yihezkel: You're right, thanks!

In this change, I wanted to reuse the task objects instead of recreating them every execution, particularly because I was concerned about a potential memory leak. But I see there isn't one, and GC will handle each of the tasks.

I reverted those changes.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
yihezkel: That's true for the `ManagedStreamingIngestClient` flow.
But this is also called in `com.microsoft.azure.kusto.ingest.QueuedIngestClientImpl#QueuedIngestClientImpl(com.microsoft.azure.kusto.data.auth.ConnectionStringBuilder, com.microsoft.azure.kusto.data.HttpClientProperties, boolean)`
specifically via this code:
`this(csb, properties == null ? null : HttpClientFactory.create(properties), autoCorrectEndpoint)`

So in this flow, the `CloseableHttpClient httpClient` is **never** created by the user.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
yihezkel: It shouldn't change the semantics, but after looking more deeply the existing code shouldn't cause the task to be scheduled more than once.
I do prefer the change I made, because the current approach creates a new task instance on each call (whether it succeeded or failed) instead of reusing. But GC precludes that from being a problem. So I backed this part of my change out.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -91,6 +94,11 @@ public void close() {
         timer = null;
```
ohadbitt, line 94: I see you reverted also the `timer = null` that on the previous commit was after closing - I actually saw it and thought this is good so we don't get null pointer above

PR #346, opened 2024-01-23
Fix CRP's servertimeout bugs in how it's set/get, and make consistent

yihezkel:
### Fixed
Fix CRP's `servertimeout` bugs in how it's set/get, and make consistent

We have 3 different ways to set the `servertimeout`:
1) Set the timeout in milliseconds, using helper method (`clientRequestProperties.setTimeoutInMilliSec(serverTimeoutOptionMillis)`)
2) Set the timeout in milliseconds, directly via OPTION_SERVER_TIMEOUT (`clientRequestProperties.setOption(ClientRequestProperties.OPTION_SERVER_TIMEOUT, serverTimeoutOptionMillis)`)
3) Set the timeout to a timespan, directly via OPTION_SERVER_TIMEOUT (`clientRequestProperties.setOption(ClientRequestProperties.OPTION_SERVER_TIMEOUT, serverTimeoutOptionTimespan)`)

There were inconsistencies between how these 3 worked.

And 3 different ways to get the `servertimeout`:
1) `clientRequestProperties.getTimeoutInMilliSec()`
2) `clientRequestProperties.getOption(OPTION_SERVER_TIMEOUT)`
3) `clientRequestProperties.getTimeoutAsString()` (which really should be `clientRequestProperties.getTimeoutAsTimespan()`)

There were inconsistencies between how these 3 worked.

The easiest way to understand this PR is to look at the requirements (mapping of inputs to expected outputs), defined in `ClientRequestPropertiesTest.provideInputsForConvertTimeoutBetweenMillisAndTimespan()`.

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
@@ -50,6 +50,7 @@ public class ClientRequestProperties implements Serializable, TraceableAttribute
     private static final String PARAMETERS_KEY = "Parameters";
     private final Map<String, Object> parameters;
     private final Map<String, Object> options;
+    static final long MIN_TIMEOUT_MS = TimeUnit.MINUTES.toMillis(1);
     static final long MAX_TIMEOUT_MS = TimeUnit.HOURS.toMillis(1);
     private String clientRequestId;
     private String application;
@@ -151,8 +152,15 @@ public class ClientRequestProperties implements Serializable, TraceableAttribute
         parameters.clear();
     }
 
+    /**
+     * Gets the amount of time a query may execute on the service before it times out. Value must be between 1 minute and 1 hour,
+     * and so if the value had been set below the minimum or above the maximum, the value returned will be adjusted accordingly.
+     */
     public Long getTimeoutInMilliSec() throws ParseException {
-        Object timeoutObj = getOption(OPTION_SERVER_TIMEOUT);
+        return getTimeoutInMilliSec(getOption(OPTION_SERVER_TIMEOUT));
+    }
+
+    private static Long getTimeoutInMilliSec(Object timeoutObj) throws ParseException {
         Long timeout = null;
         if (timeoutObj instanceof Long) {
             timeout = (Long) timeoutObj;
@@ -162,10 +170,10 @@ public class ClientRequestProperties implements Serializable, TraceableAttribute
             timeout = Long.valueOf((Integer) timeoutObj);
         }
 
-        return timeout;
+        return calculateTimeoutPerServiceLogic(timeout);
     }
 
-    private long parseTimeoutFromTimespanString(String str) throws ParseException {
+    private static long parseTimeoutFromTimespanString(String str) throws ParseException {
         Matcher matcher = KUSTO_TIMESPAN_REGEX.matcher(str);
         if (!matcher.matches()) {
             throw new ParseException(String.format("Failed to parse timeout string as a timespan. Value: '%s'", str));
@@ -174,7 +182,7 @@ public class ClientRequestProperties implements Serializable, TraceableAttribute
         if ("-".equals(matcher.group(1))) {
             throw new IllegalArgumentException(String.format("Negative timeouts are invalid. Value: '%s'", str));
         }
-        long millis = 0;
+
         String days = matcher.group(2);
         if (days != null && !days.equals("0") && !days.equals("00")) {
             return MAX_TIMEOUT_MS;
@@ -186,19 +194,29 @@ public class ClientRequestProperties implements Serializable, TraceableAttribute
                 timespanWithoutDays += matcher.group(i);
             }
         }
-        millis += TimeUnit.NANOSECONDS.toMillis(LocalTime.parse(timespanWithoutDays).toNanoOfDay());
-        return millis;
+
+        return TimeUnit.NANOSECONDS.toMillis(LocalTime.parse(timespanWithoutDays).toNanoOfDay());
     }
 
     /**
      * Sets the amount of time a query may execute on the service before it times out.
-     * @param timeoutInMs number of milliseconds before timeout. Value must be between 1 minute and 1 hour.
+     * @param timeoutInMs number of milliseconds before timeout.
+     *                    Value must be between 1 minute and 1 hour, and so value below the minimum or above the maximum will be adjusted accordingly.
      */
     public void setTimeoutInMilliSec(Long timeoutInMs) {
-        if (timeoutInMs < 60000 || timeoutInMs > MAX_TIMEOUT_MS) {
-            throw new IllegalArgumentException(String.format("Timeout must be between 60,000 and %,d milliseconds", MAX_TIMEOUT_MS));
+        options.put(OPTION_SERVER_TIMEOUT, calculateTimeoutPerServiceLogic(timeoutInMs));
+    }
+
+    private static Long calculateTimeoutPerServiceLogic(Long timeoutInMs) {
+        if (timeoutInMs != null) {
+            if (timeoutInMs < MIN_TIMEOUT_MS) {
+                return MIN_TIMEOUT_MS;
+            } else if (timeoutInMs > MAX_TIMEOUT_MS) {
+                return MAX_TIMEOUT_MS;
+            }
         }
-        options.put(OPTION_SERVER_TIMEOUT, timeoutInMs);
+
+        return timeoutInMs;
     }
 
     JsonNode toJson() {
@@ -301,17 +319,28 @@ public class ClientRequestProperties implements Serializable, TraceableAttribute
         return attributes;
     }
 
+    /**
+     * Gets the amount of time a query may execute on the service before it times out, formatted as a KQL timespan.
+     * @param timeoutObj amount of time before timeout, which may be a Long, String or Integer.
+     *                    Value must be between 1 minute and 1 hour, and so value below the minimum or above the maximum will be adjusted accordingly.
+     */
+    // TODO: Breaking change for next major release: static String getTimeoutAsTimespan(Object timeoutObj)
     String getTimeoutAsString(Object timeoutObj) {
-        String timeoutString = "";
-        if (timeoutObj instanceof Long) {
-            Duration duration = Duration.ofMillis((Long) timeoutObj);
-            timeoutString = Utils.formatDurationAsTimespan(duration);
-        } else if (timeoutObj instanceof String) {
-            timeoutString = (String) timeoutObj;
-        } else if (timeoutObj instanceof Integer) {
-            Duration duration = Duration.ofMillis((Integer) timeoutObj);
-            timeoutString = Utils.formatDurationAsTimespan(duration);
+        Long timeoutInMilliSec = getTimeoutInMilliSec(timeoutObj);
+
+        if (timeoutInMilliSec == null) {
+            return null;
         }
-        return timeoutString;
+
+        Duration duration = Duration.ofMillis(timeoutInMilliSec);
+        return Utils.formatDurationAsTimespan(duration);
+    }
+
+    /**
+     * Gets the amount of time a query may execute on the service before it times out, formatted as a KQL timespan.
+     * Value must be between 1 minute and 1 hour, and so if the value had been set below the minimum or above the maximum, the value returned will be adjusted accordingly.
+     */
+    String getTimeoutAsTimespan() {
+        return getTimeoutAsString(getOption(OPTION_SERVER_TIMEOUT));
     }
 }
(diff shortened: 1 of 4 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -151,8 +152,15 @@ public void clearParameters() {
...
+    /**
+     * Gets the amount of time a query may execute on the service before it times out. Value must be between 1 minute and 1 hour,
+     * and so if the value had been set below the minimum or above the maximum, the value returned will be adjusted accordingly.
+     */
     public Long getTimeoutInMilliSec() throws ParseException {
-        Object timeoutObj = getOption(OPTION_SERVER_TIMEOUT);
+        return getTimeoutInMilliSec(getOption(OPTION_SERVER_TIMEOUT));
+    }
+
+    private static Long getTimeoutInMilliSec(Object timeoutObj) throws ParseException {
```
yihezkel, line 163: Defining this method allows for DRY, reused by the existing `String getTimeoutAsString(Object timeoutObj)`

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -186,19 +194,29 @@ private long parseTimeoutFromTimespanString(String str) throws ParseException {
...
     /**
      * Sets the amount of time a query may execute on the service before it times out.
-     * @param timeoutInMs number of milliseconds before timeout. Value must be between 1 minute and 1 hour.
+     * @param timeoutInMs number of milliseconds before timeout.
+     *                    Value must be between 1 minute and 1 hour, and so value below the minimum or above the maximum will be adjusted accordingly.
      */
     public void setTimeoutInMilliSec(Long timeoutInMs) {
-        if (timeoutInMs < 60000 || timeoutInMs > MAX_TIMEOUT_MS) {
-            throw new IllegalArgumentException(String.format("Timeout must be between 60,000 and %,d milliseconds", MAX_TIMEOUT_MS));
+        options.put(OPTION_SERVER_TIMEOUT, calculateTimeoutPerServiceLogic(timeoutInMs));
```
yihezkel, line 207: This change was the impetus for this PR, as @yogilad said it's better to not break users' existing usages.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -301,17 +319,28 @@ public Map<String, String> getTracingAttributes() {
         return attributes;
     }
 
+    /**
+     * Gets the amount of time a query may execute on the service before it times out, formatted as a KQL timespan.
+     * @param timeoutObj amount of time before timeout, which may be a Long, String or Integer.
+     *                    Value must be between 1 minute and 1 hour, and so value below the minimum or above the maximum will be adjusted accordingly.
+     */
+    // TODO: Breaking change for next major release: static String getTimeoutAsTimespan(Object timeoutObj)
```
yihezkel, line 327: Note that this method returns a `timespan`, not just any `String`.

Also, similar existing methods already use this kind of naming.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -301,17 +319,28 @@ public Map<String, String> getTracingAttributes() {
...
-        String timeoutString = "";
-        if (timeoutObj instanceof Long) {
-            Duration duration = Duration.ofMillis((Long) timeoutObj);
-            timeoutString = Utils.formatDurationAsTimespan(duration);
-        } else if (timeoutObj instanceof String) {
-            timeoutString = (String) timeoutObj;
-        } else if (timeoutObj instanceof Integer) {
-            Duration duration = Duration.ofMillis((Integer) timeoutObj);
-            timeoutString = Utils.formatDurationAsTimespan(duration);
+        Long timeoutInMilliSec = getTimeoutInMilliSec(timeoutObj);
```
yihezkel, line 329: The primary difference in behavior here is when a `timespan` is provided.

Previously, it simply returned back whatever had been provided as input.
Now, it first calls `parseTimeoutFromTimespanString(String inputAsTimespan)` to ensure it's a valid `timespan`, and then applies `calculateTimeoutPerServiceLogic(Long inputInMillis)` for consistent adjustment of its value.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -301,17 +319,28 @@ public Map<String, String> getTracingAttributes() {
...
+
+        Duration duration = Duration.ofMillis(timeoutInMilliSec);
+        return Utils.formatDurationAsTimespan(duration);
+    }
+
+    /**
+     * Gets the amount of time a query may execute on the service before it times out, formatted as a KQL timespan.
+     * Value must be between 1 minute and 1 hour, and so if the value had been set below the minimum or above the maximum, the value returned will be adjusted accordingly.
+     */
+    String getTimeoutAsTimespan() {
```
yihezkel, line 343: It makes more sense to have this method than the existing `String getTimeoutAsString(Object timeoutObj)`.

Also, it's consistent with the existing `public Long getTimeoutInMilliSec()`.

I kept the access modifier consistent with the existing method. I suggest changing this to `public` when we make the breaking change I suggest in the previous method, since the different namings (`getTimeoutAsString` vs `getTimeoutAsTimespan`) may be confusing. Or I can do that now.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -67,21 +67,27 @@ public static String getPackageVersion() {
...
     public static String formatDurationAsTimespan(Duration duration) {
-        long seconds = duration.getSeconds();
+        long durationInSeconds = duration.getSeconds();
         int nanos = duration.getNano();
-        long hours = TimeUnit.SECONDS.toHours(seconds) % TimeUnit.DAYS.toHours(1);
-        long minutes = TimeUnit.SECONDS.toMinutes(seconds) % TimeUnit.MINUTES.toSeconds(1);
-        long secs = seconds % TimeUnit.MINUTES.toSeconds(1);
-        long days = TimeUnit.SECONDS.toDays(seconds);
-        String positive = String.format(
-                "%02d.%02d:%02d:%02d.%.3s",
```
yihezkel, line 77: My understanding is that the `timespan` standard does not include the days if there are none, nor the subseconds if there are none.

data/src/test/java/com/microsoft/azure/kusto/data/ClientRequestPropertiesTest.java
```diff
@@ -11,51 +11,88 @@
...
-    @DisplayName("test set/get timeout")
-    void timeoutSetGet() {
-        ClientRequestProperties props = new ClientRequestProperties();
-        Long expected = TimeUnit.MINUTES.toMillis(10);
-
-        // before setting value should be null
-        Assertions.assertNull(props.getTimeoutInMilliSec());
+    // Used by the 3 parameterized tests below, in the format: {inputTimespan, inputMillis, expectedTimespan, expectedMillis}. Each method uses only one of the
+    // 2 input parameters.
+    private static Stream<Arguments> provideInputsForConvertTimeoutBetweenMillisAndTimespan() {
```
yihezkel, line 31: I'm a fan of UTs being the single source of truth reflecting requirements.

data/src/test/java/com/microsoft/azure/kusto/data/ClientRequestPropertiesTest.java
```diff
@@ -72,21 +109,21 @@ void propertiesToString() throws JsonProcessingException {
     @Test
     @DisplayName("test ClientRequestProperties fromString")
     void stringToProperties() throws JsonProcessingException {
-        String properties = "{\"Options\":{\"servertimeout\":\"01:25:11.111\", \"Content-Encoding\":\"gzip\"},\"Parameters\":{\"birthday\":\"datetime(1970-05-11)\",\"courses\":\"dynamic(['Java', 'C++'])\"}}";
+        String properties = "{\"Options\":{\"servertimeout\":\"00:25:11.111\", \"Content-Encoding\":\"gzip\"},\"Parameters\":{\"birthday\":\"datetime(1970-05-11)\",\"courses\":\"dynamic(['Java', 'C++'])\"}}";
```
yihezkel, line 112: This is a great example of the previous inconsistencies:
Setting a timeout of >1 day should result in an actual timeout of 1 hour - which it sometimes did, but this is one of the cases when it didn't.

data/src/test/java/com/microsoft/azure/kusto/data/UtilitiesTest.java
```diff
@@ -18,34 +17,8 @@
 import java.net.UnknownHostException;
 import java.nio.file.Paths;
 import java.util.Objects;
-import java.util.concurrent.TimeUnit;
 
 class UtilitiesTest {
-    @Test
-    @DisplayName("Convert millis to .Net timespan")
-    void convertMillisToTimespan() {
```
yihezkel, line 26: This is effectively a CRP UT

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: I'll add the replacement, too: `String getTimeoutAsTimespan(Object timeoutObj)`

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -186,19 +194,29 @@ private long parseTimeoutFromTimespanString(String str) throws ParseException {
...
+     * @param timeoutInMs number of milliseconds before timeout.
+     *                    Value must be between 1 minute and 1 hour, and so value below the minimum or above the maximum will be adjusted accordingly.
      */
     public void setTimeoutInMilliSec(Long timeoutInMs) {
-        if (timeoutInMs < 60000 || timeoutInMs > MAX_TIMEOUT_MS) {
-            throw new IllegalArgumentException(String.format("Timeout must be between 60,000 and %,d milliseconds", MAX_TIMEOUT_MS));
+        options.put(OPTION_SERVER_TIMEOUT, calculateTimeoutPerServiceLogic(timeoutInMs));
+    }
+
+    private static Long calculateTimeoutPerServiceLogic(Long timeoutInMs) {
```
AsafMah, line 210: Not sure I like the name, maybe something like - adjustTimeoutToServiceLimits ?

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
AsafMah: I feel like with this name it's expected to return a Timespan object of some sorts. Maybe a name like getTimeoutFormatted would show the intention better? can you give other names that are consistent with that?

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -151,8 +152,15 @@ public void clearParameters() {
...
+     * Gets the amount of time a query may execute on the service before it times out. Value must be between 1 minute and 1 hour,
+     * and so if the value had been set below the minimum or above the maximum, the value returned will be adjusted accordingly.
+     */
     public Long getTimeoutInMilliSec() throws ParseException {
-        Object timeoutObj = getOption(OPTION_SERVER_TIMEOUT);
+        return getTimeoutInMilliSec(getOption(OPTION_SERVER_TIMEOUT));
+    }
+
+    private static Long getTimeoutInMilliSec(Object timeoutObj) throws ParseException {
         Long timeout = null;
```
AsafMah, line 164: early return of "if timeoutObj == null"? to make it clear

AsafMah (commented):
Seems nice, some comments about the naming mostly

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: I see the naming concern here. OTOH, I think it can also intuitively makes sense that an "As/In" doesn't necessarily mean the object's literal `class`, but instead the business description.

For example, `getTimeoutInMilliSec()` returns a Long, which I think is also intuitive.

AsafMah: I don't think we use the Csl terminology much.

how about formatTimeoutAsTimespan ?

AsafMah (approved):
Approved but see my latest reply

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yihezkel: That's what we call it, including in the Java SDK:
`public class CslTimespanFormat extends CslFormat`

yihezkel: I guess `getTimeoutFormattedAsTimespan()` could work, but that sounds overly verbose to me.

PR #348, opened 2024-01-30
Fix integration tests that should be unit tests

yihezkel:
### Changed
Fix integration tests that should be unit tests


ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientBaseTest.java
```diff
@@ -0,0 +1,32 @@
...
+                Arguments.of("https://376.568.1564.1564", "https://ingest-376.568.1564.1564"),
+                Arguments.of("https://192.168.1.1", "https://192.168.1.1"),
+                Arguments.of("https://2345:0425:2CA1:0000:0000:0567:5673:23b5", "https://2345:0425:2CA1:0000:0000:0567:5673:23b5"),
+                Arguments.of("https://127.0.0.1", "https://127.0.0.1"),
+                Arguments.of("https://localhost", "https://localhost"),
+                Arguments.of("https://onebox.dev.kusto.windows.net", "https://onebox.dev.kusto.windows.net"));
+    }
+
+    @ParameterizedTest
+    @MethodSource("provideStringsForAutoCorrectEndpointTruePass")
```
AsafMah, line 27: Where did the tests where the autocorrect is false gone?

AsafMah (changes requested):

ingest/src/test/java/com/microsoft/azure/kusto/ingest/IngestClientBaseTest.java
yihezkel: No longer relevant, since the previous tests created a new client, which branched on whether to call the test being unit tested based on that `boolean`.

Now, since we're only testing the method itself, the autocorrect `boolean` is effectively always true.

I suppose it would be good to also unit test creating a client (would require a bit of mocking), once with the `boolean` set to `true` and once with it set to `false`. But we don't need to test each URL like how it did before, nor would we want the test to run without mocking like it did before.

AsafMah (approved):

PR #353, opened 2024-02-18
Fix ConcurrentModificationException

yihezkel:
### Fixed
Fix ConcurrentModificationException

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
@@ -3,7 +3,8 @@ package com.microsoft.azure.kusto.ingest.resources;
 import com.microsoft.azure.kusto.ingest.utils.TimeProvider;
 
 import java.util.ArrayDeque;
-import java.util.Iterator;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 public class RankedStorageAccount {
     static class Bucket {
@@ -17,6 +18,7 @@ public class RankedStorageAccount {
     }
 
     private final ArrayDeque<Bucket> buckets = new ArrayDeque<>();
+    private final ReadWriteLock bucketsLock = new ReentrantReadWriteLock();
     private final String accountName;
     private final int maxNumberOfBuckets;
     private final int bucketDurationMillis;
@@ -43,7 +45,12 @@ public class RankedStorageAccount {
     private Bucket adjustForTimePassed() {
         if (buckets.isEmpty()) {
             Bucket b = new Bucket();
-            buckets.push(b);
+            try {
+                bucketsLock.writeLock().lock();
+                buckets.push(b);
+            } finally {
+                bucketsLock.writeLock().unlock();
+            }
             lastActionTimestamp = timeProvider.currentTimeMillis();
             return b;
         }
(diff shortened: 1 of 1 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
```diff
@@ -82,8 +83,9 @@ public double getRank() {
         // For each bucket, calculate the success rate ( success / total ) and multiply it by the bucket weight.
         // The older the bucket, the less weight it has. For example, if there are 3 buckets, the oldest bucket will have
         // a weight of 1, the middle bucket will have a weight of 2 and the newest bucket will have a weight of 3.
-
-        for (Bucket bucket : buckets) {
+        Iterator<Bucket> iterator = buckets.iterator();
```
yihezkel, line 86: We could alternatively lock this section from read access while `buckets` is being modified, but I don't think that's necessary, and therefore not worth the perf penalty.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
```diff
@@ -82,22 +95,25 @@ public double getRank() {
         // For each bucket, calculate the success rate ( success / total ) and multiply it by the bucket weight.
         // The older the bucket, the less weight it has. For example, if there are 3 buckets, the oldest bucket will have
         // a weight of 1, the middle bucket will have a weight of 2 and the newest bucket will have a weight of 3.
-
-        for (Bucket bucket : buckets) {
-            if (bucket.totalCount == 0) {
+        try {
+            bucketsLock.readLock().lock();
+            for (Bucket bucket : buckets) {
```
yihezkel, line 100: @AsafMah was right :-)
So since the modification happens in other code, we have to use `locks`.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
```diff
@@ -42,7 +45,12 @@ public void addResult(boolean success) {
     private Bucket adjustForTimePassed() {
         if (buckets.isEmpty()) {
```
AsafMah, line 46: ou don't need a lock here?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
```diff
@@ -53,16 +61,21 @@ private Bucket adjustForTimePassed() {
             return buckets.peek();
         }
 
-        if (bucketsToCreate >= maxNumberOfBuckets) {
-            buckets.clear();
-            buckets.push(new Bucket());
-        } else {
-            for (int i = 0; i < bucketsToCreate; i++) {
+        try {
+            bucketsLock.writeLock().lock();
```
AsafMah, line 65: Isn't there a better construct for locking/unlocking (try-with-resources?) or does the lock not support it

ohadbitt:
Can you please paste the error you got in the PR description ?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
ohadbitt: If you want we can create a closeable lock ourselves

ohadbitt (approved):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
The-Funk: Would an AtomicReference work instead? Thread-safe but no lock needed.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
yihezkel: We should if we're concerned about race conditions, which I don't think we are.

In terms of preventing `ConcurrentModificationException`, I don't think locking here would help, because `ArrayDeque.isEmpty()`'s implementation is simply `return head == tail;`.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/resources/RankedStorageAccount.java
yihezkel: We could use `synchronized` blocks which look a bit simpler, but the benefit of this is that read locks don't prevent other threads from reading.

Regarding `AtomicReference`, thanks for your suggestion, @The-Funk! But I'm not sure the code would be simpler in this scenario because we're adding thread safety to an `ArrayDeque`, which is a collection that isn't itself thread-safe (which we might be able to change, but I don't know if the code would be simpler).

AsafMah (approved):

PR #357, opened 2024-03-25
bumps dependencies, removes explicit reactor and msal4j dependency

The-Funk:
### Changed
- bumps jackson dependency for alignment with latest GA databricks runtime
- removes explicit declaration of reactor-core dependency (allow version to come from azure-core)
- removes explicit declaration of msal4j dependency (allow version to come from azure-identity)
- bumps various other dependencies for security and latest features

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <java.version>1.8</java.version>
         <!-- Use 1.0.6 for fasterxml v 2.12.x (good for databricks runtime 2.10.x) -->
-        <azure-bom-version>1.2.18</azure-bom-version>
+        <azure-bom-version>1.2.21</azure-bom-version>
 
         <!-- Versions below are for several dependencies we're using in the data & ingest modules -->
         <!-- Ideally, versions below should align with latest databricks runtime dependency versions -->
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -35,40 +35,38 @@
         <revision>6.0.0</revision> <!-- CHANGE THIS to adjust project version-->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <java.version>1.8</java.version>
-        <!-- Use 1.0.6 for fasterxml v 2.12.x (good for databricks runtime 2.10.x) -->
-        <azure-bom-version>1.2.18</azure-bom-version>
+        <azure-bom-version>1.2.22</azure-bom-version>
 
         <!-- Versions below are for several dependencies we're using in the data & ingest modules -->
         <!-- Ideally, versions below should align with latest databricks runtime dependency versions -->
         <!-- Compile dependencies -->
```
ag-ramachandran, line 42: One of the things we have to fix is the dependencies for SL4J, Reactor (removed in this version), Jackson (discussed below) and Netty (not explcitly added should be okay).

Adding custom versions causes issues downstream for apps using SDK to build on. 

Ref : https://learn.microsoft.com/en-us/azure/developer/java/sdk/troubleshooting-dependency-version-conflict

pom.xml
```diff
@@ -35,40 +35,38 @@
...
-        <junit.version>5.8.2</junit.version>
-        <mockito.version>5.7.0</mockito.version>
+        <maven-assembly-plugin.version>3.7.1</maven-assembly-plugin.version>
+        <maven-surefire-plugin.version>3.2.5</maven-surefire-plugin.version>
+        <maven-javadoc-plugin.version>3.6.3</maven-javadoc-plugin.version>
+        <maven-source-plugin.version>3.3.0</maven-source-plugin.version>
+        <maven-dependency-plugin.version>3.6.1</maven-dependency-plugin.version>
+        <flatten-maven-plugin.version>1.2.7</flatten-maven-plugin.version>
+        <junit.version>5.10.2</junit.version>
+        <mockito.version>5.11.0</mockito.version>
```
ag-ramachandran, line 69: Please correct me here :  did mockito not move out of JDK 8 to JDK 11. There is some diff in inline and normal mocks is it perhaps that ? (or) do we even use Mockito if it does not give us that error of 
"This library is compiled with Version 55 and will not work with 52"

Ref : https://github.com/mockito/mockito/releases/tag/v5.0.0 

_What should I do if I still run JDK 8?
For JDK 8 and below, you can keep on using Mockito 4._

ag-ramachandran (changes requested):

pom.xml
```diff
@@ -35,40 +35,38 @@
...
         <!-- Versions below are for several dependencies we're using in the data & ingest modules -->
         <!-- Ideally, versions below should align with latest databricks runtime dependency versions -->
         <!-- Compile dependencies -->
         <slf4j.version>1.7.36</slf4j.version>
         <commons-lang3.version>3.14.0</commons-lang3.version>
         <commons-text.version>1.11.0</commons-text.version>
-        <commons-codec.version>1.16.0</commons-codec.version>
-        <msal4j.version>1.13.10</msal4j.version>
+        <commons-codec.version>1.16.1</commons-codec.version>
+        <msal4j.version>1.14.3</msal4j.version>
```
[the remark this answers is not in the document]
The-Funk: There was no reason I could think of and I agree with your take that we should have less explicit dependencies. Removed the explicit version in [this commit](https://github.com/Azure/azure-kusto-java/pull/357/commits/4fce50241edec5e53d8e989b8f8dd5dc2be256f2).

pom.xml
```diff
@@ -35,40 +35,38 @@
...
         <apache.httpclient.version>4.5.14</apache.httpclient.version>
         <httpcore.version>4.4.16</httpcore.version>
-        <fasterxml.jackson.core.version>2.15.2</fasterxml.jackson.core.version>
-        <reactor-core.version>3.4.34</reactor-core.version>
+        <fasterxml.jackson.core.version>2.16.1</fasterxml.jackson.core.version>
```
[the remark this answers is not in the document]
The-Funk: @ohadbitt sorry about that. Went one version too far. 2.16.0 should align with latest GA ADB runtime according to the link you sent. Fixed in [this commit](https://github.com/Azure/azure-kusto-java/pull/357/commits/c13aba6a6ccb89494df35ded6adb018b71bcd32b).

pom.xml
The-Funk: @ag-ramachandran I believe you are correct. Directly beneath this line in the POM file (lines 72 to 81) it looks like someone set JDK 8 and 11 builds to utilize Mockito 4.5.1 whereas I suppose the other builds (17 and 21) use this dependency.

ag-ramachandran (approved):

AsafMah:
@The-Funk  @ag-ramachandran seems the build errors:
Error:  Used undeclared dependencies found:
Error:     com.microsoft.azure:msal4j:jar:1.14.3:compile
Error:     io.projectreactor:reactor-core:jar:3.4.34:compile

The-Funk:
> @The-Funk @ag-ramachandran seems the build errors: Error: Used undeclared dependencies found: Error: com.microsoft.azure:msal4j:jar:1.14.3:compile Error: io.projectreactor:reactor-core:jar:3.4.34:compile

This is due to the maven-dependency plugin used in this project. I'm not sure how to configure that plugin to tell it that we get these dependencies elsewhere and they don't need explicitly declared.

The-Funk:
@AsafMah I believe the commit above should resolve the build issue. I updated the dependency plugin to ignore those two used but undeclared dependencies

Issue #249, opened 2022-06-08
Asynchronous HTTP Client


The-Funk:
**Describe the solution you'd like**
Utilize an asynchronous HTTP client instead of Apache's CloseableHttpClient in order to provide a nonblocking transport of data from the ADX database to my Java application.

**Describe alternatives you've considered**
NA

**Additional context**
The current implementation of the ADX library uses the Apache HTTP Client. This client hasn't received an update in 2 years. A good asynchronous alternative might be to use the vert.x web client. https://vertx.io/docs/vertx-web-client/java/ 
https://mvnrepository.com/artifact/io.vertx/vertx-web-client

The-Funk:
Side note, you might want to offer multiple "flavors" for http client since several asynchronous clients exist that are rather opinionated and provide their own "future" or async types. vert,x is one of these.

The-Funk:
I submitted a PR to bump the Apache HTTP Client to version 5 which has async support. For the most part the code from version 4 could be reused. Very few breaking changes. Might be worth looking into.

The-Funk:
Just checking in again to see if anything is being done on this. CompletableFuture support from Apache HTTPClient 5 would be enough to resolve most of my issues I think.

AsafMah:
Sadly this feature will be postponed for now.
When we will add it, we will probably move to new standard http client in java 11.

The-Funk:
I noticed the Java 11 implementation has gone through, would you like help refactoring to use the Java 11 client?

The-Funk:
@AsafMah do you know if this is still in the pipeline along with the native Java client?

I'm also noticing that this library using msal4j, are there plans to switch to azure-identity at some point?

yogilad:
Regarding use of MSAL, we do not plan on rewriting all the auth code using Azure.Identity, however we should provide a WithAzureTokenCredential() API for customers to work with.

@AsafMah , consider adding this to Ron's task list so this gets done sooner then later.

yogilad:
This work is planned anytime soon.
The main reason is, it requires rearchitecting the SDK to avoid code duplications which comes at a high cost.

The-Funk:
@yogilad @AsafMah @ohadbitt I have extra availability in the coming weeks, would you like me to take a shot at upgrading the project to use httpclient5 again, or potentially switching to the Java 11 built-in HTTP Client?

Impact of moving to the Java 11 client:
  - potential for major rework of specific classes/methods
[...]

PR #358, opened 2024-03-26
Introduces KustoQuery wrapper object

The-Funk:
### Added
- KustoQuery object

**Why:**
Adding this object allows for extensibility without changing method signatures or chaining methods until you reach an implementation, which should reduce the surface of the library and make the code easier to read. Currently to execute a command you must pass each property needed individually. Sometimes this results in a lot of method chaining, which can get pretty verbose and hard to track. Additionally, when new options are made available to the user, more chains are needed to account for this, which means a lot more verbosity.

**Request:**
Deprecate the old methods that take each parameter individually, including the command and management methods, in favor of a single execute method that takes a KustoQuery object. This is a big change for users, however it should be a relatively simple one, and if the methods are annotated as being deprecated for a time prior to removal, this will give users time to migrate to using the KustoQuery object.

New usage:

```java
   var client = ClientFactory.createClient(engine_connection_string, httpClientProperties);

   // Example 1:
   client.execute(new KustoQuery("my query"));
   
   // Example 2
   var myProperties = new ClientRequestProperties();
   var query = new KustoQuery("my query", myProperties);
   client.executeToJsonResult(query);
```

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/Client.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/Client.java
@@ -9,6 +9,9 @@ import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
 import java.io.Closeable;
 
 public interface Client extends Closeable {
+
+    KustoOperationResult execute(KustoQuery kq) throws DataServiceException, DataClientException;
+
     KustoOperationResult execute(String command) throws DataServiceException, DataClientException;
 
     KustoOperationResult execute(String database, String command) throws DataServiceException, DataClientException;
@@ -26,6 +29,8 @@ public interface Client extends Closeable {
     KustoOperationResult executeMgmt(String database, String command) throws DataServiceException, DataClientException;
 
     KustoOperationResult executeMgmt(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException;
+
+    String executeToJsonResult(KustoQuery kq) throws DataServiceException, DataClientException;
 
     String executeToJsonResult(String database) throws DataServiceException, DataClientException;
 
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -84,6 +84,17 @@ class ClientImpl extends BaseClient {
     }
 
     @Override
+    public KustoOperationResult execute(KustoQuery kq) throws DataServiceException, DataClientException {
+        if (kq == null) {
+            throw new IllegalArgumentException("KustoQuery object cannot be null in order to be executed.");
+        }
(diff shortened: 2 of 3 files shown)


The-Funk:
I meant to open this in draft. If there is interest in the request section, I can annotate the methods in question as deprecated.

ohadbitt:
I dont think we want this to be on our public API , we try to maintain a fairly similar API in all our sdks

The-Funk:
> I dont think we want this to be on our public API , we try to maintain a fairly similar API in all our sdks

@ohadbitt the reason I really liked this is just that a wrapper object like this really shortens your method signatures and reduces the number of methods you need to maintain. The object acts as a simple container that can hold whatever needs to be sent between methods, which in the long run saves you from needing to add more methods with different signatures when that context undergoes changes. For instance, when you need to add an enhancement you can just add the additional context to the wrapper, add the logic to the method implementation that consumes the wrapper, and then add your documentation. If you package JavaDocs the documentation is just some comments over some getters/setters.

If you have a chance, can you check out my comments over on issue #249 ? I was hoping to preface the async implementation with this, but I don't mind following one of the other potential paths if you have a preferred implementation instead. I get wanting to keep the SDKs aligned for sure.

AsafMah:
We actually want to remove some of the overload in the next version.
This week Me and Ohad will brainstorm on the async design, and we'll see how the API will change.
Maybe this class should be used internally?

The-Funk:
> We actually want to remove some of the overload in the next version. This week Me and Ohad will brainstorm on the async design, and we'll see how the API will change. Maybe this class should be used internally?

I think unless you exposed the container object like this one on the API somewhere, you would have to chain/overload methods. That said, if it's important to chain methods to keep the APIs aligned, I don't mind writing some extra boilerplate. Let me know what y'all come up with on the async implementation. I've been messing around with it. There's a lot of work no matter which way we go.

The-Funk:
@ohadbitt @AsafMah  Can either of you trigger CI tests to run on this for me? I am confident it passes.

AsafMah:
Don't know since when it's a thing, but github won't run the trigger on PRs with merge conflicts - https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

So can you fix those first?

The-Funk:
> Don't know since when it's a thing, but github won't run the trigger on PRs with merge conflicts - https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request
> 
> So can you fix those first?

Since the two branches are so different and the diffs are really hard to read , I'm probably going to create a new branch and manually verify each file individually. I don't want to throw out anyone's commits but there's major changes between this branch and main, so it probably makes sense to start fresh and check each file by hand. Otherwise I'd have to cherry pick commits, and that doesn't sound fun.

PR #359, opened 2024-03-26
App and user for tracing

t-ronmoneta:
### Added
app and user for tracing fields in ingestion properties
Took those fields from KCSB and added them to ingestion Propeties in queued ingest client

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionBlobInfo.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionBlobInfo.java
@@ -30,8 +30,10 @@ public final class IngestionBlobInfo {
     private Boolean flushImmediately;
     private IngestionStatusInTableDescription ingestionStatusInTable;
     private Map<String, String> additionalProperties;
+    private String applicationForTracing;
+    private String clientVersionForTracing;
 
-    public IngestionBlobInfo(String blobPath, String databaseName, String tableName) {
+    public IngestionBlobInfo(String blobPath, String databaseName, String tableName, String applicationForTracing, String clientVersionForTracing) {
         this.blobPath = blobPath;
         this.databaseName = databaseName;
         this.tableName = tableName;
@@ -41,6 +43,8 @@ public final class IngestionBlobInfo {
         reportMethod = IngestionProperties.IngestionReportMethod.QUEUE.getKustoValue();
         flushImmediately = false;
         sourceMessageCreationTime = LocalDateTime.ofInstant(Instant.now(), ZoneId.of("UTC")).toString();
+        this.applicationForTracing = applicationForTracing;
+        this.clientVersionForTracing = clientVersionForTracing;
     }
 
     public String getBlobPath() {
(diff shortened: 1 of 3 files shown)


ohadbitt (approved):
Just make sure it works :)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
```diff
@@ -57,6 +59,8 @@ public class QueuedIngestClientImpl extends IngestClientBase implements QueuedIn
         this.resourceManager = new ResourceManager(client, httpClient);
         this.azureStorageClient = new AzureStorageClient();
         this.connectionDataSource = csbWithEndpoint.getClusterUrl();
+        this.applicationForTracing = csb.getApplicationNameForTracing();
```
AsafMah, line 62: Create a client details object and use that - these properties are not formatted/have the wrong defaults

AsafMah (changes requested):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestionBlobInfo.java
```diff
@@ -130,4 +134,20 @@ public ValidationPolicy getValidationPolicy() {
     public void setValidationPolicy(ValidationPolicy validationPolicy) {
         this.validationPolicy = validationPolicy;
     }
+
+    public String getApplicationForTracing() {
```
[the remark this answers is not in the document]
t-ronmoneta: Common practice, can (and will) be removed

ingest/src/main/java/com/microsoft/azure/kusto/ingest/QueuedIngestClientImpl.java
t-ronmoneta: When looking at the client details object current usage - the appendedClientVersionForTracing is either null, or csb.getApplicationNameForTracing() as done here. so not sure why this is wrong

AsafMah:
```
 constructor(public applicationNameForTracing: string | null, public userNameForTracing: string | null) {
        if (this.applicationNameForTracing === null) {
            this.applicationNameForTracing = ClientDetails.defaultApplication();
        }
        if (this.userNameForTracing === null) {
            this.userNameForTracing = ClientDetails.defaultUser();
        }
        this.versionForTracing = ClientDetails.defaultVersion();
    }
````

t-ronmoneta: image
this is the usage so far as well in the client details too though - so using clientdetails would give you the exact same thing

AsafMah: The getter in ClientDetails returns the default if it's null - that's the piece of code I showed you.
I'm not sure what do you mean

t-ronmoneta: ohhh in applilcation tracing, all the time i looked at the second one, which is indded the same. ok, changing

AsafMah (approved):

PR #360, opened 2024-04-11
secretless

AsafMah:
### Added
- Azure CLI Token credential (didn't have it before)
### Changed
- Secretless build, using federated service principal.


.github/workflows/build.yml
```diff
@@ -4,30 +4,29 @@ name: Build Java
 
 on:
   push:
-    branches: [ '**' ]
+    branches: [ 'master' ]
```
ohadbitt, line 7: nice catch
who's paying on builds btw?

ohadbitt (approved):

.github/workflows/build.yml
AsafMah: Free for public builds, but they will limit you if you do too much. (I think)

PR #368, opened 2024-06-30
Managed streaming - better estimation for decision for streaming vs queuing 

ohadbitt:
### Added
- A new heuristic for choosing between queuing and streaming in Managed streaming client, the default behavior is using the 
an estimation against the 4mb limit after dividing or multiplying by some factor described by the consts:

 // Regardless of the format, we don't want to stream more than 10mb
int MAX_STREAMING_STREAM_SIZE_BYTES = 10 * 1024 * 1024;
    // Used against the users input of raw data size
int MAX_STREAMING_RAW_SIZE_BYTES = 6 * 1024 * 1024;
double JSON_UNCOMPRESSED_FACTOR = 1.5d;
int NON_BINARY_FACTOR = 2;
double BINARY_COMPRESSED_FACTOR = 2d;
double BINARY_UNCOMPRESSED_FACTOR = 1.5d;

This will also allow users to stream bigger than 4mb non-compressed data
- disableRetries option to client options - default is true only for streaming clients as if stream is not repeatable it will simply fails - better to let the user deal with it (alternative considered was to always create repeatable streams - like in managed streaming)
- better error messages


AsafMah:
This is a very big PR, can you summarize the changes in more detail?

Also, for disableRetries, maybe we want to fold this in to a more robust Policy object like @yogilad suggested and worked on.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -130,14 +130,20 @@ public IngestionResult ingestFromResultSet(ResultSetSourceInfo resultSetSourceIn
      * @see IngestionProperties
      */
     protected abstract IngestionResult ingestFromStreamImpl(StreamSourceInfo streamSourceInfo, IngestionProperties ingestionProperties)
-            throws IngestionClientException, IngestionServiceException;
+            throws IngestionClientException, IngestionServiceException, IOException;
```
AsafMah, line 133: Breaking change

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -130,14 +130,20 @@ public IngestionResult ingestFromResultSet(ResultSetSourceInfo resultSetSourceIn
...
             throws IngestionClientException, IngestionServiceException {
         // trace ingestFromStream
         return MonitoredActivity.invoke(
-                (SupplierTwoExceptions<IngestionResult, IngestionClientException, IngestionServiceException>) () -> ingestFromStreamImpl(streamSourceInfo,
-                        ingestionProperties),
+                (SupplierTwoExceptions<IngestionResult, IngestionClientException, IngestionServiceException>) () -> {
+                    try {
+                        return ingestFromStreamImpl(streamSourceInfo,
+                                ingestionProperties);
+                    } catch (IOException e) {
```
AsafMah, line 143: It seems that you don't even throw IOException, in the end

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -24,10 +25,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.ByteArrayInputStream;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.SequenceInputStream;
+import java.io.*;
```
AsafMah, line 28: I think * imports are considered bad? But I guess we use them elsewhere, we don't have a strong enough code guide

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/AbstractSourceInfo.java
```diff
@@ -22,9 +20,20 @@ public void setSourceId(UUID sourceId) {
...
 
+    // An estimation of the raw (uncompressed, un-indexed) size of the data, for binary formatted files - use only if known
+    private long rawSizeInBytes;
+
+    public long getRawSizeInBytes() {
+        return rawSizeInBytes;
+    }
+
+    // An estimation of the raw (uncompressed, un-indexed) size of the data, for binary formatted files - use only if known
+    public void setRawSizeInBytes(long rawSizeInBytes) {
```
AsafMah, line 31: The fact that there is no difference between "raw size is 0" and "raw size is unknown" makes me a bit worried

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -122,9 +122,9 @@ public static void setUp() {
     @AfterAll
     public static void tearDown() {
         try {
-            queryClient.executeToJsonResult(databaseName, String.format(".drop table %s ifexists", tableName));
             ingestClient.close();
```
AsafMah, line 125: Why the order change?
If the close() fails for one the above, it would leave the table hanging

ingest/src/test/java/com/microsoft/azure/kusto/ingest/tests.java
```diff
@@ -0,0 +1,28 @@
+package com.microsoft.azure.kusto.ingest;
+
+import com.microsoft.azure.kusto.ingest.IngestionProperties;
+import com.microsoft.azure.kusto.ingest.ManagedStreamingQueuingPolicy;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.*;
+
+class ManagedStreamingQueuingPolicyTest {
```
AsafMah, line 8: File name

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -39,19 +37,21 @@
...
 
     private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
     public static final int ATTEMPT_COUNT = 3;
-    public static final int MAX_STREAMING_SIZE_BYTES = 4 * 1024 * 1024;
     public static final String CLASS_NAME = ManagedStreamingIngestClient.class.getSimpleName();
     final QueuedIngestClientImpl queuedIngestClient;
     final StreamingIngestClient streamingIngestClient;
     private final ExponentialRetry exponentialRetryTemplate;
     private CloseableHttpClient httpClient = null;
+    private ManagedStreamingQueuingPolicy queuingPolicy = ManagedStreamingQueuingPolicy.Default;
```
AsafMah, line 54: Do we want to this here, like this, when v2 is in the horizon? I think we should talk to Yochai.

I don't mean the check itself, but making it customizable for the user.

We are commiting to this interface, which I'm not a big fan of, and I don't know if we have need for this to be supported.

In general, I have two concerns:

1. Why do we have the Policy class if it only has one member, and not just put the member directly
2. Why is the member a functional thing and not an interface?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -39,19 +37,21 @@
  * Since the streaming client communicates directly with the engine, it's more prone to failure, so this class
  * holds both a streaming client and a queued client.
  * It tries {@value ATTEMPT_COUNT} times using the streaming client, after which it falls back to the queued streaming client in case of failure.
- * If the size of the stream is bigger than {@value MAX_STREAMING_SIZE_BYTES}, it will fall back to the queued streaming client.
+ * By default the policy for choosing a queued ingestion on the first try is the checking of weather the size of the estimated
```
AsafMah, line 40: Please update CHANGELOG.md for every commit

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -39,19 +37,21 @@
...
+ * raw stream size (a conversion to compressed CSV) is bigger than 4MB, it will fall back to the queued streaming client.
+ * Use SourceInfo.size to override size estimations, alternatively - use setQueuingPolicy to override the predicate heuristics.
```
AsafMah, line 42: SourceInfo.size doesn't exist (as far as I can tell), we have rawdatasize, but I think this passage implies you can lie in it to sneak in bigger files, which I don't like

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -297,18 +297,20 @@ protected IngestionResult ingestFromBlobImpl(BlobSourceInfo blobSourceInfo, Inge
...
-                        "Failed getting blob properties: " + e.getMessage(),
+                        "Failed getting blob properties: " + ExceptionsUtils.getMessageEx(e),
                         e);
             }
         }
 
-        if (blobSourceInfo.getRawSizeInBytes() > MAX_STREAMING_SIZE_BYTES) {
+        if (queuingPolicy.shouldUseQueuedIngestion(blobSize, blobSourceInfo.getRawSizeInBytes(),
+                blobSourceInfo.getCompressionType() != null, ingestionProperties.getDataFormat())) {
             log.info("Blob size is greater than max streaming size ({} bytes). Falling back to queued.", blobSourceInfo.getRawSizeInBytes());
```
AsafMah, line 314: The log isn't accurate anymore

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -379,9 +381,20 @@ protected IngestionResult ingestFromResultSetImpl(ResultSetSourceInfo resultSetS
         }
     }
 
+    private IngestionResult sendStreamToQueuedIngestion(InputStream inputStream, StreamSourceInfo streamSourceInfo, IngestionProperties ingestionProperties,
+            int size) throws IngestionClientException, IngestionServiceException {
+        log.info("Stream size is greater than max streaming size ({} bytes). Falling back to queued.", size);
```
AsafMah, line 386: Isn't it the same log again?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -393,23 +406,31 @@ protected IngestionResult ingestFromStreamImpl(StreamSourceInfo streamSourceInfo
...
         }
 
+        streamSourceInfo.setSourceId(sourceId);
         byte[] streamingBytes;
+        ByteArrayInputStream byteArrayStream;
+
+        if (queuingPolicy.shouldUseQueuedIngestion(streamSourceInfo.getStream().available(), streamSourceInfo.getRawSizeInBytes(),
+                streamSourceInfo.getCompressionType() != null, ingestionProperties.getDataFormat())) {
+            log.info("Stream size ({} bytes) is greater than max streaming size according to policy. Falling back to queued.",
+                    streamSourceInfo.getRawSizeInBytes() > 0 ? streamSourceInfo.getRawSizeInBytes() : streamSourceInfo.getStream().available());
```
AsafMah, line 416: This log and check are again copy-pasted from elsewhere

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -393,23 +406,31 @@ protected IngestionResult ingestFromStreamImpl(StreamSourceInfo streamSourceInfo
...
+        }
+
         try {
-            streamingBytes = IngestionUtils.readBytesFromInputStream(streamSourceInfo.getStream(), MAX_STREAMING_SIZE_BYTES + 1);
+            if (streamSourceInfo.getStream() instanceof ByteArrayInputStream) {
+                // We can't rely on other InputStream implementations of available()
+                byteArrayStream = (ByteArrayInputStream) streamSourceInfo.getStream();
+            } else {
+                streamingBytes = IngestionUtils.readBytesFromInputStream(streamSourceInfo.getStream(),
+                        ManagedStreamingQueuingPolicy.MAX_STREAMING_STREAM_SIZE_BYTES + 1);
```
AsafMah, line 426: Shouldn't this depend on the policy?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingQueuingPolicy.java
```diff
@@ -0,0 +1,66 @@
...
+
+            return (dataSize * BINARY_UNCOMPRESSED_FACTOR) > MAX_STREAMING_UNCOMPRESSED_RAW_SIZE_BYTES;
+        }
+
+        if (compressed) {
+            // Compressed + non-binary
+            return (dataSize * NON_BINARY_FACTOR) > MAX_STREAMING_UNCOMPRESSED_RAW_SIZE_BYTES;
+        }
+
+        if (dataFormat.isJsonFormat()) {
```
AsafMah, line 52: Can it ever get to here? wouldn't it return in the !dataformat.IsCompressible if?

AsafMah (changes requested):

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
```diff
@@ -38,9 +38,11 @@ public static CloseableHttpClient create(HttpClientProperties providedProperties
                 .setMaxConnTotal(properties.maxConnectionTotal())
                 .setMaxConnPerRoute(properties.maxConnectionRoute())
                 .evictExpiredConnections()
-                .evictIdleConnections(properties.maxIdleTime(), TimeUnit.SECONDS)
-                .disableRedirectHandling();
+                .evictIdleConnections(properties.maxIdleTime(), TimeUnit.SECONDS);
 
+        if (properties.isDisableRetries()) {
```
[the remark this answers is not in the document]
ohadbitt: Well unless i missed it - it didnt - and thats why i use it here

ingest/pom.xml
```diff
@@ -107,6 +107,7 @@
                     <ignoredUsedUndeclaredDependencies>
                         <dependency>com.microsoft.azure:msal4j:jar</dependency>
                         <dependency>io.projectreactor:reactor-core:jar</dependency>
+                        <dependency>com.fasterxml.jackson.core:jackson-core:jar</dependency>
```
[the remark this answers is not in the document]
ohadbitt: It just means we get it from Data

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/WellKnownKustoEndpointsData.java
```diff
@@ -44,8 +44,7 @@ private static WellKnownKustoEndpointsData readInstance() {
                 return objectMapper.readValue(resourceAsStream, WellKnownKustoEndpointsData.class);
             }
         } catch (Exception ex) {
-            ex.printStackTrace();
+            throw new RuntimeException("Failed to read WellKnownKustoEndpoints.json");
```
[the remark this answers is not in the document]
ohadbitt: true

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/ExceptionsUtils.java
```diff
@@ -0,0 +1,8 @@
+package com.microsoft.azure.kusto.data.exceptions;
+
+public class ExceptionsUtils {
+    // Useful in IOException, where message might not propagate to the base IOException
+    public static String getMessageEx(Exception e) {
```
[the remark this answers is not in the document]
ohadbitt: we dont remove the top leve l - only if top level is null will we get the message from inner 
Its actually just a bug in my opinion of IOExcepptions

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
ohadbitt: these are automatics

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
ohadbitt: this is not used

ohadbitt: Just a leftover unused code

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
ohadbitt: log yes - not code

ingest/pom.xml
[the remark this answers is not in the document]
AsafMah: I was sure we removed it as a dep because of blacklisting, maybe just the core part is ok? this is confusing me

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/ExceptionsUtils.java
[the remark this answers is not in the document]
AsafMah: Ah ok I see

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/AbstractSourceInfo.java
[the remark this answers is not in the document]
AsafMah: My point is, this wasn't exposed before (IIRC), adding it means we have to support it forever. So is there a need to?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/FileSourceInfo.java
```diff
@@ -23,24 +19,14 @@ public void setFilePath(String filePath) {
...
-
-    public long getRawSizeInBytes() {
-        return rawSizeInBytes;
-    }
-
-    public void setRawSizeInBytes(long rawSizeInBytes) {
-        this.rawSizeInBytes = rawSizeInBytes;
-    }
-
+    // An estimation of the raw (uncompressed, un-indexed) size of the data, for binary formatted files - use only if known
```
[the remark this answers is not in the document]
AsafMah: This is not about v2 or not v2, it's about code organization - Both StreamSourceInfo and FileSourceInfo support Compression, but in Stream it's a part of the class and for file it's calculated outside of the class.

I think it should be part of the abstract class, and FileSourceInfo can define it to be the extension or whatever.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
AsafMah: Then why do you need it in the signature? Shouldn't it always be wrapped and then it's also not a breaking change

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: ok

ingest/pom.xml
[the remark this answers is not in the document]
ohadbitt: im not sure we can have anything good here 
but anyway Spark today releases a shaded Jar - i hope this will obsolete those problems

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
ohadbitt: its in signature of the Impl ..

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
ohadbitt: 1. talked to Yochai - he doesnt mind people changing it 
2. As its a predicate - thought it is a nice Java impl ... will change to interface 

This class will most likely grow in time

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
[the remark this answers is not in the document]
ohadbitt: maybe i should try and make it throw again

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/AbstractSourceInfo.java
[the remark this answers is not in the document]
ohadbitt: It was exposed in File and Blob ... this just opens it in stream 
I think that in V1 it is part of our API whether we like it or not .. in v2 we decided to remove it - if i recall correctly the DM wants to move from size estimations by raw data size to estimation through rows count - although im not sure how its going to work . If you ask me - i think we only need data disk size - not some user estimations and not rows count

ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
AsafMah: I don't remember now - I think what I meant is that you can catch and throw our exception inside the impl.

EIther way it probably doesn't matter if the impl isn't public

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: I mean it's unlikely to grow if we're going to replace most of it for v2, but sure

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
AsafMah: 1. the name is still wrong (it's SourceInfo.ge/settRawSizeInBytes)
2. I think just putting it right next to the previous line with this wording is confusing, since it makes it seem that they are related.

AsafMah (commented):
Basically done, just go over the rest of the comments. Most of them are nothing

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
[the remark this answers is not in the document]
ohadbitt: Did not manage - will delete

data/src/main/java/com/microsoft/azure/kusto/data/ClientFactory.java
```diff
@@ -73,7 +75,9 @@ public static StreamingClient createStreamingClient(ConnectionStringBuilder csb)
      * @throws URISyntaxException if the cluster URL is invalid
      */
     public static StreamingClient createStreamingClient(ConnectionStringBuilder csb, HttpClientProperties properties) throws URISyntaxException {
-        return new ClientImpl(csb, properties);
+        HttpClientProperties httpClientProperties = Optional.ofNullable(properties)
```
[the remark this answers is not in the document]
ohadbitt: closing as i remove disable

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/FileSourceInfo.java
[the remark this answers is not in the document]
ohadbitt: Whats an extension in Java?
Anyway i dont think im going to change this as this is what we have and also how v2 behaves

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/WellKnownKustoEndpointsData.java
```diff
@@ -44,8 +44,7 @@ private static WellKnownKustoEndpointsData readInstance() {
                 return objectMapper.readValue(resourceAsStream, WellKnownKustoEndpointsData.class);
             }
         } catch (Exception ex) {
-            ex.printStackTrace();
+            throw new RuntimeException("Failed to read WellKnownKustoEndpoints.json", ex);
```
AsafMah, line 47: If you want you can annotate with @NotNull

data/src/main/java/com/microsoft/azure/kusto/data/ClientFactory.java
AsafMah, line 78: I still think that's a bit bad, since it's a performance hit (creating a new object), while not providing much readability over a simple ternary.

I know it's copied from another place, but change it there too?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
```diff
@@ -61,19 +63,21 @@ private static <TInner, TWrapper extends ResourceWithSas<TInner>, TOut> TOut res
...
-        throw new IngestionClientException(String.format("%s: All %d retries failed - used resources: %s", actionName, RETRY_COUNT,
-                totalAttributes.stream().map(x -> String.format("%s (%s)", x.get("resource"), x.get("account"))).collect(Collectors.joining(", "))));
+        throw new IngestionClientException(String.format("%s: All %d retries failed with last error: %s\n. Used resources: %s", actionName, RETRY_COUNT,
+                totalAttributes.stream().map(x -> String.format("%s (%s)", x.get("resource"), x.get("account"))).collect(Collectors.joining(", ")),
+                ExceptionsUtils.getMessageEx(ex)));
     }
 
     public static void postToQueueWithRetries(ResourceManager resourceManager, AzureStorageClient azureStorageClient, IngestionBlobInfo blob)
-            throws IngestionClientException, IngestionServiceException {
+            throws IngestionClientException, IngestionServiceException, JsonProcessingException {
```
AsafMah, line 76: That's breaking right?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/FileSourceInfo.java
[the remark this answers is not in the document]
AsafMah: Extension as in file extension.

It's about SOLID principles.

Do you see what I mean?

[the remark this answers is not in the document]
ohadbitt: No
Why is it different from the .Net v2 - you can see there the user cant set the file compression type - we had a discussion about it

data/src/main/java/com/microsoft/azure/kusto/data/ClientFactory.java
ohadbitt: OK
This is a very scala like code- so i guess im used to it - in scala you are encouraged to use this and dont have ternary 
I dont see how creating a new object on a 1 time created client is a performance hit.

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/WellKnownKustoEndpointsData.java
ohadbitt: who? instance? 
May I not?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
ohadbitt: no

AsafMah (approved):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
AsafMah: Why not? it's a public method on a public class

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/WellKnownKustoEndpointsData.java
AsafMah: the readInstance methods.

Just so callers can know to not expect null

you don't have to

CHANGELOG.md
```diff
@@ -4,7 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [6.0.0] - 2024-09-21
+## [6.0.0] - 2024-10-27
+### Added
+- A new heuristic for choosing between queuing and streaming in Managed streaming client, the default behavior is using
```
AsafMah, line 9: I think this bullet got a bit jumbled, look it over and rewrite it.

CHANGELOG.md
```diff
@@ -4,7 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [6.0.0] - 2024-09-21
+## [6.0.0] - 2024-10-27
```
AsafMah, line 7: Unreleased for next version, for easier merge

data/src/main/java/com/microsoft/azure/kusto/data/ClientFactory.java
```diff
@@ -83,7 +83,7 @@ public static StreamingClient createStreamingClient(ConnectionStringBuilder csb,
      * is customized with the given properties.
      *
      * @param csb the connection string builder
-     * @param httpClient HTTP client
+     * @param httpClient HTTAP client
```
AsafMah, line 86: Mistake?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/FileSourceInfo.java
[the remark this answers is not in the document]
AsafMah: Whatever, let's fight about it another time

AsafMah:
Go over the pending comments, but approved in principle

AsafMah (approved):

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceAlgorithms.java
ohadbitt: you are right - im not sure why its public tho

PR #370, opened 2024-07-21
HttpClient - Proxy planner support 

ohadbitt:
### Added
- Proxy planner support for http client


samples/src/main/java/Proxy.java
```diff
@@ -0,0 +1,17 @@
+import com.microsoft.azure.kusto.data.Client;
+import com.microsoft.azure.kusto.data.ClientFactory;
+import com.microsoft.azure.kusto.data.HttpClientProperties;
+import com.microsoft.azure.kusto.data.auth.ConnectionStringBuilder;
+import com.microsoft.azure.kusto.data.http.SimpleProxyPlanner;
+
+public class Proxy {
+    public static void main(String[] args) {
```
AsafMah, line 8: Also change CHANGELOG.md with changes and sample

pom.xml
```diff
@@ -35,7 +35,7 @@
         <revision>5.1.0</revision> <!-- CHANGE THIS to adjust project version-->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <java.version>1.8</java.version>
-        <azure-bom-version>1.2.24</azure-bom-version>
```
AsafMah, line 38: why?

AsafMah (changes requested):

ingest/pom.xml
```diff
@@ -105,7 +105,6 @@
                         <dependency>org.xerial:sqlite-jdbc</dependency>
                     </ignoredUnusedDeclaredDependencies>
                     <ignoredUsedUndeclaredDependencies>
-                        <dependency>com.microsoft.azure:msal4j:jar</dependency>
```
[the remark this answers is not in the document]
ohadbitt: not used in ingest library

data/src/main/java/com/microsoft/azure/kusto/data/http/SimpleProxyPlanner.java
```diff
@@ -0,0 +1,48 @@
+package com.microsoft.azure.kusto.data.http;
+
+import org.apache.http.HttpHost;
+import org.apache.http.HttpRequest;
+import org.apache.http.impl.conn.DefaultRoutePlanner;
+import org.apache.http.impl.conn.DefaultSchemePortResolver;
+import org.apache.http.protocol.HttpContext;
+
+public class SimpleProxyPlanner extends DefaultRoutePlanner {
```
AsafMah, line 10: So you don't need the class anymore - at least not here, since it is only used it the test.

Even in the test you can replace it with a lambda (java can automatically accept lambdas for 1-function interfaces).

Might be worth to keep as an example somewhere, but not in the middle of the code.

AsafMah (changes requested):
Also the changelog

AsafMah (approved):
Looks good.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -702,4 +704,21 @@ void testStreamingIngestFromBlob() throws IngestionClientException, IngestionSer
...
+                new URI(engineCsb.getClusterUrl()).getHost(),
+                "login.microsoftonline.com"
+        };
+
+        HttpClientProperties providedProperties = HttpClientProperties.builder().routePlanner(new SimpleProxyPlanner("localhost", 8080, "http", prefixes))
+                .build();
+        try (Client client = ClientFactory.createClient(engineCsb, providedProperties)) {
+            KustoOperationResult execute = client.execute(".show version");
+        } catch (Exception e) {
+            Assertions.fail(e);
```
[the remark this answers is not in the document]
ohadbitt: talked F2F Will leave as is

PR #371, opened 2024-07-24
Fix client app and version tracing

yihezkel:
### Fixed
Fix client app and version tracing


ag-ramachandran (approved):

ag-ramachandran:
Would this be version 5.1.1 for us to include in the other dependent jars

AsafMah (approved):

PR #378, opened 2024-08-01
fixes token fetch when account provided and bumps version

The-Funk:
### Fixed

MsalTokenProviders fail when getting silent parameters due to changes to a deprecated method in msal4j.

This PR replaces a usage of that deprecated method.

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
@@ -93,7 +93,7 @@ public abstract class MsalTokenProviderBase extends CloudDependentTokenProviderB
                 authorityUrl = firstPartyAuthorityUrl;
             }
 
-            return SilentParameters.builder(scopes).account(account).authorityUrl(authorityUrl).build();
+            return SilentParameters.builder(scopes, account).authorityUrl(authorityUrl).build();
         }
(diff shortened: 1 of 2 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/auth/MsalTokenProviderBase.java
```diff
@@ -93,7 +93,7 @@ SilentParameters getSilentParameters(Set<IAccount> accountSet) {
                 authorityUrl = firstPartyAuthorityUrl;
             }
 
-            return SilentParameters.builder(scopes).account(account).authorityUrl(authorityUrl).build();
+            return SilentParameters.builder(scopes, account).authorityUrl(authorityUrl).build();
```
ohadbitt, line 96: Please don’t leave it like that this is regression 
Actually account is relevant i think only for user auth
Prompt, azcli, device auth are the ones that should have this silent behaviour with account

The-Funk: @ohadbitt the line above is the fix.

When account is null, I'm not sure how to call the SilentParameters without using the deprecated method, so I left the line below that alone.

But the line you pointed out is not a regression, it is the fix replacing the deprecated SilentParameters.builder(scopes) method. This method no longer works as of azure-core >= 1.2.23

Tokens fail to download if you use azcli or service principal auth on version 5.1.1 due to this issue.

ohadbitt (commented):
acknowledged + looking good

PR #380, opened 2024-08-04
Stop tests from depending on table row store being sealed

yihezkel:
### Changed
Stop tests from depending on table row store being sealed, to shorten E2E runtime, and prevent timeouts even with the new HTTP client.


AsafMah:
In a lot of other places in the SDK and others, we run a ".clear database cache streamingingestion schema" 

Could it work for this case?

yihezkel:
> In a lot of other places in the SDK and others, we run a ".clear database cache streamingingestion schema" 
> 
> Could it work for this case?

Valentina said that command wouldn't help

AsafMah (approved):

PR #382, opened 2024-08-21
Add support for configuring SSL/TLS protocols in HttpClient

AsafMah:
Introduce a new `supportedProtocols` field in `HttpClientProperties` to allow specifying SSL/TLS protocols.


ohadbitt:
i hope this will  not be difficult in the new client

ohadbitt (approved):

PR #385, opened 2024-08-26
Changelog + run tests for #378

ohadbitt:


AsafMah:
The date should be the date we released it, so if you want to release it today, change it to today.

If you have more changes waiting, it should be Unreleased (for easier merges)

AsafMah (approved):
note comment

PR #388, opened 2024-09-16
Implements azure core http client

The-Funk:


pom.xml
```diff
@@ -32,31 +32,30 @@
     </developers>
 
     <properties>
-        <revision>5.2.0</revision> <!-- CHANGE THIS to adjust project version-->
+        <revision>6.0.0</revision> <!-- CHANGE THIS to adjust project version-->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <java.version>1.8</java.version>
-        <azure-bom-version>1.2.25</azure-bom-version>
+        <azure-bom-version>1.2.24</azure-bom-version>
```
AsafMah, line 38: Why did you lower the version?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -114,54 +110,52 @@ public static CloudInfo retrieveCloudInfoForCluster(String clusterUrl,
...
 
     private static CloudInfo fetchImpl(String clusterUrl, @Nullable HttpClient givenHttpClient) throws URISyntaxException, IOException, DataServiceException {
         CloudInfo result;
         HttpClient localHttpClient = givenHttpClient == null ? HttpClientFactory.create(null) : givenHttpClient;
         try {
-            HttpGet request = new HttpGet(UriUtils.appendPathToUri(clusterUrl, METADATA_ENDPOINT));
-            request.addHeader(HttpHeaders.ACCEPT_ENCODING, "gzip,deflate");
-            request.addHeader(HttpHeaders.ACCEPT, "application/json");
+            HttpRequest request = new HttpRequest(HttpMethod.GET, UriUtils.appendPathToUri(clusterUrl, METADATA_ENDPOINT));
+            request.setHeader(HttpHeaderName.ACCEPT_ENCODING, "gzip");
```
AsafMah, line 121: we can't handle deflate anymore?

data/src/main/java/com/microsoft/azure/kusto/data/auth/HttpClientWrapper.java
```diff
@@ -135,34 +36,32 @@ private static boolean isNotContentLength(String name) {
 
     // Implementation of the synchronous HttpClient
     @Override
-    public IHttpResponse send(com.microsoft.aad.msal4j.HttpRequest httpRequest) throws Exception {
-        HttpUriRequest request;
-
-        // Translating the request
-
-        String uri = httpRequest.url().toString();
+    public IHttpResponse send(com.microsoft.aad.msal4j.HttpRequest httpRequest) {
```
AsafMah, line 39: Do we still need the IHttpClient impl? what do we use it for?

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
```diff
@@ -1,108 +1,71 @@
...
-                    return Long.parseLong(value) * 1000;
-                }
-            }
-            // otherwise keep alive for default seconds
-            return defaultKeepAlive * 1000L;
+        if (properties.getProxy() != null) {
+            options.setProxyOptions(properties.getProxy());
         }
+
+        // Todo: Is the per route connection maximum needed anymore?
```
AsafMah, line 66: I think some of the new options are missing - the routeplanner and supported protocols, can you make sure it is paired or at least specify all of the changes?

data/pom.xml
```diff
@@ -167,7 +167,7 @@
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
```
AsafMah, line 168: Can we not remove the dep on apache http client?
Or even just some of the components?

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -114,54 +110,52 @@ public static CloudInfo retrieveCloudInfoForCluster(String clusterUrl,
...
             // trace CloudInfo.httpCall
-            HttpResponse response = MonitoredActivity.invoke(
-                    (SupplierOneException<HttpResponse, IOException>) () -> localHttpClient.execute(request),
-                    "CloudInfo.httpCall");
-            try {
-                int statusCode = response.getStatusLine().getStatusCode();
-                if (statusCode == HttpStatus.SC_OK) {
-                    String content = EntityUtils.toString(response.getEntity());
+            try (HttpResponse response = MonitoredActivity.invoke(
+                    (SupplierOneException<HttpResponse, IOException>) () -> localHttpClient.sendSync(request, Context.NONE),
```
AsafMah, line 126: Should add a fixme to make this async

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -3,83 +3,67 @@
...
     public static final String STREAMING_VERSION = "v1";
     private static final String DEFAULT_DATABASE_NAME = "NetDefaultDb";
-    private static final Long COMMAND_TIMEOUT_IN_MILLISECS = TimeUnit.MINUTES.toMillis(10);
-    private static final Long QUERY_TIMEOUT_IN_MILLISECS = TimeUnit.MINUTES.toMillis(4);
-    private static final Long STREAMING_INGEST_TIMEOUT_IN_MILLISECS = TimeUnit.MINUTES.toMillis(10);
-    private static final int CLIENT_SERVER_DELTA_IN_MILLISECS = (int) TimeUnit.SECONDS.toMillis(30);
-    public static final String CLIENT_VERSION_HEADER = "x-ms-client-version";
-    public static final String APP_HEADER = "x-ms-app";
-    public static final String USER_HEADER = "x-ms-user";
+
```
AsafMah, line 39: What happened to all of the timeout logic? do now all of the requests have the same timeout?

data/src/main/java/com/microsoft/azure/kusto/data/instrumentation/SupplierNoException.java
```diff
@@ -0,0 +1,16 @@
+package com.microsoft.azure.kusto.data.instrumentation;
+
+/**
+ * Supplier is a functional interface that can be used to execute data operations.
+ * @param <T>
+ */
+@FunctionalInterface
+public interface SupplierNoException<T> {
```
AsafMah, line 8: Is there not a built-in type for this?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -97,32 +81,69 @@ public ClientImpl(ConnectionStringBuilder csb, CloseableHttpClient httpClient, b
...
+        KustoRequest kr = new KustoRequest(command, database, properties);
+        return executeQueryAsync(kr);
     }
 
-    @Override
-    public KustoOperationResult execute(String database, String command) throws DataServiceException, DataClientException {
-        return execute(database, command, null);
+    Mono<KustoOperationResult> executeQueryAsync(@NotNull KustoRequest kr) {
+        if (kr.getCommandType() != CommandType.QUERY) {
+            kr.setCommandType(CommandType.QUERY);
```
AsafMah, line 94: Maybe we should throw if someone wants a query but doesn't set the type to query?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -97,32 +81,69 @@ public ClientImpl(ConnectionStringBuilder csb, CloseableHttpClient httpClient, b
...
+        if (kr.getCommandType() != CommandType.ADMIN_COMMAND) {
+            kr.setCommandType(CommandType.ADMIN_COMMAND);
+        }
+        return executeAsync(kr);
+    }
+
+    private Mono<KustoOperationResult> executeAsync(KustoRequest kr) {
+
+        Mono<String> resultMono = executeToJsonAsync(kr)
+                .onErrorContinue((err, src) -> LOGGER.error("Error coming from src {}", src, err));
```
AsafMah, line 115: Does on error continue stop the execution of further things down the pipeline? what about the rest of the function?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -194,43 +267,36 @@ public String executeToJsonResult(String database, String command) throws DataSe
...
-            endpointValidated = true;
+    private void validateEndpoint() throws DataServiceException, DataClientException {
+        try {
+            if (!endpointValidated) {
+                KustoTrustedEndpoints.validateTrustedEndpoint(clusterUrl,
+                        CloudInfo.retrieveCloudInfoForCluster(clusterUrl).getLoginEndpoint());
+                endpointValidated = true;
+            }
+        } catch (KustoClientInvalidConnectionStringException e) {
+            throw new DataClientException(clusterUrl, e.getMessage(), e);
```
AsafMah, line 299: Maybe we want an inhereted type of exception, so users could catch it specficially?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -260,52 +326,71 @@ public KustoOperationResult executeStreamingIngestFromBlob(String database, Stri
...
+        String contentEncoding = null;
+        String contentType;
+        if (isStreamSource) {
+            contentEncoding = "gzip";
         }
-    }
 
-    private Long populateHeadersAndGetTimeout(ClientRequestProperties properties, Map<String, String> headers) throws DataClientException {
-        Long timeoutMs = null;
+        // This was a separate method but was moved into the body of this method because it performs a side effect
```
AsafMah, line 340: Maybe we'll move it to a utils class or something? this method is long enough

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -260,52 +326,71 @@ public KustoOperationResult executeStreamingIngestFromBlob(String database, Stri
...
+                // We use UncloseableStream to prevent HttpClient From closing it
+                data = BinaryData.fromStream(new UncloseableStream(stream));
+                contentType = "application/octet-stream";
+            } else {
+                data = BinaryData.fromString(new IngestionSourceStorage(blobUrl).toString());
+                contentType = "application/json";
+            }
+
+            // Build the tracing object
+            HttpTracing tracing = HttpTracing
```
AsafMah, line 364: I assume we can't unify this method with the other one better?

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -339,139 +424,61 @@ public InputStream executeStreamingQuery(String database, String command) throws
...
     public String getClusterUrl() {
         return clusterUrl;
     }
 
+    ClientDetails getClientDetails() {
+        return clientDetails;
+    }
+
     @Override
     public void close() throws IOException {
```
AsafMah, line 481: Do we still want it to be closeable in this case?

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -324,18 +324,7 @@ public Map<String, String> getTracingAttributes() {
...
-     * @Deprecated use {@link #getTimeoutAsCslTimespan(Object)} instead.
-     */
-    @Deprecated
-    String getTimeoutAsString(Object timeoutObj) {
-        return getTimeoutAsCslTimespan(timeoutObj);
-    }
-
-    /**
-     * Gets the amount of time a query may execute on the service before it times out, formatted as a KQL timespan.
+     * Gets the amount of time a query may execute on the service before it times out, formatted as a krL timespan.
```
AsafMah, line 327: krL? (next method too)

ingest/src/test/java/com/microsoft/azure/kusto/ingest/AsyncE2ETest.java
```diff
@@ -0,0 +1,71 @@
+// package com.microsoft.azure.kusto.ingest;
```
AsafMah, line 1: What is this file?

pom.xml
The-Funk: Mistake. Will fix

data/src/main/java/com/microsoft/azure/kusto/data/auth/HttpClientWrapper.java
The-Funk: IHttpClient comes from MSAL4J. It sends HTTP requests that MSAL4J recognizes. In an upcoming PR I'd like to reduce it's necessity. We should probably get on a call about that though.

data/pom.xml
The-Funk: Not yet unfortunately, the URIBuilder that is used extensively in the Kusto library is in the client package and not the core package, so we have to keep everything for now. It was moved to the Apache HTTP core in version 5, and there are other URI Builder libraries out there that are very small footprint, however last time I tried updating to the Apache V5 URIBuilder it caused failures with those really weird URIs in the tests.

That is basically the only remaining Apache component from the http-client package that we use but it would require careful removal. I would like to remove it at a later date (soon).

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
The-Funk: That appears to be the case. The azure-core clients seem to only support a global timeout on the client. No timeout logic otherwise.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
The-Funk: Should be able to, I can readd.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
The-Funk: That would be a major behavior change compared with what you have currently. Command type has never been exposed to the user before. If we exposed the KustoRequest object they would have the capability easily, otherwise we would have to create an even more overloaded method with command type and the request is already very overloaded as it is. We have the execute method that takes commandtype but it is not public because it was never public before.

But at that point why do we have separate methods for executeQuery and executeMgmt? You had asked me specifically to deprecate the generic "execute" methods and separate methods into executeQuery and executeMgmt. If you call executeMgmt, does that not imply that you want to perform a management action? If you call executeQuery does that not imply you want to execute a query?

My recommendation would be either do the above or expose the request object so that we don't further inundate developers with even more overloaded methods, and make maintainance on our side even more difficult by forcing ourselves to method chain even more than we already are.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
The-Funk: onErrorContinue continues the pipeline in the event of an error. It was there for testing but I kind of want to rip all the async code out because I couldn't get it working.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
The-Funk: I'm not happy with our current open-telemetry implementation, so I kept it decoupled to make it easy to remove later. I'd like to get rid of this class entirely. The azure-core library has its own built-in open-telemetry implementation that might be a better fit.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -704,22 +716,4 @@ void testStreamingIngestFromBlob() throws IngestionClientException, IngestionSer
             }
         }
     }
-
-    @Test
-    void testProxyPlanner() throws URISyntaxException {
```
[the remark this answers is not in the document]
The-Funk: The routeplanner is an Apache specific concept.

No other HTTP Client has a route planner, including the default Netty HTTP client and OkHttp, as it's not really feasible to plan routes beyond indicating a proxy. It is not supportable outside of the Apache client and it was not included in Apache HttpClient5 either. It was last available in Apache HttpClient4. My recommendation would be to abandon it.

The ProxyOptions in azure http client should provide significant proxy configuration capabilities.

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientProperties.java
```diff
@@ -118,9 +129,9 @@ public static class HttpClientPropertiesBuilder {
         private Integer maxKeepAliveTime = 120;
         private Integer maxConnectionsTotal = 40;
         private Integer maxConnectionsPerRoute = 40;
-        private HttpHost proxy = null;
```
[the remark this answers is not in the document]
The-Funk: I would avoid using the builder in your comment. The word Netty should be considered a dirty word because it is a specific implementation and locks the customer into using the Netty client and locks us into using the Netty client as well. 

If you wanted to use OkHttp instead of Netty, the code in the comment above would have to be replaced first.

Inversion of dependencies says to depend on interfaces instead of implementations, in this case everything we have done thus far in this branch uses the azure-http-client interface and does not lock you in to a specific implementation.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
The-Funk: I'm not sure where that came from but fixed now.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/AsyncE2ETest.java
The-Funk: deleted

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
The-Funk: fixed

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -0,0 +1,185 @@
...
+        // Execute and get the response
+        try (HttpResponse response = httpClient.sendSync(request, Context.NONE)) {
+            return processResponseBody(response);
+        }
+    }
+
+    protected Mono<String> postAsync(HttpRequest request) {
+        // Execute and get the response
+        return httpClient.send(request)
+                .onErrorContinue((err, src) -> LOGGER.error("Error coming from src {}", src, err))
```
ohadbitt, line 44: so this is not what you need here . You probably wanted to use doOnError(). and you only need to do it once as its going to propagate all they way

pom.xml
```diff
@@ -32,31 +32,30 @@
...
         <resilience4j.version>1.7.1</resilience4j.version>
         <io.vavr.version>0.10.4</io.vavr.version>
         <!-- Test dependencies -->
         <bouncycastle.version>1.77</bouncycastle.version>
         <jsonassert.version>1.5.0</jsonassert.version>
-        <sqlite-jdbc.version>3.45.2.0</sqlite-jdbc.version>
+        <sqlite-jdbc.version>3.45.3.0</sqlite-jdbc.version>
         <annotations.version>24.1.0</annotations.version>
-
+        <reactor.test.version>3.6.6</reactor.test.version>
```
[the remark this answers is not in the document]
The-Funk: fixed

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -117,74 +130,67 @@ public static void setUp() {
...
+            Assertions.assertNotNull(DB_NAME, "DB name was not set");
+            queryClient.executeToJsonResult(DB_NAME, String.format(".drop table %s ifexists", tableName));
         } catch (Exception ex) {
             Assertions.fail("Failed to drop table", ex);
         }
     }
 
     private static boolean isManualExecution() {
         return false;
-        // return System.getenv("CI_EXECUTION") == null || !System.getenv("CI_EXECUTION").equals("1");
```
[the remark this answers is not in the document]
The-Funk: fixed

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -117,74 +130,67 @@ public static void setUp() {
...
     }
 
     private static void createTableAndMapping() {
         try {
-            queryClient.executeToJsonResult(databaseName, String.format(".drop table %s ifexists", tableName));
```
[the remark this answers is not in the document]
The-Funk: This line still exists.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -84,9 +96,9 @@ class E2ETest {
     public static void setUp() {
         tableName = "JavaTest_" + new SimpleDateFormat("yyyy_MM_dd_hh_mm_ss_SSS").format(Calendar.getInstance().getTime()) + "_"
                 + ThreadLocalRandom.current().nextInt(Integer.MAX_VALUE);
-        principalFqn = String.format("aadapp=%s;%s", appId, tenantId);
+        principalFqn = String.format("aadapp=%s;%s", APP_ID, TENANT_ID);
 
-        ConnectionStringBuilder dmCsb = createConnection(System.getenv("DM_CONNECTION_STRING"));
```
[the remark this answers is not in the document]
The-Funk: These names are unchanged, I just moved all references to environment variables to the top of the E2ETest class so that it is easier for newcomers to the codebase to see exactly which environment variables they need in order to run the E2ETest. Great care was taken to ensure none of the environment variable names changed.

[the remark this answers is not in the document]
The-Funk: This was a QoL thing I did because as a new developer coming into this codebase **the E2ETest was completely un-runnable for me.** I had no idea what environment variables I needed nor the format that said variables needed to be in. This made the task of getting the E2ETest class tests to even **run** a monumental challenge, let alone get them to pass. When you spread usage of the environment throughout the entire test class it's very hard to see the requirements to run the tests, so that every test run I would be greeted by yet another new environment variable I hadn't seen before that would cause E2ETest to fail. Now they're all in one place and someone external like myself could easily see what they need, upfront, and not waste an hour trying to track down what environment variables are needed to run E2ETest.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -397,10 +390,9 @@ private static void throwNoResultException(String baseMessage, Boolean refreshed
     @Override
     public void reportIngestionResult(ResourceWithSas<?> resource, boolean success) {
         if (storageAccountSet == null) {
-            log.error("StorageAccountSet is null, so can't report ingestion result");
-        } else {
-            storageAccountSet.addResultToAccount(resource.getAccountName(), success);
+            log.warn("StorageAccountSet is null");
         }
+        storageAccountSet.addResultToAccount(resource.getAccountName(), success);
```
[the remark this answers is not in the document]
The-Funk: woops! meant to make that initial if a guard clause. should be fixed now.

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientProperties.java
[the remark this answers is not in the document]
The-Funk: I've opened an issue over on azure-core for this.
https://github.com/Azure/azure-sdk-for-java/issues/42083

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpClientFactory.java
```diff
@@ -1,108 +1,71 @@
...
-     * response header; if the response doesn't contain a {@code Keep-Alive} header, the client will use a configurable
-     * keep-alive period.
-     */
-    static class CustomConnectionKeepAliveStrategy implements ConnectionKeepAliveStrategy {
+        // Set Keep-Alive headers if they were requested.
+        // NOTE: Servers are not obligated to honor client requested Keep-Alive values
+        if (properties.isKeepAlive()) {
+            Header keepAlive = new Header(HttpHeaderName.CONNECTION.getCaseSensitiveName(), "Keep-Alive");
+            // Keep-Alive is Non-standard from the client so core does not have an enum for it
+            Header keepAliveTimeout = new Header("Keep-Alive", "timeout=" + properties.maxKeepAliveTime());
```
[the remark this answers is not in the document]
The-Funk: If the client negotiates HTTP/2, the Keep-Alive header will be stripped because connection specific headers are forbidden in HTTP/2.

HTTP/2 is basically infinite Keep-Alive. A single connection is reused to remove the overhead of establishing new connections

data/pom.xml
The-Funk: Moved all URIBuilder code into URIUtils to make it easier to rip out the Apache builder in the near future.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
The-Funk: I thought about this comment for a day or two and came to the conclusion that if we extend DataClientException or DataServiceException for most exception cases:

1. We don't have to touch method signatures when throwing DataClient or DataService
2. We don't have to catch the exceptions that extend DataClient or DataService and rethrow them
3. Customers can then catch more specific exceptions if that's something they want, and then do with that additional context what they will

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
The-Funk: For some reason reactor is freezing, so I put in onErrorContinue's in the hopes that an error would propagate and I would see what the issue is. Commented the async implementations out for now because they are not working yet.

data/pom.xml
AsafMah: So should we just move to http core version 5 then? any reason not to?

The-Funk: Between V4 and V5 of the Apache URIBuilder either there was a regression or some of the URIs we're testing are invalid. If you upgrade to V5 most URIs in our test cases will pass however one or two fail.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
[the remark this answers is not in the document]
ohadbitt: for some reason here and the above conversation - i dont see it ..

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -307,13 +311,17 @@ void testShowPrincipals() {
...
         KustoOperationResult result = null;
-        boolean found = false;
         try {
-            result = localQueryClient.execute(databaseName, String.format(".show database %s principals", databaseName));
-            // result = localQueryClient.execute(databaseName, String.format(".show version"));
+            result = localQueryClient.executeMgmt(DB_NAME, String.format(".show database %s principals", DB_NAME));
         } catch (Exception ex) {
-            Assertions.fail("Failed to execute show database principals command", ex);
+            Assertions.fail("Failed to execute show database principals command. " +
+                    "Is USERNAME_HINT set to your email in your environment variables?", ex);
```
ohadbitt, line 318: move this as comment to the testCreateWithUserPrompt and testCreateWithConnectionStringAndUserPrompt- This method here is not used only by the prompt test

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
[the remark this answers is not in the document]
ohadbitt: We kind of test here the proxy capabilities of the http client - please use the proxy options to achive same results

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
The-Funk: done :)

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    private static void closeResourcesIfNeeded(boolean returnInputStream, HttpResponse httpResponse) {
+        // If we close the resources after returning the InputStream to the caller, they won't be able to read from it
+        if (!returnInputStream) {
+            if (httpResponse != null) {
+                httpResponse.close();
+            }
+        }
+    }
```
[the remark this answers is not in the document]
The-Funk: This is being done since the response isn't encapsulated in a try-with-resources due to usage of the response object in the catch . Let me see if there's a way to implement a try-with-resources. We could probably move some of the error handling inside the try-with-resources and in that way get rid of this but I think the way we have it now, we need this to execute inside the finally block in order to close the HTTP response.

data/src/main/java/com/microsoft/azure/kusto/data/Client.java
```diff
@@ -5,30 +5,36 @@
...
 
     KustoOperationResult executeQuery(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException;
 
+    @Deprecated
     KustoOperationResult executeMgmt(String command) throws DataServiceException, DataClientException;
 
+    @Deprecated
     KustoOperationResult executeMgmt(String database, String command) throws DataServiceException, DataClientException;
 
     KustoOperationResult executeMgmt(String database, String command, ClientRequestProperties properties) throws DataServiceException, DataClientException;
```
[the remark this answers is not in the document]
The-Funk: That's a very good point. I marked these deprecated on request without thinking about it too much, but removing the options with less arguments would involve a lot of
`executeMgmt(null, ".show tables", null)` for the users.

I'm still gung-ho that the proper answer is the request object. Java doesn't support named parameters like some of the other SDK languages, but an object would give us maximum flexibility. Or I can remove the deprecated tag and we can continue supporting the method chaining if that's preferred.. @AsafMah what do you think?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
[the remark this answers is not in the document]
The-Funk: Line 149 (under this review box in GitHub) 
return CI_EXECUTION (variable holding the system environment variable, moved to top of class as a private static final constant) and then same variable tested for equals 1

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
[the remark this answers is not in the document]
The-Funk: Ah you're talking about the policy. My misunderstanding. I removed that so we could verify that the teardown isn't timing out. I will put it back.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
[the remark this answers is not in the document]
The-Funk: @ohadbitt What are you running on port 8080 of your localhost to forward connections to your test cluster? I'll need to set my system up similarly so that it forwards connections to my test ADX cluster. Python HTTP server or something?

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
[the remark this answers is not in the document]
ohadbitt: my mistake - i see now what is the use case - please change the comment to explain this is used in streaming query where the response is returned to the user

data/src/main/java/com/microsoft/azure/kusto/data/Client.java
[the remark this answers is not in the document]
AsafMah: I don't think we need an object here.
Would an overload without a DB not be ok?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -151,8 +158,12 @@ private static boolean isManualExecution() {
 
     private static void createTableAndMapping() {
         try {
-            queryClient.executeToJsonResult(DB_NAME, String.format(".drop table %s ifexists", tableName));
+            queryClient.executeToJsonResult(DB_NAME, String.format(".drop table %s ifexists ", tableName));
```
AsafMah, line 161: NIT: why the trailing space?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
```diff
@@ -151,8 +158,12 @@ private static boolean isManualExecution() {
...
             queryClient.executeToJsonResult(DB_NAME, String.format(".create table %s %s", tableName, tableColumns));
```
AsafMah, line 162: Maybe out of scope here: do we want to check the results?

ohadbitt: Out of scope
I really dont undersantd why we started checking the results elsewhere - did engine started to return valid results on failures?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -150,7 +143,7 @@ private void refreshIngestionResources() throws IngestionClientException, Ingest
                     IngestionResourceSet newIngestionResourceSet = new IngestionResourceSet();
                     Retry retry = Retry.of("get ingestion resources", taskRetryConfig);
                     CheckedFunction0<KustoOperationResult> retryExecute = Retry.decorateCheckedSupplier(retry,
-                            () -> client.execute(Commands.INGESTION_RESOURCES_SHOW_COMMAND));
+                            () -> client.executeMgmt(Commands.INGESTION_RESOURCES_SHOW_COMMAND));
```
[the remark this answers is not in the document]
The-Funk: Yes, but I'll be removing the deprecated annotation so this won't be a problem.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -194,39 +243,28 @@ public String executeToJsonResult(String database, String command) throws DataSe
...
-        if (StringUtils.isEmpty(command)) {
-            throw new IllegalArgumentException("Command is empty");
-        }
-        command = command.trim();
-        CommandType commandType = determineCommandType(command);
-        long timeoutMs = determineTimeout(properties, commandType, clusterUrl);
-        // TODO save the uri once - no need to format everytime
-        String clusterEndpoint = String.format(commandType.getEndpoint(), clusterUrl);
+        KustoRequest kr = new KustoRequest(command, database, properties);
+        return executeToJsonResult(kr);
```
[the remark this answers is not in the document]
The-Funk: The method takes a KustoRequest object, encapsulating all of the possible input state needed in order to execute the request.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -260,52 +298,71 @@ public KustoOperationResult executeStreamingIngestFromBlob(String database, Stri
...
             }
         }
-        if (timeoutMs == null) {
-            timeoutMs = STREAMING_INGEST_TIMEOUT_IN_MILLISECS;
-        }
 
-        return timeoutMs;
+        try (InputStream ignored = (isStreamSource && !leaveOpen) ? stream : null) {
+            // Validate the endpoint
+            validateEndpoint();
```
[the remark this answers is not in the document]
The-Funk: I agree but will probably hold that until later. The endpoint is stored in global state of the client, not the request, so validating the endpoint really only needs done once, not on every request. This has always been done and we talked about changing that but I need to think of a nice and clean way to make that change.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -339,139 +396,57 @@ public InputStream executeStreamingQuery(String database, String command) throws
...
-
-        headers.put("x-ms-client-request-id", clientRequestId);
-
-        headers.put("Connection", "Keep-Alive");
-
-        UUID activityId = UUID.randomUUID();
-        String activityContext = String.format("%s%s/%s, ActivityId=%s, ParentId=%s, ClientRequestId=%s", JAVA_INGEST_ACTIVITY_TYPE_PREFIX, activityTypeSuffix,
-                activityId, activityId, activityId, clientRequestId);
-        headers.put("x-ms-activitycontext", activityContext);
+    public InputStream executeStreamingQuery(@NotNull KustoRequest kr) throws DataServiceException, DataClientException {
```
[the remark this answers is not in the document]
The-Funk: It makes sense to have the request object because the object encapsulates all of the necessary state which is input for the executeStreamingQuery method. This means that if the input state needs to change in the future, the method signature does not, and therefore changes can be made without impacting users of the existing API or adding any additional method chaining. Instead you just add to your object whatever additional state you may need, allowing for the KustoRequest to be an abstract concept, and then you can easily use BiConsumers and functional paradigms like what you find in Project Reactor to write reactive code, which is designed to work with callbacks that take a singular input and provide a singular output.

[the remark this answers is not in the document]
The-Funk: I did just change this to private though

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -339,139 +396,57 @@ public InputStream executeStreamingQuery(String database, String command) throws
...
 
-        // replace non-ascii characters in header values with '?'
-        headers.replaceAll((_i, v) -> v == null ? null : v.replaceAll("[^\\x00-\\x7F]", "?"));
-        return headers;
-    }
+        // Validate and optimize the query object
+        kr.validateAndOptimize();
```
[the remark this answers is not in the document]
The-Funk: We could but then we'd have to set content type explicitly in prepareRequest. In this instance the content type is set explicitly on the builder, in the other method I used the implicit content type which is just JSON. This would save lines though. The other thing is the prepareRequest assumes you don't want to set any special headers, whereas this one sets special headers for streaming content. I probably never should have made a prepareRequest method in the first place but I figured it would be useful for testability. It's not there for reuse so much as its existence is for breaking down what's happening step by step.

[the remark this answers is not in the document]
The-Funk: Maybe renaming that method to prepareJSONRequest would be better and then creating a prepareStreamingRequest method for this logic would make more sense?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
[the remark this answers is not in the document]
The-Funk: Removed deprecated annotations.

data/src/main/java/com/microsoft/azure/kusto/data/auth/HttpClientWrapper.java
```diff
@@ -40,93 +27,7 @@ public HttpClientWrapper(HttpClient httpClient) {
...
-            String contentType = httpRequest.getHeaders().getValue("Content-Type");
-            entityEnclosingRequest.setEntity(new InputStreamEntity(isPipe, contentLength == null ? -1 : Long.parseLong(contentLength),
-                    contentType == null ? null : ContentType.parse(contentType)));
-        }
-
-        // The types of the Monos are different, but we ignore the results anyway (since we only care about the input stream) so this is fine.
-        return Mono.create(monoSink -> {
-            try {
-                org.apache.http.HttpResponse response = httpClient.execute(request);
-                monoSink.success(new HttpResponseWrapper(httpRequest, response));
```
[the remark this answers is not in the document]
The-Funk: done

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -111,4 +120,59 @@ public static RetryConfig buildRetryConfig(Predicate<Throwable> predicate) {
...
+    }
+
+    /**
+     * Gets an HTTP response body as an InputStream.
+     * @param response The response object to convert to an InputStream
+     * @return The response body as an InputStream
+     * @throws IOException An exception indicating an IO failure
+     */
+    public static InputStream getResponseAsStream(HttpResponse response) throws IOException {
+        InputStream contentStream = response.getBodyAsBinaryData().toStream();
```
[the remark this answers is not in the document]
The-Funk: Done. But we should think about how we can remove or replace this. The EofSensorInputStream is another Apache specific concept and comes from the Apache HttpClient 4.

ingest/src/test/java/com/microsoft/azure/kusto/ingest/E2ETest.java
The-Funk: fixed. sometimes I copy paste things that I should really just type and then extra spaces get in there.

ohadbitt (approved):
approving but wait for Asaf

AsafMah (commented):
Approved, however I'll open a new issue to restore the timeout behavior.

AsafMah (approved):

AsafMah (approved):

Issue #377, opened 2024-07-31
Add support for using DefaultAzureCredential 


Soggywaters:
**Is your feature request related to a problem? Please describe.**
Alot of Azure SDKs in Java (but also other supported languages) allow the use of DefaultAzureCredential, which is very helpful when doing local development but also using managed identities on Azure. 

The Kusto auth looks to be setup [very similar](https://github.com/Azure/azure-kusto-java/tree/master/data/src/main/java/com/microsoft/azure/kusto/data/auth). 

**Describe the solution you'd like**
Instead of using the auth providers in the Kusto lib, use the Azure DefaultAzureCredentials utility method from the Azure identity library. 

**Describe alternatives you've considered**
A bunch of if-else/switch-case statements.

AsafMah:
This is planned for the next version.

For now, you can use `CallbackTokenProvider` to wrap the functionality

PR #392, opened 2024-10-28
allows users to provide their own TokenCredential auth from azure-identity

The-Funk:
This would allow users to use azure-identity directly in order to authenticate their kusto client instance.

Putting in draft for consideration.


The-Funk:
This would address #377 by allowing the client dev to use azure-identity directly when creating a Kusto Client.

PR #393, opened 2024-10-31
New client missing feats

ohadbitt:
# Removed
builder.maxConnectionsPerRoute as it was not used in new client

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Moved HTTP request building logic into a builder class.
 - [BREAKING] Redirects are disabled by default. Use ClientRequestProperties "client_max_redirect_count" option 
     to enable. Default changed to 0.
-- Removed  maxConnectionsPerRoute as it was not easily provided by azure-core.
+- [BREAKING] Removed maxConnectionsPerRoute as it was not easily provided by azure-core.
 
 ## [5.2.0] - 2024-08-27
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Moved HTTP request building logic into a builder class.
 - [BREAKING] Redirects are disabled by default. Use ClientRequestProperties "client_max_redirect_count" option 
     to enable. Default changed to 0.
+- Removed  maxConnectionsPerRoute as it was not easily provided by azure-core.
 
 ## [5.2.0] - 2024-08-27
 ### Fixed
--- a/data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
@@ -39,7 +39,7 @@ public abstract class BaseClient implements Client, StreamingClient {
         // Execute and get the response
         try (HttpResponse response = httpClient.sendSync(request, getContextTimeout(timeoutMs))) {
             return processResponseBody(response);
-        } catch (DataServiceException e){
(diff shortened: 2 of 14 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -142,8 +153,13 @@ public static DataServiceException createExceptionFromResponse(String url, HttpR
                 isPermanent);
     }
 
+    private Context getContextTimeout(long timeoutMs) {
+        int requestTimeout = timeoutMs > Integer.MAX_VALUE ? Integer.MAX_VALUE : Math.toIntExact(timeoutMs) + CLIENT_SERVER_DELTA_IN_MILLISECS;
```
AsafMah, line 157: Maybe we should have a lower hard cap than MAX_VALUE, I think in the server side it's 1 hour

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -142,8 +153,13 @@ public static DataServiceException createExceptionFromResponse(String url, HttpR
...
+        return Context.NONE.addData("azure-response-timeout", Duration.ofMillis(requestTimeout));
```
AsafMah, line 158: Is this documented somewhere? we might want to put a comment with the link for future reference

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -15,12 +15,16 @@
...
+import java.io.UncheckedIOException;
 import java.lang.invoke.MethodHandles;
+import java.time.Duration;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 
 public abstract class BaseClient implements Client, StreamingClient {
 
     private static final int MAX_REDIRECT_COUNT = 1;
+    private static final int CLIENT_SERVER_DELTA_IN_MILLISECS = (int) TimeUnit.SECONDS.toMillis(30);
```
AsafMah, line 27: Maybe we should take this to change the name, EXTRA_TIMEOUT_FOR_CLIENT or something like that

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/ExceptionUtils.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.net.URL;
+
+public class ExceptionUtils {
+    public static DataServiceException createExceptionOnPost(Exception e, URL url, String kind) {
+        boolean permanent = false;
+        String prefix = "";
+        if (e instanceof IOException) {
```
AsafMah, line 13: I think if you already change the exception message, you might as well make a proper subclass for IODataServiceException , so the users could filter on them too

AsafMah (changes requested):

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Moved HTTP request building logic into a builder class.
 - [BREAKING] Redirects are disabled by default. Use ClientRequestProperties "client_max_redirect_count" option 
     to enable. Default changed to 0.
+- Removed  maxConnectionsPerRoute as it was not easily provided by azure-core.
```
AsafMah, line 17: Add Breaking tag? 
Also you have a double space

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
ohadbitt: not exactly but ill ref the changelog

AsafMah (approved):

AsafMah (approved):

PR #395, opened 2024-11-06
Remove unused timeout from HttpClientProperties and ResourceManager

yihezkel:
Returning old computer as I'm upgrading to a new computer, and found these minor improvements lying around locally.


ohadbitt:
the timeout is already fixed in my PR
the slf4j- i dont see the need - do you think its neede ?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ResourceManager.java
```diff
@@ -74,7 +74,7 @@ public ResourceManager(Client client, long defaultRefreshTime, long refreshTimeO
         this.client = client;
         // Using ctor with client so that the dependency is used
         this.httpClient = httpClient == null
-                ? HttpClientFactory.create(HttpClientProperties.builder().timeout(Duration.ofMinutes(UPLOAD_TIMEOUT_MINUTES)).build())
+                ? HttpClientFactory.create(HttpClientProperties.builder().timeout(UPLOAD_TIMEOUT).build())
```
ohadbitt, line 77: as talked - remove the timeout ..

ohadbitt (approved):

Issue #398, opened 2024-11-21
Expected date for the next release


naracorreia01:
Hello,

Context:

We have two docker container: one is running our service image and one is running the Kusto Emulator image. 
Our service wants to **ingest** messages to Kusto Emulator. For that, we created an IngestClient to connect to Kusto Emulator. Since IngestClient is inside a container, we use the hostname of Kusto Emulator container ("kusto-emulator")  instead of localhost to make our service container communicate with it. 
The **connection between docker containers** uses **http://**.

**Current Problem:** since the communication is between two docker containers the hostname is neither localhost nor https://, as referred previously, it fails in the following step as a unsecure connection:

```
HttpPostUtils.class

private static @NotNull URI parseUriFromUrlString(String url) throws DataClientException {
        try {
            URL cleanUrl = new URL(url);
    --->     if (!"https".equalsIgnoreCase(cleanUrl.getProtocol()) && !url.toLowerCase().startsWith("http://localhost")) {
                throw new DataClientException(url, "Cannot forward security token to a remote service over insecure channel (http://)");
            } else {
                return new URI(cleanUrl.getProtocol(), cleanUrl.getUserInfo(), cleanUrl.getHost(), cleanUrl.getPort(), cleanUrl.getPath(), cleanUrl.getQuery(), cleanUrl.getRef());
            }
        } catch (MalformedURLException | URISyntaxException var2) {
            Exception e = var2;
            throw new DataClientException(url, "Error parsing target URL in post request:" + ((Exception)e).getMessage(), e);
        }
    }


public static String post(CloseableHttpClient httpClient, String urlStr, AbstractHttpEntity requestEntity, long timeoutMs, Map<String, String> headers) throws DataServiceException, DataClientException {
  --->    URI url = parseUriFromUrlString(urlStr);
        [...]
    }
```

```
ClientImpl.class

    private KustoOperationResult executeStreamingIngestImpl(String clusterEndpoint, InputStream stream, String blobUrl, ClientRequestProperties properties, boolean leaveOpen) throws DataServiceException, DataClientException {

        [...]
        this.validateEndpoint();
                AbstractHttpEntity entity = isStreamSource ? new InputStreamEntity(new UncloseableStream(stream)) : new StringEntity((new IngestionSourceStorage(blobUrl)).toString(), ContentType.APPLICATION_JSON);
                String response = (String)MonitoredActivity.invoke(() -> {
[...]

PR #399, opened 2024-11-24
Add no-auth option

AsafMah:
Solves #398


data/src/main/java/com/microsoft/azure/kusto/data/http/HttpRequestBuilder.java
```diff
@@ -107,13 +107,25 @@ public HttpRequestBuilder withURL(String url) throws DataClientException {
...
     }
 
-    public HttpRequest build() {
+    public HttpRequest build() throws DataClientException {
+        // If has authorization header, ensure it is not sent over insecure channel
+        boolean hasAuth = request.getHeaders().stream().anyMatch(h -> h.getName().equalsIgnoreCase(HttpHeaderName.AUTHORIZATION.toString()));
+        if (hasAuth) {
+            URL url = request.getUrl();
+            boolean isHttp = url.getProtocol().equalsIgnoreCase("http");
+            boolean isLocalhost = url.getHost().toLowerCase().startsWith(CloudInfo.LOCALHOST);
```
yogilad, line 116: this is a bit risky.
localhostmyass.com is a valid target 😅
You'd want to regex match the host against `^localhost(\:[0-9]+)?$`

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpRequestBuilder.java
```diff
@@ -107,13 +107,25 @@ public HttpRequestBuilder withURL(String url) throws DataClientException {
...
+
+            if (isHttp && !isLocalhost) {
+                throw new DataClientException(url.toString(), "Cannot forward security token to a remote service over insecure " +
```
yogilad, line 119: insecure means someone who is not confident in himself.
Unsecure, not secure, unenceypted, etc.

yogilad (commented):
Looks good.
Left some comments.
You can commit after fixing

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpRequestBuilder.java
AsafMah: good point (though this is the logic we already have).

I think the host method returns it without the port, so I'll just check equality

data/src/main/java/com/microsoft/azure/kusto/data/http/HttpRequestBuilder.java
```diff
@@ -107,13 +107,24 @@ public HttpRequestBuilder withURL(String url) throws DataClientException {
...
-    public HttpRequest build() {
+    public HttpRequest build() throws DataClientException {
+        // If has authorization header, ensure it is not sent over insecure channel
+        boolean hasAuth = request.getHeaders().stream().anyMatch(h -> h.getName().equalsIgnoreCase(HttpHeaderName.AUTHORIZATION.toString()));
+        if (hasAuth) {
+            URL url = request.getUrl();
+            boolean isHttp = url.getProtocol().equalsIgnoreCase("http");
+            boolean isLocalhost = url.getHost().equalsIgnoreCase(CloudInfo.LOCALHOST);
+
+            if (isHttp && !isLocalhost) {
```
ohadbitt, line 118: maybe log the case of isHttp + isLocalHost 
so users would be aware

PR #402, opened 2024-11-28
Data library async implementation

georgebanasios:
### Added
- The SDK now provides Reactor Core-based asynchronous APIs for all query, management, streaming query/ingestion (StreamingClient) endpoints, enabling non-blocking operations.

### Changed
- [BREAKING] All synchronous query/management, streaming query/ingestion (StreamingClient) APIs now delegate to their asynchronous counterparts internally and block for results.


AsafMah:
Will you let us know when it's ready to review?

georgebanasios:
@AsafMah 
Of course! This is my priority right now, and I’m fully focused on it.
I’ll let you know as soon as it’s ready for review.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -1,34 +1,38 @@
...
 import java.net.URISyntaxException;
 import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class CloudInfo implements TraceableAttributes, Serializable {
-    private static final Map<String, CloudInfo> cache = new HashMap<>();
+    private static final Map<String, CloudInfo> cache = new ConcurrentHashMap<>(); //TODO: is this correct?
```
The-Funk, line 35: Since fetches from the cloud info endpoint can now happen asynchronously and this is static or global state, it's possible multiple put/get operations could occur simultaneously. I don't think that's likely but using ConcurrentHashMap here costs next to nothing so I would call this correct.

georgebanasios:
Hello @AsafMah !

I have a question about the approach for managing shared resources, particularly regarding concurrency and synchronization.
Given that the current design uses locks to manage concurrent access to shared state (e.g., resource sets, identity token), I’m curious about the reasoning behind this decision when fully moving to async. 

Were you guys considering any specific strategies or patterns to handle concurrency in an async context? For instance, as async workflows will likely involve multiple tasks accessing and updating shared resources concurrently, I’m wondering if the locking approach will be adapted or if there's an intention to remove the locks in favor of another solution, (perhaps using atomic non-blocking operations, reactive patterns, etc.) in order to ensure that async tasks can safely read/write shared resources without blocking each other.

AsafMah:
@georgebanasios we encountered the same issues, and currently are still looking for a good solution.

Eliminating locks whenever possible is usually an advantage either way.

As far as I could find, there is no standard async-supporting lock, you might be able to simulate it with a task that periodically checks it.

georgebanasios:
> @
> 
> > @georgebanasios we encountered the same issues, and currently are still looking for a good solution.
> > Eliminating locks whenever possible is usually an advantage either way.
> > As far as I could find, there is no standard async-supporting lock, you might be able to simulate it with a task that periodically checks it.
> 
> @georgebanasios: Thank you for your contribution, we too thought about the solution you did here, but we think its better to make sure the fetch is only called once.
> 
> We conversed about it, i found this article interesting : https://nitschinger.at/Reactive-Barriers-with-Reactor Although the usage of MonoProcessor is deprecated, project reactor suggests to use instead Sinks.One. Here's a copilot sample Sinks.One sink = Sinks.one(); Mono mono = sink.asMono(); mono.subscribe(value -> System.out.println("Received: " + value)); sink.emitValue("Hello, Reactor!");

@ohadbitt Thanks for the suggestion!

Just to clarify, are you referring specifically to the fetch operations related to resource management?

ohadbitt:
> > @
> > > @georgebanasios we encountered the same issues, and currently are still looking for a good solution.
> > > Eliminating locks whenever possible is usually an advantage either way.
> > > As far as I could find, there is no standard async-supporting lock, you might be able to simulate it with a task that periodically checks it.
> > 
> > 
> > @georgebanasios: Thank you for your contribution, we too thought about the solution you did here, but we think its better to make sure the fetch is only called once.
> > We conversed about it, i found this article interesting : https://nitschinger.at/Reactive-Barriers-with-Reactor Although the usage of MonoProcessor is deprecated, project reactor suggests to use instead Sinks.One. Here's a copilot sample Sinks.One sink = Sinks.one(); Mono mono = sink.asMono(); mono.subscribe(value -> System.out.println("Received: " + value)); sink.emitValue("Hello, Reactor!");
> 
> @ohadbitt Thanks for the suggestion!
> 
> Just to clarify, are you referring specifically to the fetch operations related to resource management (i.e.., ingestion resources, identity token)?

Yes I'm referring to any result we used to take once (or once every .. )

georgebanasios:
Hello @AsafMah @ohadbitt 
The PR is now ready for your review whenever you have the time, thank you!

ohadbitt:
It seems like you tried to use sinks and decided to go back to synchronization - can you share what happened ? it does seem to work fine now btw..

georgebanasios:
> It seems like you tried to use sinks and decided to go back to synchronization - can you share what happened ? it does seem to work fine now btw..

I tried some queries with multiple threads, and I noticed that sinks (with multiple types) emit the signals serially, and that was causing some exceptions (Spec. Rule 1.3) when interacting with multiple threads. I handled that by using the busy loop option of the EmitFailureHandler which basically retries failed emissions for a specified duration, but I read that causes cpu issues on high concurrent environments.

Since the result is cached, it seemed using sinks in this scenario a bit redundant, and I went with computeIfAbsent which solves these problems.

ohadbitt:
@georgebanasios 
Sorry for taking time, its a little hard to read code and we are not experts with this reactive java - which is why ill make first this comment.
The code seems too much nested - too many Monos are created - lets say as a general rule we only want methods to return Mono if they have an async call or throws an exception which is not runtime exception (i actually now understand why all the new exceptions from the Storage libraries are Runtime exceptions... and maybe we should consider making ours as well - @AsafMah we can talk about that F2F).
So for example CloudInfo.sendRequest - could be just a simple call to the MonitoredActivity.wrap method. We can avoid all these Mono.just.flatMap and  Mono.fromCallable.

BTW as a side note (You probably ended up with a better solution.. but im just saying..) - I think sinks didnt work for you as they weren't static - we could have a static Sinks.One that publishes the cloudInfo. 

Also i think we could make better progress if we had an open Teams channel with you - do you want to send a mail address and do that ?  

Thanks!

georgebanasios:
@ohadbitt 
Yes you are right, about the nesting. The main issue I faced, just like you mentioned, was that multiple external methods in a single method throw checked exceptions, which were later caught and transformed. This is why I ended up using fromCallable and flatMap.
I’ll work on adjusting the code and provide you with an update soon!

I agree that communication would be more efficient via a Teams channel. You can email me at: georgeban70@gmail.com

ohadbitt:
@georgebanasios 
We decided on making our exceptions of type AzureException (extending RuntimeException)
Do you want to make this change here ? or should we make this PR ourselves before ?

georgebanasios:
> @georgebanasios We decided on making our exceptions of type AzureException (extending RuntimeException) Do you want to make this change here ? or should we make this PR ourselves before ?

Sure, I'll make the change and adjust the code accordingly!

ohadbitt (commented):
will take a look next week

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -124,180 +167,195 @@ private Map<String, String> updateAndGetExecuteTracingAttributes(String database
...
-            return new KustoOperationResult(response, "v1");
-        } catch (KustoServiceQueryError e) {
-            throw new DataClientException(clusterEndpoint, "Error converting json response to KustoOperationResult:" + e.getMessage(), e);
-        } catch (IOException e) {
-            throw new DataClientException(clusterUrl, e.getMessage(), e);
-        }
+        String contentEncoding = isStreamSource ? "gzip" : null;
+        String contentType = isStreamSource ? "application/octet-stream" : "application/json";
+
+        return Mono.fromCallable(() -> determineTimeout(properties, CommandType.STREAMING_INGEST, clusterUrl))
```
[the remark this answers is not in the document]
georgebanasios: Yes, I have removed almost all fromCallable.flatmaps. I think this commit you're viewing is old.

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -38,7 +38,7 @@
 
 public abstract class BaseClient implements Client, StreamingClient {
 
-    // TODO - this is never used?
+    // TODO - this is never used? should a max redirect count be included?
     private static final int MAX_REDIRECT_COUNT = 1;
```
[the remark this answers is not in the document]
georgebanasios: The "//TODO this is never used?" was a comment from Asaf actually. I don't see it being used, I'll check again though!

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -37,79 +50,84 @@ public BaseClient(HttpClient httpClient) {
...
-            return processResponseBody(response);
-        } catch (DataServiceException e) {
-            throw e;
-        } catch (Exception e) {
-            throw ExceptionUtils.createExceptionOnPost(e, request.getUrl(), "sync");
-        }
+    protected Mono<String> postAsync(HttpRequest request, long timeoutMs) {
+        return httpClient.send(request, getContextTimeout(timeoutMs))
+                .flatMap(response -> Utils.getResponseBody(response)
+                        .flatMap(responseBody -> {
```
ohadbitt, line 56: So my reasoning for moving to unchecked exceptions was that now we can use the light weight map instead of flat map for non-async calls - here is a good place for that - instead of using monos we can just throw exception

georgebanasios: Totally missed it. Fixed!

data/src/main/java/com/microsoft/azure/kusto/data/BaseClient.java
```diff
@@ -37,79 +50,84 @@ public BaseClient(HttpClient httpClient) {
...
+            int currentRedirectCounter, int maxRedirectCount) {
+        ResponseState state = new ResponseState();
+        return httpClient.send(request, getContextTimeout(timeoutMs))
+                .flatMap(httpResponse -> {
+                    state.setHttpResponse(httpResponse);
+                    int responseStatusCode = httpResponse.getStatusCode();
+                    if (responseStatusCode == HttpStatus.OK) {
+                        state.setReturnInputStream(true);
+                        return httpResponse.getBodyAsInputStream() // TODO: since we want to just close on EOF should we implement EofSensorInputStream
+                                                                   // ourselves and remove the remaining Apache dependency or not?
```
The-Funk, line 85: If this is the last thing keeping the Apache dependency, I think we should implement it and remove the dependency.

data/src/main/java/com/microsoft/azure/kusto/data/Client.java
```diff
@@ -3,26 +3,35 @@
 
 package com.microsoft.azure.kusto.data;
 
-import com.microsoft.azure.kusto.data.exceptions.DataClientException;
-import com.microsoft.azure.kusto.data.exceptions.DataServiceException;
+import reactor.core.publisher.Mono;
 
 public interface Client {
 
-    KustoOperationResult executeQuery(String command) throws DataServiceException, DataClientException;
```
The-Funk, line 11: I'm so excited to see less thrown checked exceptions.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -146,33 +156,98 @@ public static boolean isGzipResponse(HttpResponse response) {
...
+
+                        return new String(output.toByteArray(), StandardCharsets.UTF_8);
+                    } catch (IOException e) {
+                        throw Exceptions.propagate(e);
+                    }
+                }).subscribeOn(Schedulers.boundedElastic());
+    }
+
+    // Alternative an approach to streaming decompression. I had some OOM errors locally though on KustoOperationResult.createFromV2Response
+    public static Mono<String> processGzipBody1(Flux<ByteBuffer> gzipBody) {
```
AsafMah, line 197: We like this approach, it keeps it async and performs well.

We do have some suggestions:

1. You can drop the flatMap and do the work in the reduce, it saves you a step.
2. From what I see, writing to the channel already decreases the refcount, so we don't need to do it ourselves later.
3. Again, from what I understand, writing one message to the zlib codec will return one message. I don't know if it's written to be guranteed somewhere, but it does in all of our testing.

For 2+3, maybe we'll throw for now in case they aren't true, and later we can remove it and assume ok.

So we have something like this:
```java
public static Mono<String> processGzipBody(Flux<ByteBuffer> gzipBody) {
        // To ensure efficiency in terms of performance and memory allocation, a streaming
        // chunked decompression approach is utilized using Netty's EmbeddedChannel. This allows the decompression
        // to occur in chunks, making it more memory-efficient for large payloads, as it prevents the entire
        // compressed stream from being loaded into memory at once (which is required by GZIPInputStream for decompression).
        // Flux.create() emits decompressed data as it becomes available, while ensuring backpressure.
        // Decompression is offloaded to `boundedElastic()`, which prevents thread blocking and significantly improves performance.

        EmbeddedChannel channel = new EmbeddedChannel(ZlibCodecFactory.newZlibDecoder(ZlibWrapper.GZIP));

        return gzipBody
                .subscribeOn(Schedulers.boundedElastic())
                .reduce(new StringBuilder(), (stringBuilder, byteBuffer) -> {
                    channel.writeInbound(Unpooled.wrappedBuffer(byteBuffer)); // Write chunk to channel for decompression
     
                    ByteBuf decompressedByteBuf = channel.readInbound();
                    if (decompressedByteBuf == null) {
                        return stringBuilder;
                    }

                    String string = decompressedByteBuf.toString(StandardCharsets.UTF_8);
                    decompressedByteBuf.release();

                    if (!channel.inboundMessages().isEmpty()) {
                        throw new IllegalStateException("Expected exactly one message in the channel");
                    }

                    stringBuilder.append(string);

                    return stringBuilder;
                }) 
                .map(StringBuilder::toString)
                .doFinally(ignore -> channel.finish());
    }
```

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -146,33 +156,98 @@ public static boolean isGzipResponse(HttpResponse response) {
...
+        // To ensure efficiency in terms of performance and memory allocation, a streaming
+        // chunked decompression approach is utilized using Netty's EmbeddedChannel. This allows the decompression
+        // to occur in chunks, making it more memory-efficient for large payloads, as it prevents the entire
+        // compressed stream from being loaded into memory at once (which is required by GZIPInputStream for decompression).
+        // Flux.create() emits decompressed data as it becomes available, while ensuring backpressure.
+        // Decompression is offloaded to `boundedElastic()`, which prevents thread blocking and significantly improves performance.
```
AsafMah, line 203: Did boundedElastic measurably improve perf? do we want to also have it when you do json parsing for example?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -146,33 +156,98 @@ public static boolean isGzipResponse(HttpResponse response) {
...
+                            if (decompressedByteBuf.refCnt() > 0) {
+                                decompressedByteBuf.release();
+                            }
+                        }
+                        sink.complete();
+                    }).doOnError(ignore -> channel.finish());
+                }).subscribeOn(Schedulers.boundedElastic())
+                .reduce(new StringBuilder(), StringBuilder::append) // TODO: this will cause memory problems right?
+                .map(StringBuilder::toString)
+                .doFinally(ignore -> channel.finish());
```
AsafMah, line 229: Do we want instead to do finishAndReleaseAll()

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -146,33 +156,98 @@ public static boolean isGzipResponse(HttpResponse response) {
...
+                    return Flux.<String>create(sink -> {
+                        ByteBuf decompressedByteBuf;
+                        while ((decompressedByteBuf = channel.readInbound()) != null) {
+                            sink.next(decompressedByteBuf.toString(StandardCharsets.UTF_8));
+                            if (decompressedByteBuf.refCnt() > 0) {
+                                decompressedByteBuf.release();
+                            }
+                        }
+                        sink.complete();
+                    }).doOnError(ignore -> channel.finish());
```
AsafMah, line 225: If this triggers, channel.finish() gets called twice

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -71,124 +82,94 @@ public CloudInfo(boolean loginMfaRequired, String loginEndpoint, String kustoCli
...
-                    String content;
-                    if (Utils.isGzipResponse(response)) {
-                        content = Utils.gzipedInputToString(new ByteArrayInputStream(bodyAsBinaryData));
-                    } else {
-                        content = new String(bodyAsBinaryData);
-                    }
-
-                    if (content.isEmpty() || content.equals("{}")) {
-                        throw new DataServiceException(clusterUrl, "Error in metadata endpoint, received no data", true);
+        HttpRequest request = new HttpRequest(HttpMethod.GET, clusterUrls.getT2());
```
AsafMah, line 114: You can lose the Tuple, and just pass `UriUtils.setPathForUri(clusterUrl, METADATA_ENDPOINT)` here no? also saves on the computation if it's already in cache

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
```diff
@@ -54,9 +57,8 @@ public static void setOverridePolicy(Predicate<String> matcher) {
      *
      * @param uri           - Kusto endpoint
      * @param loginEndpoint The login endpoint to check against.
-     * @throws KustoClientInvalidConnectionStringException - Endpoint is not a trusted Kusto endpoint
      */
-    public static void validateTrustedEndpoint(String uri, String loginEndpoint) throws KustoClientInvalidConnectionStringException {
```
AsafMah, line 59: Here, and in all public APIs, should we keep the `throws`?
We don't need it, but it signals intent to the user.

If not, maybe just keeping the doc.

data/src/main/java/com/microsoft/azure/kusto/data/auth/HttpClientWrapper.java
```diff
@@ -21,7 +21,7 @@ public HttpClientWrapper(HttpClient httpClient) {
 
     // Implementation of the synchronous HttpClient
     @Override
-    public IHttpResponse send(com.microsoft.aad.msal4j.HttpRequest httpRequest) {
+    public IHttpResponse send(com.microsoft.aad.msal4j.HttpRequest httpRequest) { // TODO: does this need to be changed?
```
AsafMah, line 24: I think sendsync is ok here, since it's only for when it's through auth code, and even then only though stuff we still have to do through msal (Certificate stuff). Hopefully they update azure-identity with these features and we'll get rid of this and msal entirely.

data/src/main/java/com/microsoft/azure/kusto/data/auth/endpoints/KustoTrustedEndpoints.java
georgebanasios: Yes I was thinking the same, especially after checking the azure core sdk how it handles cases like that. I'll just keep it in the docs.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
georgebanasios: I'll go with finishAndReleaseAll(), just in case an error occurs and maybe a resource is not released, which I don't see it with your suggestion earlier.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -71,124 +82,94 @@ public CloudInfo(boolean loginMfaRequired, String loginEndpoint, String kustoCli
...
-                        // Fixme: Missing reason phrase to add to exception. Potentially want to use an enum.
-                        errorFromResponse = "";
+                });
+    }
+
+    private static Mono<CloudInfo> getCloudInfo(HttpResponse response, String clusterUrl) {
+        int statusCode = response.getStatusCode();
+        return Utils.getResponseBody(response)
+                .map(content -> {
+                    if (statusCode == HttpStatus.OK) {
```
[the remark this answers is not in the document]
georgebanasios: Proceeded with the suggestion.
Out of curiosity, do you have a specific plugin for that suggestion to pop up?

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
georgebanasios: It did when I was still using GzipInputStream, but I forgot to delete it.
Since we are going with Netty's channel, after testing it I don't see it improving performance.
For json parsing (response -> KustoOperationResult) I'm offloading it to boundedElastic(). ClientImpl#exeuteImplAsync for example.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
georgebanasios: Yes, you are right, removed.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
georgebanasios: As far as performance, did you notice any improvements in comparison with GZIPInputStream? Not sure how representative my tests are, compared to yours.

AsafMah: Mostly for readability, since this function is very dense and I want to reduce the mental load as much as possible.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -146,33 +157,101 @@ public static boolean isGzipResponse(HttpResponse response) {
...
+                        while ((decompressedByteBuf = channel.readInbound()) != null) {
+                            sink.next(decompressedByteBuf.toString(StandardCharsets.UTF_8));
+                            if (decompressedByteBuf.refCnt() > 0) {
+                                decompressedByteBuf.release();
+                            }
+                        }
+
+                        sink.complete();
+                    }).doOnError(throwable -> channel.finish()).cast(String.class);
+                }).subscribeOn(Schedulers.boundedElastic())
```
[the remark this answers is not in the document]
georgebanasios: I think this commit is old. I removed this entirely.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -124,34 +166,34 @@ private Map<String, String> updateAndGetExecuteTracingAttributes(String database
...
         String clusterEndpoint = String.format(kr.getCommandType().getEndpoint(), clusterUrl);
-        return processJsonResult(new JsonResult(response, clusterEndpoint));
+        return executeWithTimeout(kr, ".executeImplAsync")
+                .publishOn(Schedulers.boundedElastic())
+                .map(response -> {
+                    JsonResult jsonResult = new JsonResult(response, clusterEndpoint);
+                    return new KustoOperationResult(jsonResult.getResult(), jsonResult.getEndpoint().endsWith("v2/rest/query") ? "v2" : "v1");
+                })
+                .onErrorMap(KustoServiceQueryError.class, e -> new DataServiceException(clusterEndpoint,
+                        "Error found while parsing json response as KustoOperationResult:" + e, e, e.isPermanent()))
```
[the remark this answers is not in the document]
georgebanasios: Sure, done.

ohadbitt (commented):
still trying to remove most of the flatMaps where its not async call :)
 - it seems to me to be a big deal but maybe im just new to reactive

data/src/main/java/com/microsoft/azure/kusto/data/ExponentialRetry.java
```diff
@@ -62,4 +71,45 @@ public <T> T execute(KustoCheckedFunction<Integer, T, E1, E2> function) throws E
         return null;
     }
 
+    public Retry retry(@Nullable List<Class<? extends Throwable>> retriableErrorClasses) {
```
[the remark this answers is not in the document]
georgebanasios: Yes it will be used on ingest module. I added a javadoc that explains it.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -160,88 +202,83 @@ KustoRequestContext prepareRequest(@NotNull KustoRequest kr) throws DataServiceE
                 .withClientDetails(clientDetails)
                 .build();
 
-        // Build the HTTP request
+        return validateEndpointAsync()
+                .then(authorization
+                        .map(token -> buildKustoRequestContext(kr, clusterEndpoint, tracing, token))
+                        .switchIfEmpty(Mono.just(buildKustoRequestContext(kr, clusterEndpoint, tracing, null))));
```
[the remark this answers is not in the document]
georgebanasios: I proceeded with defer instead. It seems more clear that returning null maybe.

data/src/main/java/com/microsoft/azure/kusto/data/ExponentialRetry.java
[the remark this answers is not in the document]
georgebanasios: Like this?

georgebanasios:
> still trying to remove most of the flatMaps where its not async call :)
> 
> * it seems to me to be a big deal but maybe im just new to reactive

No, you are right. I managed to remove almost all where there is not an async call. The ones left are on BaseClient (processes the response), and on ClientImpl. I'll look into it if any remaining can be removed.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudDependentTokenProviderBase.java
```diff
@@ -30,31 +29,25 @@ public abstract class CloudDependentTokenProviderBase extends TokenProviderBase
...
+            // trace retrieveCloudInfo
+            return MonitoredActivity.wrap(
+                    CloudInfo.retrieveCloudInfoForClusterAsync(clusterUrl, httpClient),
+                    "CloudDependentTokenProviderBase.retrieveCloudInfo", getTracingAttributes())
+                    .doOnNext(cloudInfoResult -> {
+                        this.cloudInfo = cloudInfoResult;
+                        initializeWithCloudInfo(cloudInfoResult);
+                        this.initialized = true;
+                    })
+                    .then();
```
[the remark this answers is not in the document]
georgebanasios: The doOnNext returns the type of the consumer (CloudInfo in this case), so then() discards that source element and returns void.

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -1,34 +1,44 @@
 package com.microsoft.azure.kusto.data.auth;
```
ohadbitt, line 1: Im not sure if its because of some of the latests changes but now i see we call fetch multiple times for the same endpoint - as i didnt get what solved it in the first place - please do check that .

georgebanasios: Would it be too much trouble to provide an example of how you're testing that? I might be missing something. 
Via Teams if you prefer as well. 
Currently, I’m observing the same behavior as before: the fetch occurs only once for the same endpoint. Either from E2E or by manually using multiple threads.

ohadbitt: ConnectionStringBuilder c = createConnection(ENG_CONN_STR);
        queryClient = ClientFactory.createClient(c);
        streamingClient = ClientFactory.createStreamingClient(c);

        queryClient.executeMgmt(DB, ".show version");
        queryClient.executeMgmt(DB, ".show version");
        queryClient.executeMgmt(DB, ".show version");
        InputStream is = streamingClient.executeStreamingQuery("sdktestdb", "javaLoadTest |take 100", null); 
        .
        .
        I get to getCloudInfo() 4 times 
        the first execute gets there twice and the executeStreamingQuery gets twice as well

ohadbitt: started a teams chat with you if you want me to show you the run

data/src/main/java/com/microsoft/azure/kusto/data/ExponentialRetry.java
```diff
@@ -62,4 +69,37 @@ public <T> T execute(KustoCheckedFunction<Integer, T, E1, E2> function) throws E
...
+                .filter(throwable -> shouldRetry(throwable, retriableErrorClasses))
+                .doAfterRetry(retrySignal -> {
+                    long currentAttempt = retrySignal.totalRetries() + 1;
+                    log.info("Attempt {} failed.", currentAttempt);
+                })
+                .onRetryExhaustedThrow((spec, signal) -> signal.failure());
+    }
+
+    private static boolean shouldRetry(Throwable failure, List<Class<? extends Throwable>> retriableErrorClasses) {
+        if (failure instanceof DataServiceException && ((DataServiceException) failure).isPermanent()) {
```
ohadbitt, line 92: If DataServiceException.isPermanent == false 
we want to return true - this works now well. 
but if retriableErrorClasses != null -> then you will return false

data/src/main/java/com/microsoft/azure/kusto/data/auth/CloudInfo.java
```diff
@@ -71,124 +80,101 @@ public CloudInfo(boolean loginMfaRequired, String loginEndpoint, String kustoCli
...
-            });
+        // We ensure that if multiple threads request the cloud info for the same cluster url, only one http call will be made
+        // for all corresponding threads
+        try {
+            String clusterEndpoint = UriUtils.setPathForUri(clusterUrl, "");
+            return CACHE.computeIfAbsent(clusterEndpoint, key ->
+
+            // If an error occurs, each time the retryWhen subscribes to fetchCloudInfoAsync create a new instance
+            // instead of using the same fetchCloudInfoAsync Mono for all retries
+            Mono.defer(() -> fetchCloudInfoAsync(clusterEndpoint, givenHttpClient))
```
ohadbitt, line 101: Big like on the doOnError :)

data/src/main/java/com/microsoft/azure/kusto/data/ExponentialRetry.java
georgebanasios: Fixed

ohadbitt (approved):

ohadbitt:
Approved but please do add the "throws " to the async API as well

PR #404, opened 2024-12-22
Auth metadata endpoint is always rooted.

alonadam:
### Fixed
Auth metadata endpoint is always relative to the root of the cluster


georgebanasios:
Suggestion, since there is already a method to set the path of a url on UriUtils, and URIBuilders were removed previously:

`public static String replacePath(String uri, String pathToReplaceWith) throws URISyntaxException {
        return setPathForUri(uri, pathToReplaceWith);
    }`

ohadbitt (commented):
please add test with the case that failed before and now passes

AsafMah (approved):

PR #406, opened 2025-01-06
Fixes exception wrapping

georgebanasios:
Fixes exception wrapping.


data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -175,7 +175,13 @@ private Mono<KustoOperationResult> executeImplAsync(KustoRequest kr) {
                     return new KustoOperationResult(jsonResult.getResult(), jsonResult.getEndpoint().endsWith("v2/rest/query") ? "v2" : "v1");
                 })
                 .onErrorMap(KustoServiceQueryError.class, e -> new DataServiceException(clusterEndpoint, e.getMessage(), e, e.isPermanent()))
-                .onErrorMap(Exception.class, e -> new DataClientException(clusterEndpoint, ExceptionUtils.getMessageEx(e), e));
+                .onErrorMap(Exception.class, e -> {
```
ohadbitt, line 178: It seems like the old implementaion was simply catching here IOException

ohadbitt (approved):

PR #409, opened 2025-01-09
kcsb refactor

AsafMah:
Aligns KCSB parsing with the new method - having a json file that defines keywords and aliases.

Also adds:

- ToString() methods for KCSB to align with other sdks
- Keyword parsing that is case and space insensitive
- Added keyword for certificate blob, but until we figure it out it just throws for now

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
@@ -5,7 +5,6 @@ package com.microsoft.azure.kusto.data.auth;
 
 import com.azure.core.credential.TokenCredential;
 import com.microsoft.azure.kusto.data.ClientDetails;
-import org.apache.commons.lang3.NotImplementedException;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.jetbrains.annotations.NotNull;
@@ -24,7 +23,6 @@ public class ConnectionStringBuilder {
     private String applicationClientId;
     private String applicationKey;
     // Public certificate
-    private byte[] certificateBytes;
     private String initialCatalog;
     private boolean sendX509;
     private boolean aadFederatedSecurity;
@@ -76,9 +74,7 @@ public class ConnectionStringBuilder {
         this.userNameForTracing = null;
         this.appendedClientVersionForTracing = null;
         this.applicationNameForTracing = null;
-        this.certificateBytes = null;
         this.sendX509 = false;
-        this.certificateBytes = null;
         this.initialCatalog = null;
     }
 
@@ -107,9 +103,6 @@ public class ConnectionStringBuilder {
             case AUTHORITY_ID:
                 this.aadAuthorityId = value;
                 break;
-            case APPLICATION_CERTIFICATE_BLOB:
-                this.certificateBytes = Base64.getDecoder().decode(value);
-                throw new NotImplementedException("Application certificate blob is not supported yet.");
             case APPLICATION_CERTIFICATE_X5C:
                 this.sendX509 = Boolean.parseBoolean(value);
                 break;
@@ -206,7 +199,6 @@ public class ConnectionStringBuilder {
         this.userNameForTracing = other.userNameForTracing;
         this.appendedClientVersionForTracing = other.appendedClientVersionForTracing;
         this.applicationNameForTracing = other.applicationNameForTracing;
-        this.certificateBytes = other.certificateBytes;
         this.sendX509 = other.sendX509;
         this.initialCatalog = other.initialCatalog;
         this.aadFederatedSecurity = other.aadFederatedSecurity;
@@ -216,7 +208,7 @@ public class ConnectionStringBuilder {
     /**
      * Creates a ConnectionStringBuilder from a connection string. For more information please look <a href="https://docs.microsoft.com/azure/data-explorer/kusto/api/connection-strings/kusto">here</a>.
      *
-     * @param connectionString The connection string should be of the format: <p>https://clusterName.location.kusto.windows.net;AAD User ID="user@microsoft.com";Password=P@ssWord</p>
+     * @param connectionString The connection string should be of the format: <p>https://clusterName.location.kusto.windows.net;Fed=true;Application Client Id=****;Application Key=****</p>
      * @throws IllegalArgumentException If the connection string is invalid.
      */
     public ConnectionStringBuilder(String connectionString) {
--- a/data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
@@ -53,7 +53,6 @@ public class ConnectionStringBuilder {
     private String appendedClientVersionForTracing;
     private String applicationNameForTracing;
 
-
     private ConnectionStringBuilder() {
         this.aadFederatedSecurity = false;
         this.clusterUrl = null;
@@ -165,8 +164,8 @@ public class ConnectionStringBuilder {
         for (int i = 0; i < entries.size(); i++) {
             Pair<KnownKeywords, String> entry = entries.get(i);
             sb.append(entry.getLeft().getCanonicalName())
-                .append("=")
-                .append((!showSecrets && entry.getLeft().isSecret()) ? SECRET_REPLACEMENT : entry.getRight());
+                    .append("=")
+                    .append((!showSecrets && entry.getLeft().isSecret()) ? SECRET_REPLACEMENT : entry.getRight());
 
             if (i < entries.size() - 1) {
                 sb.append(";");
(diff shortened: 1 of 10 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -97,52 +70,114 @@ private ConnectionStringBuilder() {
...
                 this.accessToken = value;
                 break;
             default:
                 throw new IllegalArgumentException("Error: unsupported key " + rawKey + " in connection string");
         }
     }
 
+    public String toString(boolean showSecrets) {
+        ArrayList<Pair<KnownKeywords, String>> entries = new ArrayList<>();
+        if (!StringUtils.isBlank(clusterUrl)) {
```
yogilad, line 130: nit: new line before this one

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -97,52 +70,114 @@ private ConnectionStringBuilder() {
...
+        if (!StringUtils.isBlank(userNameForTracing)) {
+            entries.add(Pair.of(KnownKeywords.USER_NAME_FOR_TRACING, userNameForTracing));
+        }
+
+        StringBuilder sb = new StringBuilder();
+
+        for (int i = 0; i < entries.size(); i++) {
+            Pair<KnownKeywords, String> entry = entries.get(i);
+            sb.append(entry.getLeft().getCanonicalName()).append("=").append(
+                    (!showSecrets && entry.getLeft().isSecret()) ? SECRET_REPLACEMENT : entry.getRight());
```
yogilad, line 167:
```suggestion
            sb.append(entry.getLeft().getCanonicalName())
                .append("=")
                .append((!showSecrets && entry.getLeft().isSecret()) ? SECRET_REPLACEMENT : entry.getRight());
                
```

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -346,6 +385,7 @@ public static ConnectionStringBuilder createWithAadApplicationCredentials(String
         }
 
         ConnectionStringBuilder csb = new ConnectionStringBuilder();
+        csb.aadFederatedSecurity = true;
```
yogilad, line 388: Is there a no auth option in Java?

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -97,52 +70,114 @@ private ConnectionStringBuilder() {
...
         this.appendedClientVersionForTracing = null;
         this.applicationNameForTracing = null;
+        this.sendX509 = false;
+        this.initialCatalog = null;
     }
 
     private void assignValue(String rawKey, String value) {
-        rawKey = rawKey.trim().toLowerCase();
-        ConnectionStringKeyword parsedKey = stringKeywordMap.get(rawKey);
+        KnownKeywords parsedKey = KCSB_KEYWORDS.get(rawKey);
```
yogilad, line 82: Assuming `get()` throws, what's the value in the other exceptions thrown here?

data/src/main/java/com/microsoft/azure/kusto/data/auth/KcsbKeywords.java
```diff
@@ -0,0 +1,81 @@
...
+    private static KcsbKeywords readInstance() {
+        try {
+            ObjectMapper objectMapper = Utils.getObjectMapper();
+            try (InputStream resourceAsStream = KcsbKeywords.class.getResourceAsStream(
+                    "/kcsb.json")) {
+                KcsbKeywords value = objectMapper.readValue(resourceAsStream, KcsbKeywords.class);
+
+                // Validate the keywords
+                for (KeywordData keywordData : value.keywords) {
+                    KnownKeywords keyword = KnownKeywords.knownKeywords.get(keywordData.name);
```
yogilad, line 48: Is `get()` fails it means there's a keyword in the json we did not code.
Will there be an explicit error on this?

data/src/main/resources/kcsb.json
```diff
@@ -0,0 +1,218 @@
+{
```
yogilad, line 1: Quick suggestion so we can manage this easily.
Can we have a `"Version": "YYYY-MM-DD"` field that's only used to compare and align this file across repos?

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
ohadbitt: yes

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -164,12 +199,16 @@ public ConnectionStringBuilder(ConnectionStringBuilder other) {
         this.userNameForTracing = other.userNameForTracing;
         this.appendedClientVersionForTracing = other.appendedClientVersionForTracing;
         this.applicationNameForTracing = other.applicationNameForTracing;
+        this.sendX509 = other.sendX509;
+        this.initialCatalog = other.initialCatalog;
```
ohadbitt, line 203: seems like you are not really using all the new parameters - is this a next PR ?

data/src/main/java/com/microsoft/azure/kusto/data/auth/KnownKeywords.java
```diff
@@ -0,0 +1,72 @@
...
+    APPLICATION_CLIENT_ID("Application Client Id", true),
+    APPLICATION_KEY("Application Key", true),
+    USER_ID("User ID", true),
+    AUTHORITY_ID("Authority Id", true),
+    APPLICATION_TOKEN("Application Token", true),
+    USER_TOKEN("User Token", true),
+    APPLICATION_CERTIFICATE_X5C("Application Certificate SendX5c", true),
+    APPLICATION_NAME_FOR_TRACING("Application Name for Tracing", true),
+    USER_NAME_FOR_TRACING("User Name for Tracing", true),
+    PASSWORD("Password", false), // TODO - should this be enabled for alignment?
```
ohadbitt, line 20: just remove

data/src/main/java/com/microsoft/azure/kusto/data/auth/KnownKeywords.java
```diff
@@ -0,0 +1,72 @@
...
+    APPLICATION_CERTIFICATE_BLOB("Application Certificate Blob", false), // TODO - if we add a way to pass a certificate by a byte[], support this
+    APPLICATION_CERTIFICATE_THUMBPRINT("Application Certificate Thumbprint", false),
+    DSTS_FEDERATED_SECURITY("dSTS Federated Security", false),
+    STREAMING("Streaming", false),
+    UNCOMPRESSED("Uncompressed", false),
```
ohadbitt, line 25: what is uncompressed?

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
AsafMah: Recently added, but haven't been released outside of alpha.

I think I'll align it be the same logic we have in c# and python:

- If Fed=True and no auth specified - default to prompt
- If Fed is not given, but we have some other auth - assume Fed=true and use the auth
- If Fed=False and no auth given, assume no auth mode.

Sounds ok?

data/src/main/java/com/microsoft/azure/kusto/data/auth/KnownKeywords.java
AsafMah: It's a keyword in c#:

>             Uncompressed, // False by default. If set to true forces the client to not emit the Accept-Encoding header which allows the service to compress the response


We aren't using it anywhere else, it should just know that it's unsupported in case someone copy-pastes from .net

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -97,52 +70,114 @@ private ConnectionStringBuilder() {
...
+        this.sendX509 = false;
+        this.initialCatalog = null;
     }
 
     private void assignValue(String rawKey, String value) {
-        rawKey = rawKey.trim().toLowerCase();
-        ConnectionStringKeyword parsedKey = stringKeywordMap.get(rawKey);
+        KnownKeywords parsedKey = KCSB_KEYWORDS.get(rawKey);
         if (parsedKey == null) {
             throw new IllegalArgumentException("Error: unsupported key " + rawKey + " in connection string");
```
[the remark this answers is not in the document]
AsafMah: Good point - if it does reach this, it means that the keyword is marked as supported but isn't actually supported.

I updated the error to reflect this

data/src/main/resources/kcsb.json
AsafMah: Great suggestion, I'll also update in python.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
AsafMah: I think you wrote this comment twice

ohadbitt (approved):

yogilad (approved):

PR #413, opened 2025-01-27
Change: Make ManagedStreamingQueuingPolicy internal, expose just a factor

ohadbitt:
Change:
* Make ManagedStreamingQueuingPolicy internal, expose just a factor
* Dont allow users to pass raw data, provide it only if we have it


ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/BlobSourceInfo.java
```diff
@@ -66,4 +64,26 @@ public Map<String, String> getTracingAttributes() {
         }
         return attributes;
     }
+
+    /*
+     * For internal usage, adding blobExactSize
+     */
+    public static BlobSourceInfo fromFile(String blobPath, FileSourceInfo fileSourceInfo, CompressionType sourceCompressionType, boolean gotCompressed) {
```
AsafMah, line 71: If it's for internal use - can we lower the visibility?

AsafMah, line 71: The param on the outside is "shouldCompress", but here it's called "gotCompressed" - any reason?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/FileSourceInfo.java
```diff
@@ -47,4 +41,8 @@ public Map<String, String> getTracingAttributes() {
         }
         return attributes;
     }
+
+    public long fileSize() {
```
AsafMah, line 45: unused (and wrong?)

ingest/src/main/java/com/microsoft/azure/kusto/ingest/AzureStorageClient.java
```diff
@@ -97,37 +97,43 @@ void uploadStreamToBlob(InputStream inputStream, String blobName, BlobContainerC
...
 
         try (GZIPOutputStream gzout = new GZIPOutputStream(blob.getBlockBlobClient().getBlobOutputStream(true))) {
-            copyStream(inputStream, gzout, GZIP_BUFFER_SIZE);
+            return copyStream(inputStream, gzout, GZIP_BUFFER_SIZE);
         }
     }
 
-    private void copyStream(InputStream inputStream, OutputStream outputStream, int bufferSize) throws IOException {
+    // Returns original stream size
+    private int copyStream(InputStream inputStream, OutputStream outputStream, int bufferSize) throws IOException {
```
AsafMah, line 128: Why only this is private?

ingest/src/main/java/com/microsoft/azure/kusto/ingest/AzureStorageClient.java
```diff
@@ -97,37 +97,43 @@ void uploadStreamToBlob(InputStream inputStream, String blobName, BlobContainerC
...
         byte[] buffer = new byte[bufferSize];
```
AsafMah, line 129: I think there is an helper in apache for copy stream, you should check.

ingest/src/main/java/com/microsoft/azure/kusto/ingest/AzureStorageClient.java
```diff
@@ -86,8 +86,8 @@ void uploadFileToBlob(File sourceFile, BlobClient blobClient) throws IOException
         blobClient.uploadFromFile(sourceFile.getPath());
     }
 
-    void uploadStreamToBlob(InputStream inputStream, String blobName, BlobContainerClient container, boolean shouldCompress)
-            throws IOException, URISyntaxException, BlobStorageException {
+    int uploadStreamToBlob(InputStream inputStream, String blobName, BlobContainerClient container, boolean shouldCompress)
```
AsafMah, line 89: It's not clear that these methods return the size of the original stream, and only one is commented. 
I guess it's not that bad if they're internal, but it should probably be documented

ingest/src/main/java/com/microsoft/azure/kusto/ingest/ManagedStreamingIngestClient.java
```diff
@@ -461,12 +460,16 @@ protected IngestionResult ingestFromStreamImpl(StreamSourceInfo streamSourceInfo
...
-        this.queuingPolicy = queuingPolicy;
+    /**
+     * <p>setQueuingPolicyFactor</p>
+     * A factor used to tune the policy that handles the logic over which data size would the client choose to directly
+     * use queued ingestion instead of trying streaming ingestion first.
+     * Setting the factor will create a new {@link ManagedStreamingQueuingPolicy} with this factor, switching
+     * {@link ManagedStreamingQueuingPolicy.Default} which is created with no factor.
+     * @param factor - Default is 1.
+     **/
+    public void setQueuingPolicyFactor(double factor) {
```
AsafMah, line 471: Update CHANGELOG.md with all the changes, mention this specifically as a breaking change

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
```diff
@@ -236,37 +233,27 @@ private InputStream compressStream(InputStream uncompressedStream, boolean leave
...
     }
 
-    private IngestionResult ingestFromBlobImpl(BlobSourceInfo blobSourceInfo, IngestionProperties ingestionProperties, BlobClient cloudBlockBlob,
+    private IngestionResult ingestFromBlobImpl(BlobSourceInfo blobSourceInfo, IngestionProperties ingestionProperties,
             @Nullable String clientRequestId)
             throws IngestionClientException, IngestionServiceException {
         String blobPath = blobSourceInfo.getBlobPath();
-        try {
-            // No need to check blob size if it was given to us that it's not empty
-            if (blobSourceInfo.getRawSizeInBytes() == 0 && cloudBlockBlob.getProperties().getBlobSize() == 0) {
```
AsafMah, line 256: Do we check for empty blob elsewhere?

ingest/src/test/java/com/microsoft/azure/kusto/ingest/ManagedStreamingTest.java
```diff
@@ -10,59 +8,55 @@
...
-                isNull(), any(String.class), any(String.class), any(boolean.class))).thenReturn(null);
-
-        ingestionProperties = new IngestionProperties("dbName", "tableName");
-        managedStreamingIngestClient = new ManagedStreamingIngestClient(resourceManagerMock, azureStorageClientMock,
-                streamingClientMock);
-        queuedIngestClientMock = mock(QueuedIngestClientImpl.class);
-        managedStreamingIngestClientSpy = spy(
-                new ManagedStreamingIngestClient(mock(StreamingIngestClient.class), queuedIngestClientMock, new ExponentialRetry(1)));
-    }
+    // @BeforeAll
```
AsafMah, line 34: Why is everything commented out

AsafMah (commented):
Some comments

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/BlobSourceInfo.java
ohadbitt: no because its used outside the pacakge

ohadbitt: Client should care if it should compress
 . file naming should care only that it was compressed .

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/FileSourceInfo.java
```diff
@@ -23,14 +23,8 @@ public FileSourceInfo(String filePath) {
         this.filePath = filePath;
     }
 
-    // An estimation of the raw (uncompressed, un-indexed) size of the data, for binary formatted files - use only if known
-    public FileSourceInfo(String filePath, long rawSizeInBytes) {
+    public FileSourceInfo(String filePath, UUID sourceId) {
```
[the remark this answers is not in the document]
ohadbitt: its not v2 here

ingest/src/main/java/com/microsoft/azure/kusto/ingest/utils/IngestionUtils.java
```diff
@@ -39,22 +39,7 @@ public static StreamSourceInfo fileToStream(FileSourceInfo fileSourceInfo, boole
         }
 
         CompressionType compression = getCompression(filePath);
```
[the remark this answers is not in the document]
ohadbitt: not here - in the policy

ingest/src/main/java/com/microsoft/azure/kusto/ingest/AzureStorageClient.java
ohadbitt: nothing changed - this is used insiede this file - hence its private

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/FileSourceInfo.java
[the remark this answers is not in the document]
AsafMah: Yes, but just not to break the interface when we move

ingest/src/main/java/com/microsoft/azure/kusto/ingest/source/BlobSourceInfo.java
AsafMah: Ok, so maybe we want to document it like a public function?

Not sure what the right call is ehre

ingest/src/main/java/com/microsoft/azure/kusto/ingest/StreamingIngestClient.java
ohadbitt: what do you mean elsewhere ? here in java ? no

ingest/src/test/java/com/microsoft/azure/kusto/ingest/ManagedStreamingTest.java
ohadbitt: idk

AsafMah (approved):

PR #415, opened 2025-01-30
* Remove commons lang lib

ag-ramachandran:
* Remove usage of Pair
* Remove org.apache.commons.lang3 and org.apache.commons.text(has compile time dependency to org.apache.commons.lang3)
We had a strange issue in spark yesterday where commons lang with var-args in setConnector details was causing an issue. Thought of removing these libs.


data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -23,6 +22,8 @@ public class ConnectionStringBuilder {
     public static final String DEFAULT_DATABASE_NAME = "NetDefaultDb";
 
     private static final String DEFAULT_DEVICE_AUTH_TENANT = "organizations";
+    private static final String CLUSTER_URL_CANNOT_BE_NULL_OR_EMPTY = "clusterUrl cannot be null or empty";
```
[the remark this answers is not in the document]
AsafMah: Oh, we already have it with Ensure - let's use that

data/src/test/java/com/microsoft/azure/kusto/data/HeaderTest.java
```diff
@@ -147,7 +148,7 @@ public void testSetConnectorNameAndVersion() throws URISyntaxException, DataClie
     @Test
     public void testSetConnectorNoAppVersion() throws URISyntaxException, DataClientException {
         ConnectionStringBuilder csb = ConnectionStringBuilder.createWithAadManagedIdentity("https://testcluster.kusto.windows.net");
-        csb.setConnectorDetails("myConnector", "myVersion", null, null, true, "myApp");
+        csb.setConnectorDetails("myConnector", "myVersion", null, null, true, "myApp", Collections.emptyMap());
```
AsafMah, line 151: I think it should still work ok if it gets null

data/src/main/java/com/microsoft/azure/kusto/data/ClientDetails.java
```diff
@@ -1,63 +1,55 @@
...
-                    Pair.of("Kusto.Java.Client", Utils.getPackageVersion()),
-                    Pair.of(String.format("Runtime.%s", escapeField(getRuntime())), getJavaVersion())));
-        }
-    };
-    private String applicationForTracing;
-    private String userNameForTracing;
-    private String appendedClientVersionForTracing;
+
+    private static final ConcurrentHashMap<String, String> defaultValues = new ConcurrentHashMap<>();
+    public static final String DEFAULT_APPLICATION = "defaultApplication";
```
AsafMah, line 15: Cool idea to use ConcurrentHasmap's computeifabsent and then access these as keys.

But maybe instead of strings, it should be an enum?

AsafMah (commented):
Update changelog.md

AsafMah:
@ag-ramachandran any progress?

AsafMah:
@ag-ramachandran

data/src/main/java/com/microsoft/azure/kusto/data/instrumentation/Tracer.java
```diff
@@ -7,6 +7,8 @@
 import java.util.Map;
 
 public class Tracer {
+    private Tracer() {
```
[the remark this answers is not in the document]
tanmaya-panda1: yeah, added Private constructor to prevent instantiation as this class provides only static utility methods. Added a code comment as well.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -211,4 +218,398 @@ public static InputStream resolveInputStream(HttpResponse response, InputStream
...
+    /**
+     * Checks if a CharSequence is empty (""), null or whitespace only.
+     * 
+     * Whitespace is defined by {@link Character#isWhitespace(char)}.
+     *
+     * @param cs the CharSequence to check, may be null
+     * @return true if the CharSequence is empty or null or whitespace only
+     */
+    public static boolean isBlank(CharSequence cs) {
+        if (cs == null) {
```
AsafMah, line 234: Why don't your methods use each other? 
This could use isNullOrEmpty

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -211,4 +218,398 @@ public static InputStream resolveInputStream(HttpResponse response, InputStream
...
+
+    /**
+     * Removes the last character from a string.
+     * 
+     * If the string ends in {@code \r\n}, then remove both of them.
+     *
+     * @param str the string to chop
+     * @return the string with the last character removed, or null if null string input
+     */
+    public static String chop(final String str) {
```
AsafMah, line 269: I don't like the special case.
Make it into "chop" and "chopNewLine" or something

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -211,4 +218,398 @@ public static InputStream resolveInputStream(HttpResponse response, InputStream
...
+
+    /**
+     * Removes a substring from the end of a string (case sensitive).
+     *
+     * @param str    the string to process
+     * @param remove the substring to remove from the end
+     * @return the string with the suffix removed
+     */
+    public static String removeEnd(String str, String remove) {
+        if (str == null || remove == null || remove.isEmpty()) {
```
AsafMah, line 312: Again, why doesn't this use isnullorempty, 
or chop uses this, etc?

I'm not gonna comment this on every case, but do it for all methods.

data/src/main/java/com/microsoft/azure/kusto/data/Utils.java
```diff
@@ -211,4 +218,398 @@ public static InputStream resolveInputStream(HttpResponse response, InputStream
...
+        return str.toString().regionMatches(true, str.length() - suffix.length(), suffix.toString(), 0, suffix.length());
+    }
+
+    /**
+     * Custom implementation to replace Apache Commons Text TextStringBuilder.
+     * Provides formatted message building capabilities with line appending.
+     * This implementation matches the behavior of TextStringBuilder for the specific
+     * usage pattern in the codebase.
+     */
+    public static class MessageBuilder {
```
AsafMah, line 422: Do we really need this class?
We literally use it once in the code, and all it does is saves us a call to String.Format?

AsafMah (changes requested):

tanmaya-panda1:
added KustoResultSetTableDateTest with testcases which runs successfully with org.apache.commons.lang3.time.FastDateFormat(used in azure-kusto-java v6.0.2) as well as java.time.format.DateTimeFormatterBuilder.DateTimeFormatter(used in azure-kusto-java v.7.0.0-ALPHA-01)

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -645,4 +674,56 @@ public int count() {
     public boolean isNull(int columnIndex) {
         return get(columnIndex) == null;
     }
+
+    /**
+     * Helper method to normalize malformed datetime strings to make them parseable by Instant.parse()
+     * This handles common malformations like single-digit seconds, missing digits, etc.
+     */
+    private String normalizeMalformedDateTime(String dateString) {
```
tanmaya-panda1, line 682: this is added to handle 2023-07-15T14:30:4Z edge cases(KustoResultSetTableDateTest->testGetDate_WithEdgeCaseLengths_HandlesCorrectly()) which is parsed properly in the FastDateFormat.

data/src/main/java/com/microsoft/azure/kusto/data/auth/ConnectionStringBuilder.java
```diff
@@ -125,54 +124,52 @@ private void assignValue(String rawKey, String value) {
...
+            entries.put(KnownKeywords.APPLICATION_NAME_FOR_TRACING, applicationNameForTracing);
         }
 
         if (!StringUtils.isBlank(userNameForTracing)) {
-            entries.add(Pair.of(KnownKeywords.USER_NAME_FOR_TRACING, userNameForTracing));
+            entries.put(KnownKeywords.USER_NAME_FOR_TRACING, userNameForTracing);
         }
-
         StringBuilder sb = new StringBuilder();
+        entries.entrySet().stream()
```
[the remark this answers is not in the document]
AsafMah: @tanmaya-panda1 

I actually think the most readable form is
```java
        entries.forEach((key, value) ->
                sb.append(key.getCanonicalName())
                        .append("=")
                        .append((!showSecrets && key.isSecret()) ? SECRET_REPLACEMENT : value)
                        .append(";"));
```

And it should keep the order.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
```diff
@@ -569,13 +568,43 @@ public Date getDate(int columnIndex, Calendar calendar) throws SQLException {
...
                     }
                     String dateString = getString(columnIndex);
-                    FastDateFormat dateFormat;
-                    if (dateString.length() < 21) {
-                        dateFormat = FastDateFormat.getInstance("yyyy-MM-dd'T'HH:mm:ss", calendar.getTimeZone());
-                    } else {
-                        dateFormat = FastDateFormat.getInstance("yyyy-MM-dd'T'HH:mm:ss.SSS", calendar.getTimeZone());
+                    
+                    // First try the original FastDateFormat approach with strict patterns
+                    try {
```
AsafMah, line 573: Do you really need all of this?
Does the java datetimeformatter not support Z by itself?

Also there's seem to be support for lenient parsing - https://stackoverflow.com/questions/30090710/java-8-datetimeformatter-parsing-for-optional-fractional-seconds-of-varying-sign

Try messing with that before manually manipulating the string.

Either way, if there's complex logic like this I think it should go to a different method.

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
AsafMah: Can you expand more? How can we get this number? Or was this test just always there?

data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
```diff
@@ -28,4 +34,354 @@ public static String normalizeEntityName(String name) {
...
+     * Whitespace is defined by {@link Character#isWhitespace(char)}.
+     *
+     * @param cs the CharSequence to check, may be null
+     * @return true if the CharSequence is empty or null or whitespace only
+     */
+    public static boolean isBlank(CharSequence cs) {
+        if (isEmpty(cs)) {
+            return true;
+        }
+        int strLen = cs.length();
```
AsafMah, line 54: no need for variable

data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
```diff
@@ -28,4 +34,354 @@ public static String normalizeEntityName(String name) {
...
+    }
+
+    /**
+     * Removes the last character from a string, with special handling for line endings.
+     * If the string ends in {@code \r\n}, then remove both of them.
+     *
+     * @param str the string to chop
+     * @return the string with the last character removed, or null if null string input
+     */
+    public static String chopNewLine(final String str) {
```
AsafMah, line 101: Is this used anywhere? If not - remove it

data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
```diff
@@ -28,4 +34,354 @@ public static String normalizeEntityName(String name) {
...
+
+    /**
+     * Appends the suffix to the end of the string if the string does not already end with any of the suffixes.
+     *
+     * @param str the string to process
+     * @param suffix the suffix to append to the end of the string
+     * @param suffixes additional suffixes that are valid terminators
+     * @return a new String if suffix was appended, the same string otherwise
+     */
+    public static String appendIfMissing(String str, CharSequence suffix, CharSequence... suffixes) {
```
AsafMah, line 193: Why is this the only method that accepts CharSequence?

Also re-use your other methods like I mentioned before

data/src/main/java/com/microsoft/azure/kusto/data/StringUtils.java
```diff
@@ -28,4 +34,354 @@ public static String normalizeEntityName(String name) {
...
+        return str + suffix.toString();
+    }
+
+    /**
+     * Checks if a string is empty (null or zero length).
+     *
+     * @param str the string to check
+     * @return true if the string is null or has zero length
+     */
+    public static boolean isEmpty(String str) {
```
AsafMah, line 222: isempty twice?

AsafMah (commented):
More comments

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
ag-ramachandran: changed the implementation a bit to use flexible format using a formatter builder. Can you have a look on this @AsafMah

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
ag-ramachandran: changed the implementation a bit to use flexible format using a formatter builder. Can you have a look on this @AsafMah

data/src/main/java/com/microsoft/azure/kusto/data/KustoResultSetTable.java
ag-ramachandran: Fixed with DateTimeformatter builder and tests seem to pass

AsafMah (approved):

PR #416, opened 2025-02-05
Refactor error handling in Kusto API to merge multiple errors better

AsafMah:

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceQueryError.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceQueryError.java
@@ -48,7 +48,10 @@ public class KustoServiceQueryError extends AzureException {
             RuntimeException exception;
             if (isOneApi) {
                 DataWebException ex = new DataWebException(value);
-                message = ex.getApiError().getCode() + ": " + ex.getApiError().getMessage();
+                OneApiError apiError = ex.getApiError();
+                if (apiError != null) {
+                    message = apiError.getCode() + ": " + apiError.getMessage();
+                }
                 exception = ex;
             } else {
                 exception = new RuntimeException(value);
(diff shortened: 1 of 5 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/exceptions/OneApiError.java
```diff
@@ -15,6 +20,12 @@ public OneApiError(String code, String message, String description, String type,
         this.permanent = permanent;
     }
 
+    public static OneApiError[] fromJsonArray(ArrayNode array) {
+        return StreamSupport.stream(Spliterators.spliteratorUnknownSize(array.elements(), Spliterator.ORDERED), false)
```
ohadbitt, line 24: why all this splitator things ? Stream . not good enough?

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/KustoServiceQueryError.java
```diff
@@ -33,6 +28,31 @@ public KustoServiceQueryError(String message) {
...
+        }
+
+        for (int i = 0; i < jsonExceptions.size(); i++) {
+            if (isOneApi) {
+                exceptions.add(new DataWebException(jsonExceptions.get(i).toString()));
+            } else {
+                exceptions.add(new RuntimeException(jsonExceptions.get(i).toString()));
+            }
+            sb.append(jsonExceptions.get(i).toString());
+            sb.append("\n");
```
ohadbitt, line 50: did we decide this together with yochai and all ? that we will print them all as message ?
anyway i dont think the message should contain jsonExceptions.get(i).toString(),
maybe a UT is needed so we see the output and know this is what we want

data/src/main/java/com/microsoft/azure/kusto/data/exceptions/OneApiError.java
AsafMah: I think I checked and that's the "correct" way to to it.

AsafMah: If you have another direction let me know

ohadbitt (approved):

PR #417, opened 2025-02-05
Enable OneBox hostnames that include the port

yihezkel:
### Fixed
Onebox URLs with port were rejected, and are now accepted.

### DIFF
--- a/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
+++ b/ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
@@ -2,6 +2,7 @@ package com.microsoft.azure.kusto.ingest;
 
 import com.microsoft.azure.kusto.data.exceptions.ExceptionUtils;
 import com.microsoft.azure.kusto.ingest.source.CompressionType;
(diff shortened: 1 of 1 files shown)


ingest/src/main/java/com/microsoft/azure/kusto/ingest/IngestClientBase.java
```diff
@@ -2,6 +2,7 @@
 
 import com.microsoft.azure.kusto.data.exceptions.ExceptionUtils;
 import com.microsoft.azure.kusto.ingest.source.CompressionType;
+import org.apache.commons.lang3.StringUtils;
```
AsafMah, line 5: are you using this import?

yihezkel: Via `StringUtils.isEmpty()` on line 58

AsafMah: Right now there is a PR #415 that removes this dep entirely.

Can you use `CoreUtils.isNullOrEmpty`?

AsafMah (approved):

PR #418, opened 2025-02-05
Enable Quickstart to run commands, as opposed to just queries

yihezkel:
### Fixed
Enable Quickstart to run commands, as opposed to just queries. I assume it previously worked, and I'm not sure why it stopped working.


AsafMah:
I think it might have to do with the 6.0.0 changes and the refactoring of the api to sync/async

@ohadbitt should also take a look

ohadbitt (approved):

ohadbitt:
> I think it might have to do with the 6.0.0 changes and the refactoring of the api to sync/async
> 
> @ohadbitt should also take a look

yes some bad refactor

georgebanasios:
I think this might have been removed in the past by mistake, because I don't see it even on the release prior to 6.0.0

AsafMah (approved):

PR #423, opened 2025-03-27
TImeout fixes

AsafMah:
-Add client grace period to request timeout and introduce server timeout bypass option
- Adjust timeout calculation in `ClientImpl` to include a 30-second grace period.
- Add `OPTION_SERVER_TIMEOUT_NO_REQUEST_TIMEOUT` for bypassing request timeout in `ClientRequestProperties`.
- Update minimum timeout to 1 second from 1 minute for more precise configuration.

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -417,6 +417,11 @@ class ClientImpl extends BaseClient {
             throw new DataClientException(clusterUrl, "Failed to parse timeout from ClientRequestProperties");
         }
 
+        Object skipBoolean = properties == null ? null : properties.getOption(ClientRequestProperties.OPTION_SERVER_TIMEOUT);
+        if (skipBoolean instanceof Boolean && (Boolean) skipBoolean) {
+            return Long.MAX_VALUE;
+        }
+
         if (timeoutMs == null) {
             switch (commandType) {
                 case ADMIN_COMMAND:
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -189,7 +189,8 @@ class ClientImpl extends BaseClient {
     private Mono<String> executeWithTimeout(KustoRequest request, String nameOfSpan) {
         return prepareRequestAsync(request)
                 .zipWhen(requestContext -> {
-                    long timeoutMs = determineTimeout(request.getProperties(), request.getCommandType(), clusterUrl);
+                    ClientRequestProperties properties = request.getProperties() == null ? new ClientRequestProperties() : request.getProperties();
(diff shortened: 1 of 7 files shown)


ohadbitt (approved):

ohadbitt:
just add a changelog

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -416,6 +417,11 @@ private long determineTimeout(ClientRequestProperties properties, CommandType co
             throw new DataClientException(clusterUrl, "Failed to parse timeout from ClientRequestProperties");
         }
 
+        Object skipBoolean = properties == null ? null : properties.getOption(ClientRequestProperties.OPTION_SERVER_TIMEOUT);
```
yogilad, line 420: The way we figure out what timeouts to apply seem convoluted.
Maybe consolidate everything to this API and put it in 415 instead of trying to figure out what 415 really meant.

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -172,10 +173,15 @@ public void clearParameters() {
     }
 
     /**
-     * Gets the amount of time a query may execute on the service before it times out. Value must be between 1 minute and 1 hour,
+     * Gets the amount of time a query may execute on the service before it times out. Value must be between 1 second and 1 hour,
      * and so if the value had been set below the minimum or above the maximum, the value returned will be adjusted accordingly.
      */
     public Long getTimeoutInMilliSec() {
+        Object noTimeout = getOption(OPTION_SERVER_TIMEOUT_NO_REQUEST_TIMEOUT);
```
yogilad, line 180: This feels wrong.
I don't think this public API should consider the other property just cause we need it somewhere

AsafMah: IDK, the api doesn't get the property as-is, it always does a conversion to ms anyway.

So it's not about getting the option itself, but the effective timeout, which is why i think it's corect to have both options considered.

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
AsafMah: Not sure I understand - 415 is irrelevant to this line. 

We don't check the result of 415 here, we independently check the header

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
yogilad: Effective timeout sounds like the job that `determineTimeout` is doing.
If this is also this APIs job, name it accordingly

data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
yogilad: Why though?

PR #426, opened 2025-03-30
Fix client grace period

AsafMah:

### DIFF
--- a/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
+++ b/data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
@@ -413,16 +413,16 @@ class ClientImpl extends BaseClient {
     }
 
     private long determineTimeout(ClientRequestProperties properties, CommandType commandType, String clusterUrl) {
-        Long timeoutMs;
-        try {
-            timeoutMs = properties == null ? null : properties.getTimeoutInMilliSec();
-        } catch (ParseException e) {
-            throw new DataClientException(clusterUrl, "Failed to parse timeout from ClientRequestProperties");
-        }
-
-        Object skipBoolean = properties == null ? null : properties.getOption(ClientRequestProperties.OPTION_SERVER_TIMEOUT);
+        Object skipBoolean = properties.getOption(ClientRequestProperties.OPTION_NO_REQUEST_TIMEOUT);
         if (skipBoolean instanceof Boolean && (Boolean) skipBoolean) {
             return Long.MAX_VALUE;
+        }
+
+        Long timeoutMs;
+        try {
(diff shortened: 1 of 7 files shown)


data/src/main/java/com/microsoft/azure/kusto/data/ClientImpl.java
```diff
@@ -416,6 +420,11 @@ private long determineTimeout(ClientRequestProperties properties, CommandType co
             throw new DataClientException(clusterUrl, "Failed to parse timeout from ClientRequestProperties");
         }
 
+        Object skipBoolean = properties == null ? null : properties.getOption(ClientRequestProperties.OPTION_SERVER_TIMEOUT);
```
yogilad: No Request Timeout + do this before reading the value

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -172,10 +173,15 @@ public void clearParameters() {
     }
 
     /**
-     * Gets the amount of time a query may execute on the service before it times out. Value must be between 1 minute and 1 hour,
+     * Gets the amount of time a query may execute on the service before it times out. Value must be between 1 second and 1 hour,
      * and so if the value had been set below the minimum or above the maximum, the value returned will be adjusted accordingly.
      */
     public Long getTimeoutInMilliSec() {
+        Object noTimeout = getOption(OPTION_SERVER_TIMEOUT_NO_REQUEST_TIMEOUT);
```
yogilad: Undo breaking

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
```diff
@@ -288,6 +285,25 @@ public static ClientRequestProperties fromString(String json) throws JsonProcess
         return null;
     }
 
+    private static Object unwrapJsonValue(JsonNode jsonNode) {
```
ohadbitt, line 288: is this for returning null ? what changed in behavior here?

ohadbitt (approved):
isnt this a change in behavior ?

data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java
AsafMah: So before we had a subtle bug, when serializing the props and deserializing.

If we had a number there, it would read it as a string (because of line 274), and when it will try to parse it it will fail (since it's not in the proper timespan format). This makes sure that when deserializing it preserves the type of the object.

yogilad (approved):

PR #440, opened 2025-09-13
Feature/add ingestion source blob

tanmaya-panda1:
add ingestion source blobtype to QueuedIngest


ingest-v2/src/main/kotlin/com/microsoft/azure/kusto/ingest/v2/QueuedIngestionApiWrapper.kt
```diff
@@ -0,0 +1,234 @@
...
+                    table = table,
+                    ingestionMethod = IngestionMethod.QUEUED,
+                    operationId = operationId,
+                    storedResults = emptyList()
+                )
+            } else {
+                val errorMessage = "Failed to submit queued ingestion to $database.$table. " +
+                    "Status: ${response.status}, Body: ${response.body()}"
+                logger.error(errorMessage)
+                throw IngestException(errorMessage)
```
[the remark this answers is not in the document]
AsafMah: You should also parse OneApiExceptions, but maybe out of scope here.

ingest-v2/src/test/kotlin/com/microsoft/azure/kusto/ingest/v2/ConfigurationApiWrapperTest.kt
```diff
@@ -23,7 +25,7 @@ class ConfigurationApiWrapperTest {
 
     private fun endpointAndExceptionClause(): Stream<Arguments?> {
         return Stream.of(
-            Arguments.of(System.getenv("DM_CONNECTION_STRING"), false),
+            Arguments.of(System.getenv("DM_CONNECTION_STRING") ?: "https://test.kusto.windows.net", false),
```
AsafMah, line 28: Remove the hardcoded name

ingest-v2/src/main/kotlin/com/microsoft/azure/kusto/ingest/v2/common/serialization/OffsetDateTimeSerializer.kt
```diff
@@ -0,0 +1,28 @@
...
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import java.time.OffsetDateTime
+import java.time.format.DateTimeFormatter
+
+/**
+ * Custom serializer for OffsetDateTime to handle JSON serialization/deserialization.
+ */
+object OffsetDateTimeSerializer : KSerializer<OffsetDateTime> {
```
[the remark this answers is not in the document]
ag-ramachandran: Great point, I think it is because the format is an offset datetime when generated from the model. In Jackson and other java libs the java times are automatically serialized, but when used with Kotlin, these needed an explicit serializer. Will try and see if there is a workaround, for now to hold the deserialized values, we need an extra class

ingest-v2/src/main/kotlin/com/microsoft/azure/kusto/ingest/v2/common/models/mapping/ColumnMapping.kt
```diff
@@ -0,0 +1,96 @@
...
+import kotlinx.serialization.Serializable as KSerializable
+
+@KSerializable
+data class ColumnMapping(
+    val columnName: String,
+    val columnType: String,
+    val properties: MutableMap<String, String> = mutableMapOf(),
+) {
+
+    fun setPath(path: String) {
```
AsafMah: Getters/Setters aren't the kotlin way

You can use props, I.E

```kotlin
var path: String?
    get() = properties[MappingConstants.Path.name]
    set(value) =  properties[MappingConstants.Path.name] = value
```

ingest-v2/src/main/kotlin/com/microsoft/azure/kusto/ingest/v2/source/BlobSource.kt
```diff
@@ -42,5 +44,6 @@ class BlobSource : IngestionSource {
 
     // No resources to close; method intentionally left empty.
     override fun close() {
+        TODO("Not yet implemented")
```
AsafMah: The comment above contradicts the todo

PR #471, opened 2026-04-02
Add maven-assembly-plugin, maven-source-plugin, and dokka-maven-plugi…

tanmaya-panda1:
This pull request enhances the build and release process for the `ingest-v2` module by introducing new Maven plugins and updating the release workflow to generate and publish additional artifact types, including a "jar-with-dependencies," source jar, and Javadoc jar. These changes improve artifact distribution and documentation.

**Build and artifact generation improvements:**

* Added the `maven-assembly-plugin` to `ingest-v2/pom.xml` to automatically create a "jar-with-dependencies" during the package phase, enabling easier distribution of a standalone jar with all dependencies included.
* Added the `maven-source-plugin` to generate and attach a sources jar during the verify phase, improving source code availability for consumers.
* Added the `dokka-maven-plugin` for generating and attaching a Javadoc jar using Dokka, enhancing documentation for Kotlin code.
* Introduced a `dokka.version` property for managing the Dokka plugin version.

**Release workflow updates:**

* Updated `.github/workflows/release.yml` to upload the new "jar-with-dependencies" artifact for `kusto-ingest-v2` as part of the release process.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,6 +98,7 @@ jobs:
             staging/kusto-ingest-${{ steps.get_version.outputs.VERSION }}-sources.jar
             staging/kusto-ingest-${{ steps.get_version.outputs.VERSION }}.pom
             staging/kusto-ingest-v2-${{ steps.get_ingest_v2_version.outputs.INGEST_V2_VERSION }}.jar
+            staging/kusto-ingest-v2-${{ steps.get_ingest_v2_version.outputs.INGEST_V2_VERSION }}-jar-with-dependencies.jar
             staging/kusto-ingest-v2-${{ steps.get_ingest_v2_version.outputs.INGEST_V2_VERSION }}-javadoc.jar
             staging/kusto-ingest-v2-${{ steps.get_ingest_v2_version.outputs.INGEST_V2_VERSION }}-sources.jar
             staging/kusto-ingest-v2-${{ steps.get_ingest_v2_version.outputs.INGEST_V2_VERSION }}.pom
--- a/ingest-v2/pom.xml
+++ b/ingest-v2/pom.xml
@@ -25,6 +25,7 @@
         <slf4j.version>2.0.9</slf4j.version>
         <spotless.version>2.46.1</spotless.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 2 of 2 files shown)


ingest-v2/pom.xml
```diff
@@ -361,6 +362,59 @@
...
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-source-plugin</artifactId>
+                <version>${maven-source-plugin.version}</version>
+                <executions>
+                    <execution>
+                        <id>attach-sources</id>
```
ag-ramachandran, line 396: can we just do mvn sources:jar javadoc:jar instead of introducing a new plugin

.github/workflows/release.yml
```diff
@@ -98,6 +98,7 @@ jobs:
             staging/kusto-ingest-${{ steps.get_version.outputs.VERSION }}-sources.jar
             staging/kusto-ingest-${{ steps.get_version.outputs.VERSION }}.pom
             staging/kusto-ingest-v2-${{ steps.get_ingest_v2_version.outputs.INGEST_V2_VERSION }}.jar
+            staging/kusto-ingest-v2-${{ steps.get_ingest_v2_version.outputs.INGEST_V2_VERSION }}-jar-with-dependencies.jar
```
ag-ramachandran, line 101: why should we copy the one with dependencies ?, we need not have uber jars

ag-ramachandran (changes requested):

ingest-v2/pom.xml
tanmaya-panda1: Actually, I tried with maven-javadoc-plugin which does not work for pure Kotlin modules. It silently produces no output (no jar file created) because the javadoc tool only processes .java files. Since ingest-v2 has sourceDirectory set to src/main/kotlin with only .kt files, dokka-maven-plugin is the standard and necessary tool for generating Kotlin API documentation.

.github/workflows/release.yml
tanmaya-panda1: wanted to produce the same set of jars as it is done for kusto-ingest and kusto-data hence added

ag-ramachandran (approved):

